PR #3, opened 2026-07-11
feat(#DBSYNC-11): stream large-file upload/download to fix OOM

mbsanchez:
## Summary
Fixes OOM on large-file transfers by converting the transfer layer to constant-memory streaming, and makes downloads crash-safe. Delivered as 4 vertical slices (S1∥S2 → S3 → S4).

- **S1 — Download streaming + atomic replace:** `fetch_and_write_file` streams the response via `std::io::copy` into a same-directory temp file, then atomically replaces the target (Windows `MoveFileExW(MOVEFILE_REPLACE_EXISTING)` via `windows-sys` / POSIX `fs::rename`). One shared helper routes all 3 download call sites; no partial file ever appears at the final path.
- **S2 — Streaming upload <150MB:** replaced `fs::read` with `Body::sized(file, len)`; added a 150MB threshold const + pure `choose_upload_strategy`.
- **S3 — Chunked upload-session ≥150MB:** `start`/`append_v2`/`finish` with an 8MB reused buffer and advancing cursor offset; a premature-EOF guard prevents a busy-loop if the file is truncated mid-upload.
- **S4 — Upload progress:** emits `upload-progress {path,transferred,total}` from the session path via a set-once `static APP_HANDLE: OnceLock<AppHandle>`; minimal `App.tsx` listener.

## Stack
desktop-tauri

## Jira Ticket
#DBSYNC-11

## Design notes
- **No async migration, no new reqwest feature flags** — `reqwest::blocking::Response: Read` (download) and `Body::sized`/`From<Vec<u8>>` (upload) all work under the existing `blocking` feature. Only new dep: `windows-sys` (Windows-only, already present transitively via `winreg`).
- **AppHandle plumbing deviation (from planned D1):** storing `AppHandle` as a field on the `.manage()`-registered `AppState` crashed the test binary (`STATUS_ENTRYPOINT_NOT_FOUND`) on this toolchain (rustc 1.94.1 / tauri 2.10.3 / Windows). Switched to a module-level `static OnceLock<AppHandle>` set once in `setup()` — functionally equivalent, set-once, no-op when unset. Documented inline in `state.rs`.
- Scheduler thread and `sync_running` guard untouched. Upload-session resumability across retries is an explicit non-goal (a retry starts a fresh session).

## Test Plan
- [x] Automated tests pass — `cargo test` **23/23** (in `apps/desktop/src-tauri`)
[...]

PR #4, opened 2026-07-11
fix(#DBSYNC-11): make large-file uploads actually complete (recover, retry, resume)

mbsanchez:
## Summary
Follow-up to PR #3. The chunked upload-session path shipped there routed >=150MB files correctly, but a **real 6.47GB upload still never completed**: the job zombied in `status='running'` with no recovery, and a single transient network error on `upload_session/append_v2` aborted the whole session with no retry and no resume (next attempt restarted from byte 0). This makes large-file upload actually robust.

- **F1 — Recover stale `running` jobs** on startup (`recover_running_jobs`: reset to `queued`, `attempt_count=0`). `pick_next_due_job` never re-selects `running`, so any interruption (app close, crash, network drop mid-chunk) previously froze the job forever.
- **F2 — HTTP timeouts** on transfer paths (`http_client()`: connect 15s, request 600s) so a dead connection fails cleanly and retries instead of hanging.
- **F3 — Per-chunk retry with backoff** (`retry_transient`) on transport / `too_many_write_operations` / 429 / 5xx errors; wraps every `start`/`append_v2`/`finish`.
- **F4 — Real error detail** (`describe_reqwest_error` walks the `source()` chain) so failures show `connection reset` / `timed out` instead of only "error sending request".
- **F5 — Resumable upload session**: checkpoint `{session_id, offset, file_len, file_mtime}` persisted in `sync_jobs` after every chunk (additive DB columns). On retry/restart it resumes from the last confirmed byte instead of re-uploading GBs; file identity (size+mtime) is validated so a file changed mid-upload can't commit a mixed/corrupt object; a fresh session is started once if Dropbox reports the session invalid/expired.

## Stack
desktop-tauri

## Jira Ticket
#DBSYNC-11 (reopened — the AC "large files upload/download" was not met in practice)

## Review fixes applied (from code review)
- **M1**: dropped the bare `409` clause from `is_session_invalid_error` (a transient `too_many_write_operations` 409 no longer forces a full multi-GB re-upload) and made rate-limits/5xx retryable in `retry_transient`.
- **M2**: resume now validates `file_len` + `file_mtime` against the checkpoint before reusing a session — prevents committing `OLD[0..off] + NEW[off..]` if the local file changed between attempts.
- **L3**: request timeout raised 300s → 600s for slow uplinks.

## Test Plan
- [x] `cargo test` — **33 passed / 0 failed** (in `apps/desktop/src-tauri`)
- [x] `cargo clippy` — clean on changed files (3 pre-existing baseline warnings elsewhere)
- [x] `npm --workspace apps/desktop run build` — OK
- [x] **Real-world verified**: a genuine 6.47GB firmware `.zip` now uploads via the chunked session with visible per-chunk progress in the activity log (previously stuck retrying the oversized single-request endpoint, then zombied in `running`).

### Deferred / notes
- Download-side resume (HTTP Range) not added — downloads keep the crash-safe temp+atomic-replace + job-level retry. Metadata-only calls still use `Client::new()` (DBSYNC-14 scope).
- `incorrect_offset` after a crash-window falls back to one full session restart rather than seeking to the server's `correct_offset` (correct, not optimal).

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/dropbox_transfer.rs
+++ b/apps/desktop/src-tauri/src/dropbox_transfer.rs
@@ -1,8 +1,9 @@
 use std::collections::HashMap;
 use std::fs::{self, File};
 use std::io;
-use std::io::Read;
+use std::io::{Read, Seek, SeekFrom};
 use std::path::{Path, PathBuf};
+use std::time::Duration;
 
 use reqwest::blocking::{Body, Client};
 use tauri::{Emitter, EventTarget};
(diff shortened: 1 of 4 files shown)


mbsanchez (commented):
## CTO Review — DBSYNC-11: large-file resumable upload  ✅ APPROVED

Independent CTO gate on `07a834b`. All prior-review findings (M1 bare-409, M2 file-identity guard, L3 timeout) are applied and hold. Verified locally: `cargo test` **33 passed / 0 failed**, `cargo clippy` **no new warnings** (4 total, all pre-existing & outside the diff), `vite build` clean.

### Recovery (F1) — correct
- `recover_running_jobs()` runs inside `setup()`, single-threaded, **before** `start_background_scheduler` — race-free; can't zombie a live job since nothing is in-flight at process start.
- Resets only `status='running'` → `queued`, `attempt_count=0`, and **preserves** the checkpoint columns so the upload resumes. Confirmed by `recover_running_jobs_preserves_upload_checkpoint`.

### Resume (F5) — correct
- Offset persisted only **after** a 2xx append, so it always equals server-committed bytes.
- Resume guard `off <= len && ck_len == len && ck_mtime == file_mtime` prevents appending new bytes onto a stale session — any mismatch starts fresh at offset 0.
- Restart-on-invalid-session bounded to one via the `restarted` flag; a second `is_session_invalid_error` returns `Err` (no infinite loop, no wrong-size finish). Checkpoint cleared on success and nulled on terminal failure in `mark_job_failed`.

### Retry (F3) — correct
- `retry_transient` retries transport / rate-limit / 429 / 5xx only — plain `400` not retried. Retrying `finish` is non-corrupting (commit is `mode: overwrite`, idempotent on content).

### Session classification (F4/M1) — correct
- `is_session_invalid_error` matches only `lookup_failed/not_found/incorrect_offset/closed/not_closed`, **not** a bare 409 — `too_many_write_operations` stays in the retry path (test-confirmed).

### DB migration — safe
- Additive via `add_column_if_missing`; read connection opened after `migrate`; all job queries select explicit columns, so `SyncJobRow` is unaffected.

### Non-blocking nit
- `file_mtime` falls back to `0` when unreadable; an identity match could then hinge on `len` alone — extremely narrow given `ck_len == len` must also hold. Fine to leave.

**Verdict on resume/retry/recovery correctness: sound. APPROVED.**

mbsanchez:
## QA Result: PASSED ✅

Stack `desktop-tauri` — automated suite + real-world manual verification.

**Automated**
- `cargo test --lib` → **33 passed, 0 failed** (exit 0): recover_running_jobs (+ preserves checkpoint), upload_checkpoint round-trip / file-identity, retry_transient classification (transport/429 retried, 400 not), is_session_invalid_error markers (bare-409 excluded), plus the existing atomic_replace / choose_upload_strategy / is_final_chunk suites.
- `cargo clippy` → no new warnings from changed files.
- `npm --workspace apps/desktop run build` → OK.

**Manual (real-world) — the previously-deferred large-file ACs are now genuinely met**
- A real 6.47 GB firmware `.zip` now uploads via the chunked upload-session with **visible per-chunk progress** in the activity log — the exact case that previously (a) hit the oversized single-request endpoint and then (b) zombied in status='running'. Confirmed live by the reporter.
- Zombie recovery (F1) re-picks the interrupted job on startup; per-chunk retry + resumable session (F5) keep the multi-GB upload progressing across transient errors.

All acceptance criteria met (automated + real-world). Ready for PO approval.

PR #6, opened 2026-07-12
fix(#DBSYNC-12): preserve unsynced local edits on remote-wins download

mbsanchez:
## Summary
Fixes silent data loss: a remote-wins `download` job atomically overwrote the local file with no check, destroying unsynced local edits when a file changed on both sides. (Upload and remote-delete conflicts were already guarded; the download path was not.)

- **Download-time guard** in `download_remote_file_internal`, BEFORE the atomic replace: if the on-disk file exists and its content hash differs from the last-synced baseline (`local_file_index`), preserve the local edit via `create_conflicted_copy`, record it in `sync_conflicts` (`add_conflict`) + bump the counter (`record_conflict`), then proceed with the download (remote-wins; the local edit survives as the copy).
- Pure `download_would_conflict(target_exists, on_disk_hash, baseline_hash)` predicate (unit-tested). Reuses existing conflict infra — **no new table, no new IPC command**.
- New `sync-conflict` event via the set-once `APP_HANDLE` (mirrors `emit_upload_progress`) + a minimal `App.tsx` listener (`pushLog` + `refreshDashboard`). Payload `{ path, conflictPath, reason }` (camelCase).
- **Retry dedupe by content** (`conflict_copy_with_content_exists`): a retrying download re-observing the same local edit reuses the existing conflicted copy instead of spawning duplicate copies/rows; a genuinely different later edit still produces a new copy. Survives restarts (copies on disk); best-effort (never suppresses a real conflict).

## Stack
desktop-tauri

## Jira Ticket
#DBSYNC-12

## Design notes
- **Critical invariant:** `create_conflicted_copy` runs strictly BEFORE `fetch_and_write_file` — the copy captures the intact local edit, never the overwritten remote content. Every fallible step is `?`-propagated, so a failed copy/DB-write aborts the download before overwriting.
- The fix is centralized inside `download_remote_file_internal`, so all THREE call sites (scheduled download job, manual download command, `.cloudsc` hydration) are protected with zero call-site changes. Hydration is naturally safe (target absent → no conflict).
- `baseline None → not a conflict` (only reachable at the manual/hydration call sites; the sync path always has an index row).

## Test Plan
- [x] `cargo test` — **43/43** (in `apps/desktop/src-tauri`): 4 predicate cases, content-dedup test, `emit_sync_conflict` no-op-without-handle.
- [x] `cargo clippy` — clean on changed files.
- [x] `npm --workspace apps/desktop run build` — OK (new listener + event type compile).

### desktop-tauri validation / deferred to manual QA (live Dropbox)
- [ ] Edit a file locally, change the same file on Dropbox, let a tick run: local edit preserved as `(conflicted copy TS).ext`, target holds remote content, conflict appears in the Conflicts list + counter increments, `sync-conflict` log line appears immediately.
- [ ] Regression: brand-new download (target absent) and no-edit re-download (hash == baseline) do NOT create spurious conflicts; `.cloudsc` hydration does not flag a conflict.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/dropbox_transfer.rs
+++ b/apps/desktop/src-tauri/src/dropbox_transfer.rs
@@ -10,11 +10,12 @@ use tauri::{Emitter, EventTarget};
 
 use crate::auth_session::get_access_token;
 use crate::models::{
-    DropboxListFolderResponse, ListRemoteFolderResponse, RemoteEntry, UploadProgressEvent,
-    UploadSessionStartResponse,
+    DropboxListFolderResponse, ListRemoteFolderResponse, RemoteEntry, SyncConflictEvent,
+    UploadProgressEvent, UploadSessionStartResponse,
 };
 use crate::path_util::{
-    hash_file, is_path_allowed, normalize_dropbox_path, parse_prefix_csv,
+    create_conflicted_copy, hash_file, is_path_allowed, normalize_dropbox_path, parse_prefix_csv,
+    relpath_under,
 };
 use crate::state::AppState;
 use crate::storage::db::FileIndexRow;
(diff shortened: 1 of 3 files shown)


mbsanchez (commented):
## CTO Review — DBSYNC-12 (remote-wins download conflict resolution) — APPROVED

Data-loss-sensitive change reviewed against the two existing conflict precedents (`sync_pipeline.rs` pending-upload conflict, `remote_index.rs` remote-delete conflict). The safety-critical invariants all hold.

### Ordering / data-loss safety — SAFE
- The whole conflict guard runs **before** `fetch_and_write_file` (the atomic overwrite).
- `create_conflicted_copy` uses `fs::copy` → the original is never consumed; it survives a failed copy.
- Every fallible step is `?`-propagated ahead of the overwrite: `hash_file`, `get_local_file`, `conflict_copy_with_content_exists`, `create_conflicted_copy`, `relpath_under`, `add_conflict`. A failure aborts the download before any write. No path overwrites after a failed copy; no path copies after the overwrite.

### Dedupe correctness — CORRECT
- `conflict_copy_with_content_exists` matches by **content hash**, not existence: a different later edit still produces a new copy; only same-content retries (`max_attempts=5`, re-invokes the download each attempt) are deduped. Read/hash errors → `false`, so it can never suppress a genuine conflict. The `"{stem} (conflicted copy "` prefix matches `create_conflicted_copy`'s naming for both extension and no-extension files.

### Predicate & ordering
- `download_would_conflict` 3-branch logic correct; `baseline None → not a conflict` is sound (only manual/hydration sites; sync path always has an index row).
- `add_conflict` (`?`) precedes `record_conflict` (best-effort `if let Ok`), emit last; `emit_sync_conflict` no-ops when `APP_HANDLE` is unset. No unwrap/panic in prod.

### Frontend
- `App.tsx` mirrors the `upload-progress` listener exactly (active flag + unlisten cleanup). Payload camelCase matches the `SyncConflictEvent` DTO. No leak.

### Build / lint / test
- `cargo clippy`: only the 4 pre-existing baseline warnings — none in changed files.
- `cargo test`: all green (5 new tests incl. content-dedupe + no-op emit).
- `tsc && vite build`: clean.

Non-blocking (acceptable per the content-addressed design; do not gate merge):
1. A new edit reproducing content identical to a still-present older conflict copy is deduped (no new row), but the identical bytes remain on disk → no data loss.
2. If `add_conflict` fails after the copy is written, that attempt leaves an orphan copy with no DB row; retry dedupe then skips and the download proceeds with the local edit still preserved.

**Verdict: APPROVED.** Ordering, dedupe, and data-loss safety are all sound.

mbsanchez:
## QA Result: PASSED ✅ (automated-only + code inspection)

Stack `desktop-tauri` — backend Rust + minimal frontend listener.

**Test commands**
- `cargo test --lib` → **43 passed, 0 failed** (exit 0): the 4 `download_would_conflict` predicate cases, `conflict_copy_dedup_matches_by_content_not_existence`, `emit_sync_conflict_is_a_no_op_without_app_handle`.
- `cargo clippy` → 0 warnings from changed files.
- `npm --workspace apps/desktop run build` → **built OK** (exit 0): new `SyncConflictEvent` type + `sync-conflict` listener compile.

**AC coverage**
- Automated / structurally verified: conflict predicate (target-absent / baseline-None / hash==baseline / hash!=baseline); content-based retry dedupe; emit no-op-without-handle; DTO + listener build; ordering (copy strictly before atomic overwrite) confirmed by CTO review.

**Deferred to human/manual QA** (needs a live Dropbox account — no HTTP mocking in the crate):
- Edit-both-sides round trip: local edit preserved as `(conflicted copy TS).ext`, target holds remote content, conflict recorded in the Conflicts list + counter increments, `sync-conflict` log line appears immediately.
- Regression: brand-new download (target absent) and no-edit re-download (hash == baseline) do NOT create spurious conflicts; `.cloudsc` hydration does not flag a conflict.

Automated + inspection ACs met. Ready for PO approval; real edit-both-sides validation to be done at human QA.

PR #7, opened 2026-07-12
fix(#DBSYNC-13): make sync gate and token refresh race-free

mbsanchez:
## Summary
Fixes two TOCTOU concurrency defects in the desktop backend.

- **P1 — `sync_running`:** the scheduler checked the flag and set it in **two separate lock acquisitions**, so two callers could both observe `false` and start a concurrent sync. Moved from a mutex-guarded `bool` on `SyncEngine` to an `Arc<AtomicBool>` on `AppState` (mirrors `OauthListenerControl.cancel`); every gate is now one `compare_exchange(false, true, AcqRel, Acquire)` and every release a `store(false, Release)`. `current_status()` takes the flag as a parameter. Deleting `is_/set_sync_running` from `SyncEngine` made any missed call site a hard compile error (none remained).
- **P2 — token refresh:** `get_access_token` dropped the token-cache lock before the blocking HTTP refresh, so two threads could double-refresh and orphan the access token (Dropbox rotates the refresh token). Serialized behind a dedicated `token_refresh_lock` with a **double-checked cache re-read**, extracted into `refresh_token_guarded(injected refresher)`. `get_access_token` / `force_refresh_session` / `verify_dropbox_token_internal` route through it.

## Stack
desktop-tauri

## Jira Ticket
#DBSYNC-13

## Design / review notes
- Gate/release balance verified: no gated body has an early `return`/`?` between a successful `compare_exchange` and its `store(false)`; every internal `Err` is captured & logged, never propagated before release → no stuck-flag / wedged-sync.
- Orderings: `AcqRel`/`Acquire` on the CAS, `Release` on the store, `Acquire` on status reads.
- **Review fix (Medium):** `force` skips the cache short-circuit — a forced refresh (401 recovery) must always exchange the refresh token, even against a time-fresh-but-revoked cached token; it still serializes on the mutex (no concurrent refresh).
- **Review fix (Low):** the `Mutex<()>` recovers from poison (`into_inner`) so a refresher panic can't permanently wedge every future refresh.
- No lock-ordering inversion (only nesting is `token_refresh_lock` → `token_cache`); the refresher never touches `sync_engine`.
- No API/IPC/frontend change — `SyncStatus.sync_running` shape unchanged (only its source).

## Test Plan
- [x] `cargo test` — **45/45** (in `apps/desktop/src-tauri`): N-thread `compare_exchange` single-winner test; concurrent-refreshers-collapse-to-one test.
- [x] `cargo clippy` — clean on changed files.
- [x] `npm --workspace apps/desktop run build` — OK (SyncStatus shape unchanged).

### Deferred to manual QA (live app / network)
- [ ] No overlapping real sync when a scheduler tick coincides with an IPC trigger.
- [ ] A single real network refresh under concurrent expired-token access (no rotation-induced logout).

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/auth_session.rs
+++ b/apps/desktop/src-tauri/src/auth_session.rs
@@ -1,3 +1,6 @@
+use std::sync::atomic::Ordering;
+use std::sync::{Arc, Mutex};
+
 use chrono::{Duration, Utc};
 use reqwest::blocking::Client;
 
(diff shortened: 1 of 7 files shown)


mbsanchez (commented):
## CTO review — DBSYNC-13 (concurrency / TOCTOU) — APPROVED

Independent gate after the implement-time review. Verified both concurrency fixes on commit `e6616e9`. Build + 45 tests + clippy all green (no new warnings; the 4 clippy warnings are the pre-existing baseline in `run_events.rs`, `storage/db.rs`, `lib.rs`).

### S1 — `Arc<AtomicBool>` single-flight gate
**Gate/release balance verified for all 6 entry points** — the load-bearing invariant. In each, `store(false, Release)` is the final statement of the gated body with no `?` or early `return` between the successful CAS and release:
- `commands.rs` scheduler, `trigger_sync_tick`, `trigger_hydrate_cloudsc_placeholder`, `trigger_download_remote_file`, `trigger_hydrate_remote_folder`
- `open_handlers.rs` drain loop (the early-return there is on CAS **failure**, gate not held)

Orderings `AcqRel/Acquire` (CAS), `Release` (store), `Acquire` (load) are correct. All `is_/set_sync_running` removed; the 3 `current_status(flag)` callers pass the atomic or `false` where the syncing case is already short-circuited. `SyncStatus` shape unchanged; `AppState` stays `Clone`.

### S2 — serialized token refresh (`refresh_token_guarded`)
- `force=true` on 401 recovery (`force_refresh_session`) correctly bypasses the cache short-circuit and always exchanges, while still serializing on `token_refresh_lock` — no concurrent endpoint hit.
- `force=false` (`get_access_token`) double-checks the cache after acquiring the lock, collapsing concurrent refreshers to one exchange.
- The refresh mutex is **not** held across the blocking `refresher()` network call; the `token_cache` guard is dropped before IO. Lock nesting is consistently `token_refresh_lock → token_cache`; the refresher never touches `sync_engine`. No deadlock.
- Poison recovery via `into_inner()` is sound for `Mutex<()>` and prevents a refresher panic from permanently wedging refresh.
- `TokenSession` construction and the `get_access_token` signature are preserved.

### Tests
`compare_exchange_gate_allows_exactly_one_winner` and `concurrent_refreshers_collapse_to_a_single_refresh` directly assert the single-winner and single-flight invariants.

**Verdict:** gate/release balance correct on every path (incl. spawned threads and Err branches); refresh serialization correct and IO-safe. **APPROVED.**

_Non-blocking:_ a panic (vs `Err`) inside a gated body would leave the atomic `true` and wedge sync — behavior-equivalent to the previous poisoned-mutex path and out of scope. A future RAII/scopeguard release would make it panic-safe.

mbsanchez:
## QA Result: PASSED ✅ (automated-only + inspection)

Stack `desktop-tauri` — backend Rust concurrency change (no frontend surface).

**Test commands**
- `cargo test --lib` → **45 passed, 0 failed** (exit 0): `compare_exchange_gate_allows_exactly_one_winner` (S1 single-winner), `concurrent_refreshers_collapse_to_a_single_refresh` (S2 single-flight).
- `cargo clippy` → 0 warnings from changed files.
- `npm --workspace apps/desktop run build` → **OK** (exit 0) — SyncStatus shape unchanged.

**AC coverage**
- Automated / build-verified: `Arc<AtomicBool>` flag + both constructors; flag removed from SyncEngine + `current_status(flag)`; all 6 gates via `compare_exchange` + 6 releases via `store(false)`; 3 callers + tray migrated; `token_refresh_lock` + double-checked guarded-refresh helper; `get_access_token`/`force_refresh`/`verify` routed. Gate/release balance on every path (incl. spawned threads + Err branches) confirmed by CTO review.

**Deferred to human/inspection QA** (requires the live app / concurrent network):
- No overlapping real sync when a scheduler tick coincides with an IPC trigger.
- A single real network refresh under concurrent expired-token access (no rotation-induced logout).

Automated + inspection ACs met. Ready for PO approval.

PR #13, opened 2026-07-12
fix(#DBSYNC-14): bound every reqwest call with per-operation timeouts

mbsanchez:
## Summary
No request timeout was set, so a stalled Dropbox server could park an OS thread indefinitely and freeze the sync scheduler. This sets `connect_timeout` 10s and per-operation request timeouts:

| Operation | Timeout |
|---|---|
| Quick JSON (metadata / list_folder / delete / token-verify / refresh) | **30s** (shared-client default) |
| Upload-session chunks (start / append_v2 / finish, ≤8 MiB) | **120s** each |
| Whole-file single-request transfers (single-shot upload ≤150 MiB; streamed file download/hydration) | **300s** |
| Async OAuth token exchange (`complete_oauth`) | **30s** (was unbounded) |

- Timeout errors are now explicitly classified transient in `retry_transient` (added `"timed out"` / `"timeout"` matching) → they retry with backoff.
- `sync_running` is already released on the `Err` path by DBSYNC-13's gate (`store(false, Release)` runs unconditionally after the tick), so a timeout can't wedge the scheduler — confirmed, no change needed.

## Stack
desktop-tauri (Rust backend)

## Jira Ticket
#DBSYNC-14

## Deviation from AC (documented in code)
The AC says a flat "120s for transfers". Whole-file **single-request** transfers (the streamed download/hydration, and single-shot uploads up to the 150 MiB session threshold) use **300s** instead: reqwest's blocking `timeout` bounds the *entire* request including the full body stream, so a 120s cap would abort a legitimate large-file transfer on a slow link mid-stream. Bounded upload-session **chunks** (≤8 MiB) keep 120s. A future ranged/chunked download could tighten this.

## Test Plan
- [x] `cargo test --lib` — **50 passed**.
- [x] `cargo build --lib` — **0 warnings**.
- [x] `cargo clippy --lib` — no new warnings from changed files.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/auth/oauth.rs
+++ b/apps/desktop/src-tauri/src/auth/oauth.rs
@@ -90,7 +90,11 @@ pub fn start_oauth() -> Result<(String, String, String), String> {
 pub async fn complete_oauth(code: String, verifier: String) -> Result<TokenResponse, String> {
     let app_key = resolve_app_key()?;
     let redirect_uri = resolve_redirect_uri();
-    let response = Client::new()
+    let client = Client::builder()
+        .timeout(std::time::Duration::from_secs(30))
+        .build()
+        .map_err(|e| format!("failed to build oauth http client: {e}"))?;
+    let response = client
         .post("https://api.dropboxapi.com/oauth2/token")
         .form(&[
             ("code", code),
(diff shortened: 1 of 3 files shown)


mbsanchez:

Reviewed the timeout matrix. Verified: connect 10s + 30s default on the shared client covers all quick JSON calls; upload-session chunks (≤8 MiB) at 120s and whole-file single-request transfers (single-shot upload ≤150 MiB + streamed download/hydration) at 300s — the 300s split is the correct call (blocking `timeout` bounds the whole streamed body; 120s would regress large-file hydration). Async `complete_oauth` now bounded at 30s (was unbounded). Timeouts explicitly classified transient in `retry_transient` → retryable. `sync_running` reset-on-timeout already guaranteed by DBSYNC-13's unconditional gate release. cargo test 50 green, build 0 warnings, clippy clean on changed files. The AC deviation is documented in code and justified. Next: QA.

PR #14, opened 2026-07-12
perf(#DBSYNC-16): batch remote index via one recursive list_folder (kill N+1)

mbsanchez:
## Summary
- `refresh_remote_index_and_enqueue_downloads_internal` made **one `get_metadata` HTTP request per local file** — 1000 files = 1000 sequential requests per sync tick, exhausting Dropbox rate limits.
- New `fetch_all_remote_file_metadata(state)`: **one recursive `list_folder`** from the account root + cursor pagination, building a `HashMap` keyed by `path_display.to_lowercase()` (Dropbox is case-insensitive). The loop now does an in-memory lookup by `normalize_dropbox_path(&rel).to_lowercase()` instead of a per-file HTTP call.
- Deletion-propagation, `should_download`, `upsert_remote_file`, `enqueue_job` — **byte-for-byte unchanged**; only the source of `remote_meta` changed.

## Mass-delete safety
`fetch_all_remote_file_metadata` returns `Err` on **any** page/parse failure — never a partial map. A path is treated as "remote-deleted" only when a **complete** listing omits it; a partial fetch aborts the tick via `?` before any local deletion is enqueued.

## Stack
desktop-tauri (Rust backend)

## Jira Ticket
#DBSYNC-16

## Test Plan
- [x] `cargo test --lib` — **54 passed** (+4 new: file→lowercased-key+parsed-ts; folder / empty-hash / missing-path_display → None).
- [x] `cargo build --lib` — 0 warnings.
- [x] `cargo clippy --lib` — no new warnings from `remote_index.rs`.

## Notes
- Single-file `fetch_remote_file_metadata` kept for its on-demand caller in `dropbox_transfer.rs`.
- Batch calls inherit the shared client's 30s timeout (DBSYNC-14); each page is bounded.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/remote_index.rs
+++ b/apps/desktop/src-tauri/src/remote_index.rs
@@ -1,9 +1,9 @@
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 
 use chrono::Utc;
 
 use crate::auth_session::get_access_token;
-use crate::models::DropboxEntry;
+use crate::models::{DropboxEntry, DropboxListFolderResponse};
 use crate::path_util::normalize_dropbox_path;
 use crate::state::AppState;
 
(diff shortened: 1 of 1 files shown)


mbsanchez:

Reviewed the N+1 fix. Verified: (1) one recursive `list_folder` + cursor pagination replaces the per-file `get_metadata` loop; map keyed by `path_display.to_lowercase()`, looked up by `normalize_dropbox_path(&rel).to_lowercase()` — consistent case-insensitive keys. (2) **Mass-delete safety is the critical property and it's correct**: the batch returns `Err` on any page/parse failure (never a partial map), so 'absent ⇒ remote-deleted' only fires on a complete listing; a partial fetch aborts via `?` before any `local_delete` is enqueued. (3) Deletion-propagation / should_download / upsert / enqueue unchanged — only the source of `remote_meta` moved from HTTP to map lookup. (4) Pure `remote_meta_from_entry` helper unit-tested (file/folder/empty-hash/missing-path). cargo test 54 green, build 0 warnings, clippy clean on the file. KISS. Next: QA.

PR #16, opened 2026-07-12
chore(#DBSYNC-44): resolve remaining clippy style lints

mbsanchez:
## Summary
Clears the 4 remaining `cargo clippy --lib` style warnings on `main` (style-only, no behavior change):
- **db.rs** — `current_data_dir_os()`: cfg arms return via tail expression instead of `return …;` (removes the `needless_return` introduced by DBSYNC-24).
- **lib.rs** — pass `app` / `app_handle` directly instead of `&app` / `&app_handle` (`needless_borrow` ×2).
- **run_events.rs** — collapse the `WebviewEvent::DragDrop` `if let` into the outer match-arm pattern (`collapsible_match`).

## Stack
desktop-tauri (Rust backend)

## Jira Ticket
#DBSYNC-44

## Test Plan
- [x] `cargo clippy --lib` — **0 warnings** (was 4).
- [x] `cargo build --lib` — 0 warnings.
- [x] `cargo test --lib` — **54 passed**.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/lib.rs
+++ b/apps/desktop/src-tauri/src/lib.rs
@@ -62,7 +62,7 @@ pub fn run() {
             let paths = crate::open_handlers::cloudsc_paths_from_argv(&argv);
             if !paths.is_empty() {
                 tracing::info!(?paths, "file open on running instance");
-                crate::open_handlers::handle_cloudsc_paths_from_os(&app, paths);
+                crate::open_handlers::handle_cloudsc_paths_from_os(app, paths);
             }
             // Do not raise an empty dashboard when the user is fully set up (tray-only workflow).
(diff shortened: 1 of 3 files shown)


mbsanchez:

Style-only lint cleanup. Verified each is behavior-preserving: db.rs cfg arms as tail exprs (only one compiles per target, so it's the function's tail — same value); `app`/`app_handle` are already `&AppHandle` so dropping the extra `&` passes the same reference; the collapsed WebviewEvent arm matches exactly the same DragDrop::Drop case and non-matching webview events still fall to `_ => {}` as before. cargo clippy --lib now 0 warnings, build 0 warnings, test 54 green. Next: QA.

PR #25, opened 2026-07-13
fix(#DBSYNC-45): propagate remote deletion of re-hydrated files; log .cloudsc changes

mbsanchez:
## Bug A (correctness) — remote deletion of a re-hydrated file wasn't propagated
`download_remote_file_internal` (hydration/download) recorded `local_files` but **not** `remote_files`. That row is otherwise only written by the `refresh_remote_index` sweep when a file is present on both sides — so a freshly re-hydrated file had no `remote_files` row, and deleting its remote copy before the next sweep left `prev_remote = None`, which the deletion branch can't distinguish from "never uploaded" → no `local_delete` (the re-add repro: A synced+deleted works; re-add A, hydrate, delete → not removed locally).

**Fix:** after a hydrate/download, `upsert_remote_file` with the **downloaded content hash** (= Dropbox `content_hash`, DBSYNC-9 — exactly what the deletion/should-download checks compare, so correct even if the best-effort rev/mtime metadata fetch fails). Local edits after hydration still route to the conflict branch (no silent delete).

## Bug B (UX) — .cloudsc create/prune not in Activité
`cloudsc_ops` now emits one **summarised** `placeholder-changed` `{created,removed}` event per index sweep (no per-file flood on a large initial index). New `usePlaceholderEvents` hook logs placeholder add/prune into the flyout Activité feed (per-item when ≤6, otherwise a summary line).

## Stack
desktop-tauri · Jira #DBSYNC-45

## Test Plan
- [x] `cargo build` + `cargo clippy -- -D warnings` clean; `cargo test --lib` **64 pass**.
- [x] `npm --workspace apps/desktop run build` (tsc+vite) — 0 TS errors.
- [ ] **Manual QA (`npm run dev:win`)**:
  - Re-add repro: file A in remote → hydrate → delete A remotely → **A is now removed locally** (verify the previously-failing second round works).
  - Local-edit guard: hydrate, edit locally, delete remotely → conflict copy preserved (not silently deleted).
  - Activité shows a "Placeholder ajouté/retiré" entry when a `.cloudsc` is created/pruned.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/cloudsc_ops.rs
+++ b/apps/desktop/src-tauri/src/cloudsc_ops.rs
@@ -2,6 +2,7 @@ use std::collections::HashSet;
 use std::fs;
 use std::path::{Path, PathBuf};
 
+use tauri::{Emitter, EventTarget};
 use walkdir::WalkDir;
 
 use crate::auth_session::get_access_token;
@@ -30,6 +31,11 @@ pub(crate) fn index_remote_folder_children_as_cloudsc_placeholders_internal(
 
     let client = &state.http_client;
     let mut created = 0usize;
+    // DBSYNC-45: collect placeholder create/prune names to emit ONE
+    // `placeholder-changed` event per index call (no per-file event flood on a
+    // large initial sweep); the frontend logs them into the Activité feed.
+    let mut created_names: Vec<String> = Vec::new();
+    let mut removed_names: Vec<String> = Vec::new();
 
     // Page through the folder with cursor + has_more; a folder with more than one
     // page of entries (Dropbox returns up to ~2000 per page) would otherwise be
@@ -108,6 +114,7 @@ pub(crate) fn index_remote_folder_children_as_cloudsc_placeholders_internal(
             };
             write_cloudsc_placeholder_file(&placeholder_path, &meta)?;
             created += 1;
+            created_names.push(child_name.clone());
         }
 
         if !entries_resp.has_more {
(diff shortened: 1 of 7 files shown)


mbsanchez:
## CTO Review + QA: APPROVED / PASSED ✅ (user-validated after live data-loss debugging)

Root cause turned out to be a path-separator mismatch, fixed in layers, plus a pre-existing data-loss hole in the deletion propagation:
- **Path normalization** (relpath_under + the 3 filesystem-derived keys in scan_local_changes + db index layer + a startup migration, using char(92) since a literal '\' is parsed unreliably by SQLite) — all relative-path keys are now '/'-canonical. This stops the hydrate→upload oscillation and makes get_remote_file(local_rel) match so remote deletions of hydrated files propagate.
- **Hydration** now records remote_file_index provenance.
- **Data-loss guard**: never propagate a local→remote delete (file or recursive folder delete_v2) for a DEHYDRATED path — if a '<rel>.cloudsc' placeholder exists it was dehydrated, not deleted; untrack locally only. This was destroying the user's remote copies during the oscillation.
- .cloudsc create/prune now surface in Activité.

Diagnosed live against the user's app.db (jobs 43-60 showed the delete/upload cascade), killed the runaway process, cleaned the stale DB state. cargo build + clippy -D warnings clean, 64 tests. **User validated** on their real folder: restored files come down as .cloudsc, hydrate no longer re-uploads, remote deletion of a hydrated file propagates locally, and nothing is spuriously deleted remotely.

Merging → main.

Issue #29, opened 2026-07-13
DBSYNC-28: Zeroize PKCE verifier + restrict DROPBOX_APP_KEY visibility (AFK)


mbsanchez:
Vertical slice of **DBSYNC-28** (desktop-tauri, security hardening). Single cohesive change in the auth area.

## Context
- The PKCE code verifier lives in `SyncEngine.pending_pkce_verifier: Option<String>` (`sync/engine.rs`) — a plain heap `String` with no zeroization; its bytes linger in freed memory after login. The getter `pending_pkce_verifier()` clones it, and that clone (used at `auth/oauth_complete.rs:26`) also lingers.
- `DROPBOX_APP_KEY` is `pub const` (`auth/oauth.rs:12`), wider visibility than needed.

## Approach (KISS)
1. Add the `zeroize` crate. Change the field to `Option<zeroize::Zeroizing<String>>`; `set_oauth_context` wraps with `Zeroizing::new`; `clear_oauth_pending` sets `None` (drop wipes); the getter returns `Zeroizing<String>` so the copy at the call site wipes too. Adjust `oauth_complete.rs:26` to the new return type (Deref<Target=String> keeps the token-exchange call working).
2. `DROPBOX_APP_KEY`: `pub const` → `pub(crate) const` (only referenced in-crate).
3. Document the threat model in a doc comment: the app key is the public OAuth `client_id`, not a secret — restricting visibility is hygiene, not a confidentiality control. `pending_oauth_state` (CSRF token) is intentionally left un-zeroized (not sensitive).

## Acceptance Criteria
- [ ] `pending_pkce_verifier` uses `zeroize::Zeroizing` so the verifier is wiped on clear and on drop.
- [ ] The verifier copy handed to the token exchange (`oauth_complete.rs`) is also a `Zeroizing` value (no lingering plain `String` clone).
- [ ] `DROPBOX_APP_KEY` is `pub(crate) const`.
- [ ] A doc comment states the app key is a public OAuth client_id, not a secret.
- [ ] `cargo build`, `cargo test`, `cargo clippy --all-targets -- -D warnings` clean; OAuth login flow still works.

Type: AFK · Scope: apps/desktop/src-tauri (src/sync/engine.rs, src/auth/oauth.rs, src/auth/oauth_complete.rs, Cargo.toml) · Blocked by: none · User stories: 1,2,3

mbsanchez:
[...]

PR #30, opened 2026-07-13
fix(security): zeroize PKCE verifier, restrict app key visibility (#DBSYNC-28)

mbsanchez:
## Summary
Two auth-hygiene fixes for **DBSYNC-28** (#29).

- **Zeroize PKCE verifier**: `SyncEngine.pending_pkce_verifier` is now `Option<Zeroizing<String>>` — wiped on clear and on drop. The getter returns `Zeroizing<String>` and `complete_oauth` takes `Zeroizing<String>`, building the token-exchange form via `.as_str()` so no plain-`String` clone of the verifier escapes. `pending_oauth_state` (CSRF token) stays a plain `String` (not sensitive).
- **Restrict app key**: `DROPBOX_APP_KEY` `pub const` → `pub(crate) const` (used only in-crate), with a doc comment on the threat model — it is the public OAuth `client_id`, not a secret; the change is visibility hygiene, not a confidentiality control.
- Added `zeroize` 1.9 as a direct dependency.

## Stack
desktop-tauri

## Jira Ticket
#DBSYNC-28

## Test Plan
- [x] `cargo test` — 69/69 pass
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] OAuth reconnect smoke — login/token-exchange flow unchanged (validated by user)

## Notes
Zeroization-on-drop is a type-level guarantee (`Zeroizing`), not directly unit-testable (freed memory). The transient urlencoded request body inside reqwest is the best-effort limit — it is owned by reqwest and dropped after send.

🤖 Generated with Claude Code

https://claude.ai/code/session_01XFcb2z2QG5oQe6v8GyDwhd

### DIFF
--- a/apps/desktop/src-tauri/Cargo.lock
+++ b/apps/desktop/src-tauri/Cargo.lock
@@ -940,6 +940,7 @@ dependencies = [
  "walkdir",
  "windows-sys 0.59.0",
  "winreg",
+ "zeroize",
 ]
 
 [[package]]
@@ -6374,18 +6375,18 @@ dependencies = [
 
 [[package]]
 name = "zeroize"
-version = "1.8.2"
+version = "1.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b97154e67e32c85465826e8bcc1c59429aaaf107c1e4a9e53c8d8ccd5eff88d0"
+checksum = "e13c156562582aa81c60cb29407084cdb54c4164760106ab78e6c5b0858cf64e"
 dependencies = [
  "zeroize_derive",
 ]
 
 [[package]]
 name = "zeroize_derive"
-version = "1.4.3"
+version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "85a5b4158499876c763cb03bc4e49185d3cccbabb15b33c627f7884f43db852e"
+checksum = "3c50655cbb0fe3fc43170059e702f1ce5e19b84cec58dc87b037a09935c2f328"
 dependencies = [
  "proc-macro2",
  "quote",
--- a/apps/desktop/src-tauri/Cargo.toml
+++ b/apps/desktop/src-tauri/Cargo.toml
@@ -33,6 +33,7 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 url = "2"
 urlencoding = "2"
 walkdir = "2"
+zeroize = "1.9.0"
 
 [dev-dependencies]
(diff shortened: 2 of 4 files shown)


mbsanchez:

Independent review of the diff. No blockers, no majors.

Verified:
- Getter returns a fresh `Zeroizing<String>` (clone of `Zeroizing` is itself zeroized on drop) — no plain-String copy escapes the engine.
- `oauth_complete.rs` call site + `complete_oauth(verifier: Zeroizing<String>)` type-line-up confirmed by `clippy --all-targets -D warnings` (clean).
- Form built via `.as_str()` on all fields → no owned copy of the verifier at the call site; it drops+wipes at function end. The transient serde_urlencoded body (owned by reqwest) is the honestly-documented best-effort limit.
- `set_oauth_context(verifier: String)` MOVES the single heap allocation into `Zeroizing::new` — bytes never duplicated.
- `pending_oauth_state` left plain: correct — the CSRF state token is non-confidential (travels in the authorize URL).
- `DROPBOX_APP_KEY` → `pub(crate)`: no out-of-crate reference breaks; doc threat model accurate (public client_id, PKCE protects the flow).

Optional follow-ups (non-blocking):
- `pkce_verifier()`/`start_oauth()` could return `Zeroizing<String>` for airtight coverage; residual is stack pointer metadata only (every hop is a move).
- **Out of scope**: `refresh_access_token_blocking` handles the long-lived `refresh_token` as a plain `String` — arguably more sensitive than the one-shot verifier. Worth a future hardening ticket.

Ready for QA.

mbsanchez:
## QA Result: PASSED ✅ (automated-only)

| AC | Evidence |
|---|---|
| verifier uses `zeroize` (wiped on clear/drop) | `engine.rs:23` `pending_pkce_verifier: Option<Zeroizing<String>>` (type-level guarantee) + compiles/clippy |
| verifier copy to token exchange is Zeroizing | getter returns `Zeroizing<String>`; `complete_oauth(verifier: Zeroizing<String>)`; form via `.as_str()` |
| `DROPBOX_APP_KEY` is `pub(crate) const` | `oauth.rs:20` |
| threat-model documented | `oauth.rs:15` doc comment (public OAuth client_id, not a secret) |
| OAuth login still works | user reconnect smoke (login/token exchange unchanged) |

Gates: `cargo test` 69/69 · `cargo clippy --all-targets -- -D warnings` clean · frontend unaffected (no FE changes).
Exception (Branch E): zeroization-on-drop and the login round-trip aren't unit-testable — covered by the type guarantee + user smoke.

Ready for PO approval.

PR #34, opened 2026-07-13
fix(ux): reflect Syncing/Idle in tray tooltip on sync transitions (#DBSYNC-48)

mbsanchez:
## Summary
The system-tray tooltip was stuck on `Idle` during a sync: it was only refreshed by the 60s scheduler poll, which almost always misses the short `sync_running=true` window.

Adds `auth_session::refresh_tray_tooltip(state)` — reads the already-global `crate::state::APP_HANDLE` (background sync threads carry no `AppHandle`) and re-renders the tooltip from `current_tray_status_label`. Called at **every** `sync_running` transition (start after the compare_exchange-to-true, end after `store(false)`): the 4 trigger commands, the scheduler tick, and the `.cloudsc`-open drain in `open_handlers`. Event-driven, no busy-loop, no change to the poll cadence.

## Stack
desktop-tauri

## Jira Ticket
#DBSYNC-48

## Test Plan
- [x] `cargo test` — 69/69 pass
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] Tray tooltip shows `Syncing` during a sync and `Idle`/`Error` after — validated by user

🤖 Generated with Claude Code

https://claude.ai/code/session_01XFcb2z2QG5oQe6v8GyDwhd

### DIFF
--- a/apps/desktop/src-tauri/src/auth_session.rs
+++ b/apps/desktop/src-tauri/src/auth_session.rs
@@ -256,6 +256,17 @@ pub(crate) fn update_tray_tooltip(app: &tauri::AppHandle, label: &str) {
     }
 }
 
+/// Refreshes the tray tooltip to reflect the current sync status. Uses the
+/// globally-stored `AppHandle` (set in `setup()`), so background sync threads —
+/// which don't carry an `AppHandle` — can refresh it right at each
+/// `sync_running` transition instead of waiting for the 60s scheduler poll,
+/// which almost always misses the short "Syncing" window (DBSYNC-48).
+pub(crate) fn refresh_tray_tooltip(state: &AppState) {
+    if let Some(app) = crate::state::APP_HANDLE.get() {
+        update_tray_tooltip(app, &current_tray_status_label(state));
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::{refresh_token_guarded, TokenSession};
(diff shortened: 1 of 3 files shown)


mbsanchez:

No blockers. Verified the primary risk (deadlock): `current_tray_status_label` short-circuits to "Syncing" and returns BEFORE locking `sync_engine`, so all 6 start-refreshes (sync_running already true) never touch the engine lock; all end-refreshes run after the scoped `if let Ok(engine)` blocks have dropped their guard and after `store(false)` — no re-entrancy.

Coverage: all 6 sync_running false→true→false paths refresh on both transitions (4 trigger commands + scheduler + cloudsc-open drain). Start is synchronous on the IPC thread before the worker spawns → `Syncing → work → Idle/Error` ordering guaranteed. APP_HANDLE is set in setup() before any command/scheduler runs (`.get()` is a safe no-op otherwise). Exactly 2 refreshes per sync + the unchanged 60s poll — no busy-loop.

Nits (non-blocking): scheduler uses `update_tray_tooltip` directly (has AppHandle in scope) vs the wrapper — justified.

Ready for QA.

mbsanchez:
## QA Result: PASSED ✅ (automated-only)

| AC | Evidence |
|---|---|
| tooltip → Syncing on start, Idle/Error on finish | refresh at all 6 sync_running transitions; user smoke confirmed Syncing→Idle |
| no busy-loop / excessive updates | exactly 2 refreshes per sync + unchanged 60s poll |
| build clean | see gates |

Gates: `cargo test` 69/69 · `cargo clippy --all-targets -- -D warnings` clean.
UI behavior validated by user smoke (tray tooltip). Ready for PO approval.

PR #62, opened 2026-07-16
feat(settings): tray Settings entry + user local ignore patterns (#DBSYNC-36)

mbsanchez:
## Summary
DBSYNC-36 slices **S1 + S2** (the two AFK slices). S3 (disconnect) and S4 (startup-at-login) follow in their own PRs.

**S1 — tray Settings entry**: the tray menu gains a "Settings" item above "Exit" that opens (and focuses — no duplicate window) the existing `setup` window, which already renders the Dropbox account / sync folder / selective-sync settings panel.

**S2 — user-defined local ignore patterns**:
- `Thumbs.db` now ignored by default alongside `.DS_Store` / editor temps.
- Users add extra patterns, persisted in `app_config` as `ignore_globs_csv` (**no schema change**), OR'd into the single `is_ignored_local_path` predicate.
- KISS hand-rolled matching (no new crate): case-insensitive **exact basename**, **`*`-suffix**, or **exact relative path**. Real `**/` globs are out of scope.
- Folded `fs_watcher` + `cloudsc_ops` onto the combined predicate so user patterns apply to both the full scan and the fs watcher.
- Frontend **"Ignored files"** section (local-only), visually distinct from the remote selective-sync control.

## Stack
desktop-tauri

## Jira Ticket
DBSYNC-36 (In Review)

## Test Plan
- [x] `cargo test --lib` — 134 pass (5 new path_util + 1 db), `cargo clippy --lib` clean
- [x] `tsc` + `vite build` pass
- [x] desktop-tauri: exercised affected commands; built + launched, tray "Settings" opens the panel, "Ignored files" saves; no terminal errors

### Notes
- Local ignore (this PR) is a distinct axis from remote selective-sync (`exclude_prefixes_csv`) — kept separate in storage and UI.
- Out of scope (own tickets/PRs): bandwidth throttle (de-scoped), S3 disconnect, S4 startup-at-login.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/cloudsc_ops.rs
+++ b/apps/desktop/src-tauri/src/cloudsc_ops.rs
@@ -15,8 +15,8 @@ use crate::models::{
     CloudscMeta, CloudscPlaceholderInfo, DropboxListFolderResponse,
 };
 use crate::path_util::{
-    hash_file, is_path_allowed, normalize_dropbox_path, parse_prefix_csv, relpath_under, safe_join,
-    should_ignore_local_path,
+    hash_file, is_ignored_local_path, is_path_allowed, normalize_dropbox_path, parse_prefix_csv,
+    relpath_under, safe_join,
 };
 use crate::state::AppState;
 
@@ -770,7 +770,7 @@ fn dehydrate_folder_cfapi(state: &AppState, root: &Path, folder_rel: &str) -> Ap
             Ok(r) => r.to_string_lossy().replace('\\', "/"),
             Err(_) => continue,
         };
-        if file_rel.ends_with(".cloudsc") || should_ignore_local_path(&file_rel) {
+        if file_rel.ends_with(".cloudsc") || is_ignored_local_path(&file_rel) {
             continue;
         }
         // Already cloud-only — never re-hash it (that would hydrate it).
@@ -851,7 +851,7 @@ fn dehydrate_folder_collapse(
             cloudsc_children.push(entry.path().to_path_buf());
             continue;
         }
-        if should_ignore_local_path(&file_rel) {
+        if is_ignored_local_path(&file_rel) {
             continue; // ignored/ephemeral — not deleted; blocks the collapse (→ degrade)
         }
         if !is_file_fully_synced(state, root, &file_rel)? {
--- a/apps/desktop/src-tauri/src/commands.rs
+++ b/apps/desktop/src-tauri/src/commands.rs
@@ -19,6 +19,7 @@ use crate::dropbox_transfer::{download_remote_file_internal, hydrate_remote_fold
 use crate::error::AppResult;
 use crate::models::*;
 use crate::oauth_listener::{start_oauth_callback_listener, stop_oauth_listener};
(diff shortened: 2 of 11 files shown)


mbsanchez (commented):
## Code review — DBSYNC-36 S1+S2 (Auth by Claude - Code-Reviewer)

**Verdict: APPROVE** — no blockers, no majors. Tests (20 path_util + db round-trip) green, clippy clean, tsc clean.

### Data-safety (the important one): SAFE
Folding user ignore-globs into `is_ignored_local_path` does **not** cause spurious remote deletes.
The scan applies the predicate symmetrically — the "seen" insertion (`sync_pipeline.rs:438/499`) and
the deletion loops (`460/516`) both skip ignored paths, and `enqueue_targeted_deletions` only runs
past the `:260` filter + `:364` descendant guard. A newly-ignored, already-synced file is skipped on
both sides, so its Dropbox copy is preserved. `remove_local_file` never touches the remote.

### Findings (all minor/nit — fine to merge, worth a follow-up)
1. **minor** `remote_index.rs:287` — the ignore predicate is not applied on the remote-download
   reconcile, so an ignored file with a remote row can still be downloaded/overwritten from another
   device's edit. Not Dropbox data-loss, and moot for local-only junk, but the ignore is only
   half-applied vs. the UI's "skipped on this device only." Apply the predicate in the reconcile loop
   or tighten the copy. Can defer past S2.
2. **minor** `lib.rs:329/343` — `cleanup_stale_upload_state` runs *before* `set_user_ignore_globs`.
   That ordering is what keeps it safe (cleanup sees empty user-globs on a fresh process, so it never
   strips index rows of real files the user just ignored). Please pin it with a comment, or exclude
   user-globs from `cleanup_stale_upload_state`'s predicate — that cleanup is semantically for
   temp/junk "that should never have been tracked," not user-ignored real files.
3. **nit** `path_util.rs:218` — a bare `*` matches nothing (good: won't nuke all sync), but silently
   no-ops a pattern users might expect to mean "everything." Consider validation feedback later.
4. **nit** ASCII-only case folding; non-ASCII basenames won't fold. Extension suffixes still match.
5. **nit** `path_util.rs:452` test mutates a process-wide static under parallel tests; safe today but a
   panic before reset would leak `*.log` to other tests. Latent flake.

### Verified clean
IPC boundary typed (`Result<_,String>`, `?` on `AppResult`, no unwrap/panic); RwLock fails open on
poison (no panic); empty/whitespace patterns can't match-all; `\`->`/` normalized; S1 tray reuses the
`setup` window (no duplicate, focus works), quit/selective-sync untouched; frontend hook mirrors
`useSelectiveSync`; camelCase serde matches the TS type.

mbsanchez:
Review follow-up: **minor #2 resolved** in 18a04b0 — added `is_builtin_ignored_local_path` (OS junk + editor temps, no user globs) and switched `cleanup_stale_upload_state` to it, removing the load-order fragility (a user-ignored real file can no longer have its index row stripped by cleanup). New test covers the decoupling. 135 lib tests pass, clippy clean.

Deferred as follow-ups (non-blocking): minor #1 (apply ignore on remote-download reconcile / tighten UI copy) + the nits.

PR #63, opened 2026-07-16
feat(settings): disconnect / sign-out from Dropbox (#DBSYNC-36)

mbsanchez:
## Summary
DBSYNC-36 slice **S3** (security-reviewed). Adds a **Disconnect** action to the Settings panel's Dropbox-account section, behind a confirmation, that fully signs the user out and returns the app to onboarding — **without touching the user's files on Dropbox**.

`disconnect_dropbox` runs in a deliberate, data-safe order:
1. Best-effort `POST /2/auth/token/revoke` (needs the token → **before** clearing; a network/API failure is logged and never blocks local sign-out).
2. Clear the in-memory token cache.
3. `SecureStore::clear_session()` — deletes every credential `has_stored_credentials` checks (chunked session keys, expires, legacy session blob, legacy access token). **Log-and-continue** on keyring error so local state is always cleared.
4. `reset_sync_state` + new `clear_sync_folder` → fresh onboarding. Local prefs (include/exclude/ignore CSV) **kept**.
5. Clear the sync engine's in-memory tracked path (new `clear_tracked_path` → `None`).

After disconnect, `has_stored_credentials` is false and `get_sync_folder` is None, so the existing startup-requirements refresh auto-routes to onboarding. Never logs token values.

## Stack
desktop-tauri

## Jira Ticket
DBSYNC-36 (S3 of 4; S1+S2 merged in #62; S4 startup-at-login follows)

## Test Plan
- [x] `cargo test --lib` — 136 pass (new: disconnect clears sync_folder but keeps local prefs), `cargo clippy --lib` clean
- [x] `tsc` + `vite build` pass
- [ ] **Manual/HITL (deferred to reviewer/PO)**: sign out end-to-end — session gone after app restart (no auto-resurrect), UI back at onboarding, best-effort revoke observed, user's Dropbox files untouched. Keyring clear is not headless-testable.

## Security notes
- Revoke-before-clear ordering; log-and-continue keyring clear (never a half-cleared state that leaves the session but drops the cache); token values never logged.
- Disconnect clears credentials + sync state only — never deletes synced files.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/auth_session.rs
+++ b/apps/desktop/src-tauri/src/auth_session.rs
@@ -240,6 +240,37 @@ pub(crate) fn verify_dropbox_token_internal(state: &AppState) -> AppResult<bool>
     Ok(true)
 }
 
+/// Best-effort revoke of the current Dropbox access token (DBSYNC-36 disconnect).
+/// MUST be called before the local session is cleared — it needs the still-valid
+/// token to identify which grant to revoke. Never panics, never logs the token
+/// value; a failure (no stored session, network error, non-2xx) only produces a
+/// warning log since local sign-out proceeds either way.
+pub(crate) fn revoke_dropbox_token_best_effort(state: &AppState) {
+    let session = match state.secure_store.get_session() {
+        Ok(s) => s,
+        Err(_) => return, // nothing to revoke
+    };
+
+    let result = state
+        .http_client
+        .post("https://api.dropboxapi.com/2/auth/token/revoke")
+        .bearer_auth(session.access_token.as_str())
+        .json(&serde_json::json!({}))
+        .send();
+
+    match result {
+        Ok(resp) if resp.status().is_success() => {
+            tracing::info!("dropbox access token revoked");
+        }
+        Ok(resp) => {
+            tracing::warn!(status = %resp.status(), "dropbox token revoke returned non-success");
+        }
+        Err(e) => {
+            tracing::warn!(error = %e, "dropbox token revoke request failed");
+        }
+    }
+}
+
 pub(crate) fn current_tray_status_label(state: &AppState) -> String {
     if state.sync_running.load(Ordering::Acquire) {
         return "Syncing".to_string();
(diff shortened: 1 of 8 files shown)


mbsanchez (commented):
## APPROVE — DBSYNC-36 S3 (disconnect / sign-out) — Auth by Claude - Code-Reviewer

Security-critical review done. The clear path is complete and correct; local tests green (136 pass incl. the new clear-path test), clippy + tsc clean.

**Security & data-safety verified**
- **Credential completeness ✅** `clear_session` deletes the exact union of keys any read path touches: chunked `SESSION_ACCESS`/`SESSION_REFRESH` (via `clear_chunked_key`, which sweeps `base.0..base.63` — all chunks, not just chunk 0), `SESSION_EXPIRES`, legacy session blob, and legacy `dropbox-access-token`. Matches everything `has_stored_credentials` / `get_session` / `get_token` read → no resurrection on restart.
- **Token never logged ✅** revoke + clear log only status/error; `bearer_auth` doesn't log.
- **Ordering ✅** revoke (token still valid) → cache → keyring → DB; DB clears run regardless of the best-effort keyring result.
- **Data-safety ✅** `reset_sync_state` + `clear_sync_folder` are SQL-only, no filesystem ops — synced files untouched; include/exclude/ignore prefs preserved (proven by the new test).
- **IPC ✅** no args, `Result<(),String>`, no unwrap/panic.
- **Frontend ✅** confirm gate, button disabled while in-flight, `refreshStartupRequirements()` in `finally` routes back to onboarding, no token exposure; button only shown when fully connected.

**Non-blocking follow-ups**
- *(minor)* `secure_store.rs` `clear_session` always returns `Ok(())`, swallowing every `delete_credential` error — the `if let Err(e)` warn branch in `disconnect_dropbox` is dead code, and a genuine keyring delete failure would silently leave credentials. Consider returning the last real error (treating `NoEntry` as success), or drop the `Result` and the dead handling at the call site.
- *(nit)* The `disconnect_dropbox` doc comment overstates the cache→keyring ordering guarantee (the open window is a reader repopulating the cache *from* the keyring; it self-heals via the 401 path). Soften the wording.
- *(nit)* Blocking revoke on the command thread is bounded to the 30s client timeout — fine, just an up-to-30s "Disconnecting…" stall when offline.

None block merge.

mbsanchez:
Review follow-up: **minor resolved** in d9420be — `clear_session` now surfaces a genuine keyring delete failure (missing entry still = success) instead of always returning Ok, so the warn branch in `disconnect_dropbox` is meaningful and a real failure no longer silently leaves credentials. Softened the ordering doc-comment (nit). 136 lib tests pass, clippy clean. Nits: the up-to-30s offline revoke stall left as-is (bounded, best-effort).

PR #65, opened 2026-07-16
fix(sync): mass-deletion circuit breaker (safety net) (#DBSYNC-64)

mbsanchez:
## Summary
**Emergency data-safety net** (part 1 of #DBSYNC-64) after a data-loss incident where the local scan propagated **612 remote deletes + 334 local deletes** of real files that had vanished from disk via a failing CfAPI placeholder op — no existing guard caught the batch.

The full local scan (`scan_local_changes_internal`) now collects FILE and FOLDER deletion candidates **before enqueuing any**, and if a single pass would delete a suspicious number:
- **>= 200 absolute**, OR
- **>= 10% of the tracked set** (above a 25-item floor)

…it **BLOCKS the whole batch**: nothing enqueued, index rows kept, a **sticky error** surfaced, and an **OS notification** fired. No sync client should delete hundreds of files from one reconcile without a sanity check — this would have prevented the incident **regardless of the CfAPI root cause**.

To proceed with a genuine bulk delete, the user explicitly confirms via the new **`confirm_pending_deletions`** command (one-shot override consumed by the next scan).

## Stack
desktop-tauri

## Jira Ticket
DBSYNC-64 (data-loss). This is **part 1** (the safety net). Remaining on the ticket: the remote-reconcile → local-delete direction needs the same breaker, plus the deeper CfAPI root cause (atomic dehydration; extend the delete-suppression guard to native CfAPI placeholders).

## Test Plan
- [x] `cargo test --lib` 138 pass (new: `is_mass_deletion` thresholds + `scan_blocks_mass_deletion_until_overridden` — 30 tracked files gone → blocked, index kept, error set; override → batch let through), `cargo clippy --lib` clean
- [ ] Live (test account, disposable folder): induce a bulk local absence → confirm sync pauses with the alert and nothing is deleted; confirm the override lets a genuine bulk delete through.

## Notes
- Threshold rationale: the incident deleted ~10% of ~9100 tracked files; both the 10% fraction and the 200 absolute rules catch it.
- The block sets a sticky engine error (tray shows Error) + notifies; the frontend confirm button is a follow-up (the `confirm_pending_deletions` command exists now).

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/commands.rs
+++ b/apps/desktop/src-tauri/src/commands.rs
@@ -464,6 +464,18 @@ pub fn resolve_conflict(
     Ok(())
 }
 
+/// User override for the mass-deletion circuit breaker (DBSYNC-64): authorize the
+/// NEXT scan to propagate the deletions it blocked. One-shot — the flag is consumed
+/// by the next scan pass. Only call this after the user has reviewed the blocked
+/// deletion and confirmed it is an intentional bulk delete.
+#[tauri::command]
+pub fn confirm_pending_deletions(state: tauri::State<AppState>) -> Result<(), String> {
+    state
+        .db
+        .set_app_config("mass_delete_override_once", "1")?;
+    Ok(())
+}
+
 #[tauri::command]
 pub fn scan_local_changes(state: tauri::State<AppState>) -> Result<usize, String> {
     scan_local_changes_internal(state.inner()).map_err(String::from)
--- a/apps/desktop/src-tauri/src/lib.rs
+++ b/apps/desktop/src-tauri/src/lib.rs
@@ -394,6 +394,7 @@ pub fn run() {
             commands::get_sync_dashboard,
             commands::retry_failed_jobs,
             commands::resolve_conflict,
+            commands::confirm_pending_deletions,
             commands::scan_local_changes,
             commands::process_sync_queue,
             commands::sync_tick,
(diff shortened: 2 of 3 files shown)


mbsanchez (commented):
## Code review — DBSYNC-64 part 1 (mass-delete circuit breaker) — Auth by Claude - Code-Reviewer

Initial verdict: **REQUEST_CHANGES**. The data-safety core is correct (refactor preserves the original file/folder deletion filters + `!walk_had_error` gating exactly; `known` correctly borrowed not moved; block keeps all index rows; the incident case 946 ≥ 200 trips; small deletes pass; override is one-shot; tests + clippy green). Three findings to address before merge:

1. **The pause signal was wiped within the same scan (Warning).** `block_mass_deletion` set `engine.set_last_error`, but the scan's trailing `refresh_queue_depth_internal` recomputes `last_error` from failed jobs only — the block enqueues none → the error was cleared before the scan returned (in production the remote-refresh succeeds and reaches that line). User only saw a transient toast per tick.
2. **The test passed for the wrong reason (Warning).** It asserted `last_error.is_some()` only because the no-token remote refresh bailed early — a path that doesn't happen in production.
3. **Override could linger (Warning).** `consume_mass_delete_override` was only reached when `is_mass_deletion` was already true (`&&` short-circuit); a confirm followed by a non-mass pass left `"1"` set, which could silently authorize a later unrelated mass deletion.

### Resolved in d1adc28
1. The pause is now a **durable** `app_config` flag `mass_delete_blocked`; `refresh_queue_depth_internal` surfaces it with precedence over a job error, so it persists every tick until cleared (cleared on a normal/overridden deletion pass).
2. The test now asserts the **durable flag** (set on block, cleared on override) — reflects production, not the no-token bail.
3. The override is now **consumed on any pass that has deletion candidates**, so a stale confirmation can't survive to a later batch.

Net: **APPROVE** after d1adc28. `cargo test --lib` 138 pass, clippy clean. (Minor deferred: `tracked_count` counts `.cloudsc`/ignored rows in the denominator — negligible given the absolute cap.)

PR #66, opened 2026-07-16
fix(cfapi): atomic dehydration — never lose the file if the placeholder op fails (#DBSYNC-64)

mbsanchez:
## Summary
**Part 2 root-cause fix** for #DBSYNC-64 — removes the **trigger** of the data-loss incident.

`dehydrate_one_file_cfapi` did: untrack → `fs::remove_file` (delete the real file) → `create_dehydrated_placeholder`. When the CfAPI create failed (observed live: `CfCreatePlaceholders` → `HRESULT(0x8007017C)`), the real file was already **gone with no placeholder** → permanent local data loss (and a fragile state that fed the mass-delete cascade).

Now **atomic** via move-aside + restore-on-failure. New platform-agnostic, unit-testable helper `dehydrate_file_atomic(state, rel, abs, row, create_placeholder)`:
- untrack first (keeps the DBSYNC-45 remove→create guarantee);
- **MOVE** the real file aside to `<name>.dbsync-dehydrate.tmp` (the `.tmp` suffix makes the scan ignore it during the swap) instead of deleting it;
- create the placeholder at the freed path;
- success → drop the aside copy, re-track with the original hash;
- **FAILURE → rename the aside file back (RESTORE)**, re-track, return `Err` (CRITICAL log if the restore rename itself fails).

A failed CfAPI op can no longer lose the file.

## Stack
desktop-tauri

## Jira Ticket
DBSYNC-64 (part 2 — atomic dehydration). Remaining on the ticket: event-driven NOTIFY_DELETE deletes (replace scan inference), remote-reconcile-direction breaker, frontend confirm button.

## Test Plan
- [x] `cargo test --lib` 140 pass — new `dehydrate_file_atomic_failure_restores_original_file` (data-loss regression guard: failed create leaves the original intact + re-tracked) + `dehydrate_file_atomic_success_drops_local_copy_and_retracks`; `cargo clippy --lib` clean
- [ ] Live (test account, disposable folder): free space on a folder; if a CfAPI create fails, confirm the real file survives (restored) instead of vanishing. The real CfAPI call isn't unit-testable (needs a live sync root).

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/cloud_filter.rs
+++ b/apps/desktop/src-tauri/src/cloud_filter.rs
@@ -568,7 +568,10 @@ fn is_connected() -> bool {
 /// handler (`FileIdentity` carries `remote_path`; the handler also re-derives it
 /// from the local path). Marked in-sync so no upload is triggered. The caller MUST
 /// ensure `parent\name` does not already exist (`CfCreatePlaceholders` fails
-/// otherwise). Returns whether the placeholder was created.
+/// otherwise) — callers that are replacing a real file should MOVE it aside first
+/// (see `cloudsc_ops::dehydrate_file_atomic`, DBSYNC-64) rather than deleting it, so
+/// a failed placeholder create never loses data. Returns whether the placeholder
+/// was created.
 pub(crate) fn create_dehydrated_placeholder(
     parent: &Path,
     name: &str,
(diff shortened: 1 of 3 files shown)


mbsanchez (commented):
**APPROVE** — data-safety property holds: the local file is never lost on any path (traced: move-aside-fails / create-succeeds / create-fails+restore-succeeds / the critical create-fails+restore-fails where the full file survives at `.dbsync-dehydrate.tmp`). DBSYNC-45 untrack-first preserved; `is_editor_temp_path` ignores the `.tmp` aside so a concurrent scan can't upload it. 140 lib tests, clippy clean.

One should-fix from review, now resolved in the latest commit: the re-track in the critical restore-failed sub-branch ran unconditionally, which would have let the next scan spuriously delete the still-intact cloud copy — it's now gated on restore success (row left untracked when restore fails). Follow-up (own item): a startup sweep to auto-recover/log stray `.dbsync-dehydrate.tmp` leftovers from a crash mid-op.

PR #67, opened 2026-07-16
fix(sync): mass-delete circuit breaker on the remote-sweep local-delete direction (#DBSYNC-64)

mbsanchez:
## Summary
Completes **part 1** of the DBSYNC-64 breaker — the **remote→local** direction (the 334 `local_delete`s in the incident).

The remote sweep (`refresh_remote_index_and_enqueue_downloads_internal`) enqueues `local_delete` for tracked local files absent from the remote snapshot; an incomplete/wrong snapshot mass-deletes local files. Now gated by the same breaker as the local scan.

- The 3 breaker primitives (`block_mass_deletion`, `consume_mass_delete_override`, `clear_mass_delete_blocked`) are made `pub(crate)` and **reused** (no duplication; same durable pause flag + one-shot override).
- New pure, unit-testable helper `remote_sweep_delete_candidates` → absent rels + the count that would actually `local_delete` (remote `content_hash == local hash`; a diverged file is a conflict, not a delete). If it trips `is_mass_deletion` and there's no override → the whole absent batch is skipped this sweep (nothing deleted) + pause/notify; present files still download.
- **Delta/longpoll path is deliberately NOT gated** — `.tag=="deleted"` entries are authoritative real deletes, not absence-inference.
- Confirmed `fetch_all_remote_file_metadata` fully paginates and errors on any page failure (never a silent partial snapshot).

## Stack
desktop-tauri

## Jira Ticket
DBSYNC-64 (part 1 completion — both scan directions now covered). Follow-up: `seed_remote_delta_cursor` does the same absence-inference and should get the same guard.

## Test Plan
- [x] `cargo test --lib` 144 pass (new: candidates-flagged-as-mass + diverged/present/pending excluded), `cargo clippy --lib` clean

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/remote_index.rs
+++ b/apps/desktop/src-tauri/src/remote_index.rs
@@ -253,50 +253,14 @@ pub(crate) fn refresh_remote_index_and_enqueue_downloads_internal(
     }
 
     let (remote_by_path, _cursor) = fetch_all_remote_file_metadata(state)?;
-
     let pending_targets = pending_job_targets(state)?;
 
-    let mut enqueued = 0usize;
-
-    // PRESENT files: reconcile immediately, same as before — never gated by the
-    // breaker (it only guards inferred deletions).
-    for local in &local_files {
-        let rel = &local.relative_path;
-        if rel.ends_with(".cloudsc") || pending_targets.contains(rel) {
-            continue;
-        }
-        if let Some(remote_meta) = remote_by_path.get(&normalize_dropbox_path(rel)?.to_lowercase())
-        {
-            enqueued += reconcile_remote_present(state, rel, remote_meta)?;
-        }
-    }
-
-    // ABSENT files + mass-deletion circuit breaker, remote→local direction
-    // (DBSYNC-64 extension). A full sweep INFERS "deleted remotely" from a path's
-    // absence in this snapshot; unlike the cursor-delta's explicit `.tag=="deleted"`
-    // entries (apply_remote_delta, always authoritative and never gated here), a
-    // wrong/incomplete snapshot would misclassify still-present files as absent and
-    // mass-delete local copies. So: compute how many of the absent files would
-    // actually enqueue a `local_delete` (an unmodified local copy — a diverged one
-    // becomes a conflict, never a delete, so it doesn't count), and block the whole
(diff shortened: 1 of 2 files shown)


mbsanchez (commented):
**Self-review — REQUEST_CHANGES** (data-safety). Sweep gating is correct; 144 lib tests pass, clippy clean. Applying fixes:

**Verified good**
- `remote_sweep_delete_candidates` counts a delete-candidate on *exactly* `reconcile_remote_absent`'s condition (prev remote row ∧ local row ∧ `local.hash == prev.content_hash`); .cloudsc/pending/present skips match. No over/under-count.
- PRESENT files reconcile before the gate — never blocked. Blocked absent batch skipped wholesale; safe, self-heals next tick.
- `apply_remote_delta` (`.tag=="deleted"`) stays ungated — authoritative deletes must propagate. `fetch_all_remote_file_metadata` errors on any page failure (no silent partial).

**🔴 Blocking (being fixed): `seed_remote_delta_cursor` is the same inferred mass-delete, ungated** (remote_index.rs:437-463) — reachable with a full local index via the cursor-**reset** path (`apply_remote_delta` re-snapshots through `seed_...` without clearing local_file_index). Gating it with the same reused helpers.

**🟡 (being fixed): durable pause flag clobber** — both directions wrote one `MASS_DELETE_BLOCKED_KEY`; the remote sweep runs after the local scan and a benign sweep cleared the pause the scan just set. Switching to per-direction keys OR'd for the UI.

**🟡 (accepted): shared one-shot override** consumed by whichever direction runs first — annoyance, not unsafe; left as-is.

**🟢 nit (being fixed):** drop the redundant `get_local_file` re-read (use `local.hash`).

mbsanchez:
Review findings resolved in the latest commit: **A (blocking)** — `seed_remote_delta_cursor` now routes through the shared gated `reconcile_remote_snapshot_with_breaker` (covers the cursor-reset re-snapshot path). **B** — per-direction pause keys (`mass_delete_blocked_scan`/`_remote` via a `MassDeleteSource` enum), no more cross-direction clobber. **D** — redundant `get_local_file` re-read dropped. **C** accepted as documented. 146 lib tests, clippy clean.

PR #68, opened 2026-07-17
feat(cfapi): event-driven deletes via NOTIFY_DELETE/RENAME (authoritative signal) (#DBSYNC-64)

mbsanchez:
## Summary
**Part 2 correctness fix (increment 1)** for #DBSYNC-64 — the real fix for the root cause. Drive remote deletes from CfAPI's **authoritative** "user deleted/renamed this placeholder" callbacks instead of only inferring deletions from the scan's "tracked file absent on disk" (which conflated real deletes, evictions, and glitches).

**Additive this round**: the scan + circuit breaker stay as a backstop (duplicate `delete` jobs dedup). Disabling scan-inference (so it's an offline-only fallback) is a follow-up **after live validation**.

- Register `NOTIFY_DELETE`/`_COMPLETION` + `NOTIFY_RENAME`/`_COMPLETION` on the connected provider. Verified against windows 0.61.3 source: `AckDelete=CF_OPERATION_PARAMETERS_0_7`, `AckRename=_0_6`, each `{Flags, CompletionStatus}`.
- `on_notify_delete/rename`: never veto, record `rel→is_dir` in module statics, **ACK promptly + unconditionally** (the platform waits for the ACK). Panic-firewalled like `on_fetch_data`.
- `on_*_completion`: resolve rel via a `CONNECTED_ROOT` static (no DB read from the ABI callback) → `enqueue_remote_delete_async` spawns a thread, enqueues the Dropbox `delete`, untracks the index, drains through the **normal** path (so `delete_suppressed_by_dehydration` still applies — additive, not a bypass).
- **Rename**: conservative — always `delete(source)`; a target still under the root is re-uploaded at the new path by the existing scan/watcher (a proper Dropbox move is a documented follow-up). Pure `classify_rename_target` helper + 4 unit tests.

Also fixes a **test-binary load failure** (`STATUS_ENTRYPOINT_NOT_FOUND` / 0xc0000139): taking the completion handlers as fn pointers made `refresh_tray_tooltip`'s call graph reachable → pulled `TaskDialogIndirect` (ComCtl32 v6), which the manifest-less `cargo test` harness can't load. Gated those tray-refresh calls behind `#[cfg(not(test))]` (same pattern as the notification-plugin gate).

## Stack
desktop-tauri

## Jira Ticket
DBSYNC-64 (part 2, increment 1).

## Test Plan
- [x] `cargo test --lib` 150 pass (4 new `classify_rename_target`), `cargo clippy --lib` clean, full clean rebuild verified
- [ ] **Live (test account, disposable folder)** — the CfAPI callbacks aren't unit-testable; validate: delete a placeholder → propagates to Dropbox via NOTIFY_DELETE completion; an eviction/dehydration does NOT propagate; a folder delete; a rename within/out of the root. Watch for callback pairing (in-flight maps are keyed by `rel`, not `TransferKey` — a rapid delete+recreate of the same rel could mis-pair; low risk with the scan backstop still active).

## Follow-ups
- Disable scan-inferred deletes when the provider is connected (make the scan an offline-only fallback).
- Proper Dropbox move on rename-within-root (preserve revision history).

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/cloud_filter.rs
+++ b/apps/desktop/src-tauri/src/cloud_filter.rs
@@ -42,14 +42,19 @@ use windows::Win32::Security::Authorization::ConvertSidToStringSidW;
 use windows::Win32::Security::{GetTokenInformation, TokenUser, TOKEN_QUERY, TOKEN_USER};
 use windows::Win32::Storage::CloudFilters::{
     CfConnectSyncRoot, CfConvertToPlaceholder, CfCreatePlaceholders, CfExecute,
-    CfHydratePlaceholder, CfRevertPlaceholder, CfSetInSyncState, CfSetPinState, CF_CALLBACK_INFO,
-    CF_CALLBACK_PARAMETERS, CF_CALLBACK_REGISTRATION, CF_CALLBACK_TYPE,
-    CF_CALLBACK_TYPE_CANCEL_FETCH_DATA, CF_CALLBACK_TYPE_FETCH_DATA, CF_CONNECTION_KEY,
+    CfHydratePlaceholder, CfRevertPlaceholder, CfSetInSyncState, CfSetPinState,
+    CF_CALLBACK_DELETE_FLAG_IS_DIRECTORY, CF_CALLBACK_INFO, CF_CALLBACK_PARAMETERS,
+    CF_CALLBACK_REGISTRATION, CF_CALLBACK_RENAME_FLAG_IS_DIRECTORY, CF_CALLBACK_TYPE,
+    CF_CALLBACK_TYPE_CANCEL_FETCH_DATA, CF_CALLBACK_TYPE_FETCH_DATA,
+    CF_CALLBACK_TYPE_NOTIFY_DELETE, CF_CALLBACK_TYPE_NOTIFY_DELETE_COMPLETION,
+    CF_CALLBACK_TYPE_NOTIFY_RENAME, CF_CALLBACK_TYPE_NOTIFY_RENAME_COMPLETION, CF_CONNECTION_KEY,
     CF_CONNECT_FLAGS, CF_CONNECT_FLAG_REQUIRE_FULL_FILE_PATH, CF_CONNECT_FLAG_REQUIRE_PROCESS_INFO,
     CF_CONVERT_FLAG_MARK_IN_SYNC, CF_CREATE_FLAG_NONE, CF_FS_METADATA, CF_HYDRATE_FLAG_NONE,
-    CF_IN_SYNC_STATE_IN_SYNC, CF_IN_SYNC_STATE_NOT_IN_SYNC, CF_OPERATION_INFO,
-    CF_OPERATION_PARAMETERS, CF_OPERATION_PARAMETERS_0, CF_OPERATION_PARAMETERS_0_0,
-    CF_OPERATION_TRANSFER_DATA_FLAG_NONE, CF_OPERATION_TYPE_TRANSFER_DATA,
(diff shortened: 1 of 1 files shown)


mbsanchez (commented):
**REQUEST_CHANGES** — ACK unions, panic firewall, off-thread DB all verified correct against windows-0.61.3. But the *policy* of which deletes propagate has a data-loss hole. Fixing:

**🔴 Critical — self-initiated placeholder deletes get propagated to Dropbox.** `on_notify_delete`/`on_notify_rename` never check `CF_CALLBACK_INFO.ProcessInfo.ProcessId`, though we connect with `CF_CONNECT_FLAG_REQUIRE_PROCESS_INFO` for exactly this. CfAPI fires NOTIFY_DELETE for **our own** `fs::remove_file`/`fs::remove_dir` (free-up-space collapse, conflict use-remote, local_delete) — those delete fully-synced placeholders whose content still lives on Dropbox → `delete_suppressed_by_dehydration` doesn't suppress (no per-file .cloudsc) → the Dropbox copies get deleted (folder remove_dir → recursive delete_v2). Reintroduces the incident class. Fix: if ProcessInfo is non-null and ProcessId == GetCurrentProcessId(), ACK but don't mark in-flight.

**🟠 High — rename source resolved inconsistently** (raw PCWSTR vs full_local_path's VolumeDosName prepend) → relpath_under fails → rename→delete never fires + PENDING_RENAMES_IN_FLIGHT leaks. Resolve source like the target.

**🟠 High — moving a dehydrated placeholder within root = data loss** (delete source, scan skips the placeholder at the new path → orphaned). For this increment: make within-root moves a no-op (defer to scan + a proper move_v2 follow-up); only outside-root = delete source.

**🟡 Minor:** in-flight entries leak when a completion never fires (cap/clear); one thread per completion is heavy under bulk (breaker backstops); the cfg(not(test)) tray gate is correct but fragile.

**✅ Verified good:** ACK union fields/ParamSize/StructSize; ACK unconditional & outside catch_unwind (no hang on panic); panic firewall on all 4 callbacks; no DB/network in the ABI callback; drain-time dehydration guard preserved; poison-safe mutexes, no deadlock/spin; classify_rename_target + tests. 150 tests, clippy clean.

mbsanchez:
Review findings resolved in db60e77:
- **🔴 C1** — `is_own_process(info)` (`ProcessInfo.ProcessId == GetCurrentProcessId()`, field verified from crate source): the app's own deletes/renames (free-up-space, conflict use-remote, local_delete) still ACK but are NOT propagated to Dropbox. Only foreign-process (Explorer) actions do.
- **🟠 H1** — shared `full_path_from(vol, normalized)` resolves the rename source with the VolumeDosName like the target (fixes the always-None source_rel → the leak + dead rename path).
- **🟠 H2** — within-root moves are now a no-op (deferred to scan; proper move_v2 is a follow-up); only out-of-root propagates a source delete.
- **🟡 M1** — no in-flight leaks (self-ops never inserted; every completion consumes).

154 lib tests, clippy clean.

**Live-validation / defense-in-depth notes** (deferred, scan backstop still active this increment): (a) confirm `ProcessInfo` is reliably non-null for genuine Explorer deletes (REQUIRE_PROCESS_INFO requests it) — a null currently falls through as 'foreign' (would propagate); given the scan backstops any missed user delete while a wrongly-propagated self-delete is the catastrophic direction, we may want to tighten null→don't-propagate once live behavior is confirmed. (b) callback pairing across delete+recreate. (c) directory-delete single-callback assumption.

PR #72, opened 2026-07-17
feat(#DBSYNC-65): plumb captured delete-rev end-to-end (slice 1/2)

mbsanchez:
## Summary
Slice 1 of DBSYNC-65 (closes #70). **Pure plumbing — zero behavior change.** Captures the Dropbox `rev` at delete-enqueue time and persists it on a new `sync_jobs.delete_parent_rev` column, threaded end-to-end to the drain. The rev is stored but **not yet enforced** (Slice 2 / #71 activates the `parent_rev` guard), so remote-delete behavior is byte-for-byte unchanged.

### Key changes
- **storage/db.rs**: additive migration `delete_parent_rev TEXT` (+ old-DB CHECK-rebuild literal for correctness); `SyncJobRow.delete_parent_rev`; both projections (`pick_next_due_job`, `list_recent_jobs`); new dedicated `enqueue_delete_job(target, parent_rev)` whose `ON CONFLICT ... DO UPDATE SET` includes `delete_parent_rev=excluded.delete_parent_rev` — the crux: a delete re-enqueued for an already-queued path (restore-then-redelete) refreshes the rev instead of keeping a stale one. Generic `enqueue_job` (28 call sites) left untouched.
- **sync_pipeline.rs**: `process_local_file_deletion` captures the remote rev before untrack; `process_known_folder_deletion` passes `None` (folders aren't keyed in `remote_file_index`). Drain arm untouched.
- **cloud_filter.rs**: `enqueue_remote_delete_async` captures the rev gated on `!is_dir` (falls back to `None` on lookup error, never aborts the enqueue).

## Stack
desktop-tauri (Rust backend only; frontend untouched)

## Jira Ticket
DBSYNC-65

## Test Plan
- [x] Automated tests pass (`cargo test --lib` → 161 passed, 0 failed; 6 new)
- [x] `cargo clippy -- -D warnings` clean
- [x] desktop-tauri: no behavior change this slice (parent_rev not sent); enqueue/drain plumbing exercised by unit tests

### New tests
- `file_deletion_captures_remote_rev_on_enqueue`
- `file_deletion_with_no_remote_index_row_enqueues_none_rev`
- `folder_deletion_always_enqueues_none_rev`
- `re_enqueuing_delete_refreshes_stale_parent_rev` (crux regression)
- `enqueue_delete_job_persists_parent_rev`
- `enqueue_delete_job_on_conflict_updates_rev_not_just_source_path`

## Notes
[...]

PR #73, opened 2026-07-17
feat(#DBSYNC-65): activate delete_v2 parent_rev guard (slice 2/2)

### LINKED ISSUE #71: DBSYNC-65 Slice 2: Activate the delete_v2 parent_rev guard
Vertical slice 2 of DBSYNC-65 (https://manuelbsan.atlassian.net/browse/DBSYNC-65).

**Type:** AFK · **Blocked by:** Slice 1 · **Scope:** apps/desktop/src-tauri/src/dropbox_transfer.rs, sync_pipeline.rs

Activates the guard: enforce the captured rev as a Dropbox `files/delete_v2` `parent_rev` precondition (server-authoritative — local rev-compare is NOT race-safe due to longpoll lag).

## Acceptance criteria
- [ ] delete_remote_file_internal accepts an optional parent_rev and, when present, includes it in the delete_v2 request body (files only); when absent, request identical to today (unit test on request construction).
- [ ] A parent_rev conflict response is classified as a no-op success and the job is dropped, in the same branch as the existing `not_found` handling (direct unit test mirroring the not_found test).
- [ ] Drain arm passes the job's stored rev to the delete function.
- [ ] A queued file delete whose remote rev changed is skipped (no destructive delete) and the job cleared (unit test via injected delete boundary).
- [ ] A queued file delete whose rev still matches deletes as before (regression test).
- [ ] Folder deletes (NULL rev) sent without parent_rev, behave as today; existing dehydration/registration-grace suppression still runs first (test).
- [ ] `cargo test --lib` green.

mbsanchez:
## Summary
Slice 2 of DBSYNC-65 (closes #71). **Activates the fix.** Enforces the captured rev (plumbed in Slice 1 / #72) as a Dropbox `files/delete_v2` `parent_rev` precondition for file deletes. A queued delete now only proceeds if the server's current rev still matches the rev the delete was authorized against. If the file was restored/changed on the server after the delete decision (the stale-delete-races-a-cloud-restore window), Dropbox rejects the delete, the client treats it as a no-op and drops the job — no re-deletion of restored content.

### Key changes
- **dropbox_transfer.rs**: pure `classify_delete_response(status_success, body) -> DeleteOutcome {Deleted, AlreadyGoneOrConflict, Error}` (unit-testable without HTTP mocking); `delete_remote_file_internal` gains `parent_rev: Option<&str>`, conditionally inserted into the `delete_v2` body (key omitted when None — never `null`); the failure branch treats `not_found` **and** `path_write/conflict` as a no-op + drop. Manual-QA-only doc comment added.
- **sync_pipeline.rs**: drain `"delete" =>` arm passes `job.delete_parent_rev.as_deref()`. `delete_suppressed_by_dehydration` still runs first (unchanged).

## Stack
desktop-tauri (Rust backend only)

## Jira Ticket
DBSYNC-65

## Test Plan
- [x] `cargo test --lib` → 166 passed, 0 failed (5 new)
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [ ] **Manual QA (required — confirm the real Dropbox error body):** sync a file → delete it locally so a delete job is queued with a non-null `delete_parent_rev` → restore the file from another Dropbox client (bumps rev) → let the queued delete drain → capture the raw `files/delete_v2` status+body. Confirm it matches the `path_write/conflict` assumption; if not, adjust the classifier match + add a regression test with the observed body. End state: the restored file must remain in Dropbox (not deleted), job dropped (not retried).

### New tests
- `classify_delete_response_success_is_deleted`
- `classify_delete_response_not_found_variants_are_noop`
- `classify_delete_response_path_write_conflict_is_noop` (synthetic body — reconstructed from files.stone, pending live confirmation)
- `classify_delete_response_unrelated_error_is_error`
- `classify_delete_response_malformed_or_empty_body_is_error`

## Notes
- **Open item**: Dropbox has no dedicated `DeleteError` tag for a `parent_rev` mismatch (verified against files.stone). `path_write/conflict` is a defensive match; under-match fails safe (loud retry, not data loss). The manual-QA step above must confirm/correct it before the ticket is fully closed.
- Depends on #72 (Slice 1, merged).

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/dropbox_transfer.rs
+++ b/apps/desktop/src-tauri/src/dropbox_transfer.rs
@@ -944,9 +944,10 @@ pub(crate) fn classify_delete_response(status_success: bool, body: &str) -> Dele
         return DeleteOutcome::AlreadyGoneOrConflict;
     }
     // parent_rev mismatch: Dropbox has no dedicated DeleteError tag for it
-    // (verified against files.stone). Best-available signal is path_write/conflict.
-    // Match defensively; confirm the exact live body via manual QA before closing.
-    if body.contains("path_write") && body.contains("conflict") {
+    // (verified against files.stone). Best-available signal is the contiguous
+    // `path_write/conflict` error_summary. Match defensively; confirm the exact
+    // live body via manual QA before closing.
+    if body.contains("path_write/conflict") {
         return DeleteOutcome::AlreadyGoneOrConflict;
     }
     DeleteOutcome::Error
--- a/apps/desktop/src-tauri/src/dropbox_transfer.rs
+++ b/apps/desktop/src-tauri/src/dropbox_transfer.rs
@@ -924,7 +924,42 @@ pub(crate) fn upload_local_file_internal(
     Ok(())
 }
 
-pub(crate) fn delete_remote_file_internal(state: &AppState, relative: &str) -> AppResult<()> {
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) enum DeleteOutcome {
+    Deleted,
+    AlreadyGoneOrConflict,
+    Error,
+}
+
+/// Classifies a `files/delete_v2` response. `AlreadyGoneOrConflict` means
+/// treat as a no-op and drop the job: the target is already gone
+/// (`not_found`) OR a `parent_rev` precondition failed because the file was
+/// restored/changed on the server (`path_write/conflict`). Anything else on a
+/// failure status is a real `Error`.
+pub(crate) fn classify_delete_response(status_success: bool, body: &str) -> DeleteOutcome {
+    if status_success {
+        return DeleteOutcome::Deleted;
+    }
+    if body.contains("path_lookup/not_found") || body.contains("path/not_found") {
+        return DeleteOutcome::AlreadyGoneOrConflict;
+    }
+    // parent_rev mismatch: Dropbox has no dedicated DeleteError tag for it
+    // (verified against files.stone). Best-available signal is path_write/conflict.
+    // Match defensively; confirm the exact live body via manual QA before closing.
+    if body.contains("path_write") && body.contains("conflict") {
+        return DeleteOutcome::AlreadyGoneOrConflict;
+    }
+    DeleteOutcome::Error
+}
+
+/// Performs a live `files/delete_v2` call against the Dropbox API; manual-QA-only
+/// for the network path (the response-classification logic is covered by the
+/// pure `classify_delete_response` tests below).
+pub(crate) fn delete_remote_file_internal(
+    state: &AppState,
+    relative: &str,
+    parent_rev: Option<&str>,
+) -> AppResult<()> {
     if relative.ends_with(".cloudsc") {
         return Ok(());
     }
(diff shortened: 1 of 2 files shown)


mbsanchez (commented):
## CTO Review — PR #73 (DBSYNC-65 Slice 2: activate parent_rev guard)

**Stack profile:** desktop-tauri / Rust backend only — applied the `rust` review rows (no unwrap in prod path, pure testable core, clippy `-D warnings`).

**Independent verification (my run, not reported):**
- `cargo test --lib` → **166 passed, 0 failed** (5 new classifier tests present & green).
- `cargo clippy --all-targets -- -D warnings` → **clean (exit 0)**.

### Verdict: APPROVED (with one non-blocking condition)
**Condition:** the mandated manual-QA must capture the real `files/delete_v2` `parent_rev`-mismatch body and confirm/correct the `path_write/conflict` match before DBSYNC-65 → Done. This does not block merge — see the data-safety rationale below.

### Why the unconfirmed body is an accepted risk, not a blocker
The actual data-loss protection is the **`parent_rev` precondition itself**: the Dropbox server refuses to delete the restored file regardless of how the client classifies the response. `classify_delete_response` only decides *drop vs. retry*, never *delete vs. keep*. If the real mismatch body differs from the defensive `path_write/conflict` guess, the response falls through to `Error` (HTTP 409 → not retryable per `error.rs:70`, so it surfaces once and stops — no infinite loop, no deletion). Under-match fails safe in the ticket's favor; the manual-QA only upgrades "loud one-shot failure" into "silent clean drop."

### Focus-area findings

1. **Classifier over-match (focus area 1) — NOT too loose. Definitive judgment: acceptable, minor tightening recommended.**
   The `body.contains("path_write") && body.contains("conflict")` AND-guard was checked against every enumerated hard failure: `no_write_permission`, `insufficient_space`, `disallowed_name`, `team_folder`, `too_many_write_operations`, `too_many_files`, `operation_suppressed`. None of their tag/error_summary bodies contain the substring `conflict`, and `too_many_write_operations` does not contain the literal `path_write`. So no legitimate hard failure is swallowed as a no-op. The only theoretical false positive is a genuine `path_write/conflict/{file,folder}` that isn't a rev mismatch — and per ticket intent (favor "don't delete restored content") swallowing that as a no-op is the safe choice.
   - **Minor (nit):** prefer the contiguous `body.contains("path_write/conflict")` (the `error_summary` format) over two independent `contains`. It is strictly more precise (eliminates any cross-field coincidence), self-documenting, and all 5 existing tests stay green (the synthetic body and error_summary both contain the contiguous form). Not required for approval.

2. **parent_rev omission when None (focus area 2) — CORRECT.** `body["parent_rev"] = json!(rev)` runs only inside the `if let Some(rev)` arm; the key is absent (never `null`) when None. Runtime `Value` mutation produces a flat `{"path":..,"parent_rev":..}`.

3. **Files-only + ordering (focus area 3) — CORRECT.** Folders pass `None` (`cloud_filter.rs:853` `is_dir → None`; test `folder_deletion_always_enqueues_none_rev`). Drain arm passes `job.delete_parent_rev.as_deref()` and `delete_suppressed_by_dehydration` still runs first (`sync_pipeline.rs:802` before `:806`).

4. **Ordinary case / no regression (focus area 4) — CORRECT.** Matching rev → 200 → `Deleted`; `not_found` no-op is preserved, now folded into the classifier.

5. **Rust idioms (focus area 5) — CLEAN.** No unwrap/expect on the prod path (`resp.text().unwrap_or_else(..)`). `DeleteOutcome` + `classify_delete_response` are `pub(crate)` and pure (no I/O). `AppError::Dropbox { status: u16, message: String }` fields match (`error.rs:20`). The new `info` log prints only `rel` — no token/PII.

6. **Tests (focus area 6) — MEANINGFUL.** 5 pure tests including the over-match guard (`unrelated_error_is_error` covers `no_write_permission` + `insufficient_space`) and fail-safe empty-body. Synthetic `path_write/conflict` body matches the files.stone shape and is explicitly flagged as unconfirmed. Acknowledged gap (no HTTP-mock e2e drain test) is consistent with repo convention.

**No blockers, no majors.** One minor/nit (tighten to contiguous `path_write/conflict`).

mbsanchez:
Addressed the review nit in decc405: the conflict match is now the contiguous `body.contains("path_write/conflict")` (error_summary substring) instead of `path_write` && `conflict` separately — strictly more precise, no cross-field false positive. All 5 classifier tests stay green; `cargo test --lib` 166 passed; clippy clean.

PR #76, opened 2026-07-18
fix(#DBSYNC-66): cloud-restore data-loss fix — grace window + delete completeness + coalescing + native-only materialization

### LINKED ISSUE #74: DBSYNC-66 Slice 1: Subtree-scoped materialization grace window (stop restore-churn re-deletes)
Vertical slice 1 of DBSYNC-66 (https://manuelbsan.atlassian.net/browse/DBSYNC-66) — **ROOT CAUSE fix, data loss.**

**Type:** AFK · **Blocked by:** none · **Scope:** apps/desktop/src-tauri/src/ (cloud_filter.rs, cloudsc_ops.rs, sync_pipeline.rs)

A cloud folder restore's placeholder materialization makes CfAPI/Explorer churn placeholders (delete-and-recreate), firing NOTIFY_DELETE that the event path propagates as real files/delete_v2 → re-deletes restored files. Fix: a subtree-scoped, time-windowed grace primitive mirroring `in_post_registration_grace()` (DBSYNC-62), armed by the restore driver at the subtree ROOT prefix BEFORE placeholders are recreated, checked at event time AND drain time. Prefix-scoped + short window so genuine deletes elsewhere still propagate. A time window is required because the existing placeholder-EXISTENCE check loses the race (delete drains ~26s before recreate).

## Acceptance criteria
- [ ] New prefix-keyed primitive (`OnceLock<Mutex<HashMap<String, Instant>>>`) with `mark_materialization(prefix)` + pure `in_materialization_grace(rel)`, mirroring the `in_post_registration_grace` trio; short window constant (~covers 26s churn lag).
- [ ] Pure-helper unit tests: in-window under armed prefix → true; expired → false; different/unarmed subtree → false; prefix-boundary safety (`foo/bar` vs `foo/barbaz`).
- [ ] Window armed at ALL materialization/restore drivers (create_remote_only_placeholder, materialize_remote_only_file_if_absent, folder-restore create_dir_all) by the restored subtree ROOT prefix, at the START of the op.
- [ ] Event-time suppression in on_notify_delete: a NOTIFY_DELETE under an active prefix is NOT marked in-flight → completion enqueues NO remote delete (test).
- [ ] Drain-time suppression: delete_suppressed_by_dehydration returns true under an active prefix, alongside the existing registration-grace check (test).
- [ ] Scoping guard: a delete for an UNRELATED path (no armed prefix) STILL propagates (test).
- [ ] Window expiry: after the window elapses, deletes under the formerly-armed prefix propagate normally (test).
- [ ] cargo test --lib green; clippy clean.
- [ ] MANDATORY live Windows repro: restore a whole folder from the cloud → ZERO files/delete_v2 for the restored subtree, all files survive; a concurrent genuine delete elsewhere still propagates.

mbsanchez:
## Summary
Complete fix for DBSYNC-66 (closes #74) — the cloud folder **restore → app re-deletes the restored files** data-loss bug, validated live on the test account (final state clean: no re-deletion, empty Dropbox trash, no empty-folder skeleton, native placeholders only, zero `.cloudsc`).

The branch grew past the original "Slice 1 grace window" as live testing surfaced the full root-cause set. All in one PR now:

### 1. Subtree-scoped materialization grace window (root cause)
`cloud_filter.rs`: prefix-keyed, time-windowed `MATERIALIZATION_IN_FLIGHT` (mirrors `in_post_registration_grace`). A restore's placeholder churn fires `NOTIFY_DELETE`; while the restored subtree is in its grace window those are recognized as churn (not user deletes) and suppressed at `on_notify_delete` + `delete_suppressed_by_dehydration`. Armed at the materialization drivers (parent-prefix), refreshed on each suppressed churn-delete, short window (120s) so genuine deletes elsewhere still propagate.

### 2. remote_index subtree cleanup on delete (kills "delete twice")
`dropbox_transfer.rs` + `db.rs`: on a genuine `delete_v2` success, clear the remote_index row for the path **and its whole subtree** (`remove_remote_subtree`, boundary-safe `LIKE ... ESCAPE`) so the sweep can't re-materialize just-deleted content. Matters for coalesced folder deletes.

### 3. Folder-delete completeness
`cloudsc_ops.rs` + `sync_pipeline.rs`: after a delete empties an ancestor, prune it locally + on Dropbox + from `known_folders` — confirming emptiness **against Dropbox** (never the stale local index), never `remove_dir_all`. No empty skeleton left on either side.

### 4. Coalesced folder deletes
`cloud_filter.rs`: debounce the burst of per-descendant `NOTIFY_DELETE` events and issue one recursive `delete_v2` per minimal root (`minimal_delete_roots`) instead of one delete per descendant. (Reduction is bounded by which folder events CfAPI actually fires — see follow-up ticket for the residual latency.)

### 5. Recursive materialization + no `.cloudsc` on Windows
`cloudsc_ops.rs`: recurse into a newly-created folder so a restored subtree materializes fully in one sweep; and on Windows **never** fall back to `.cloudsc` — defer the sweep until CfAPI is connected, and remove stale `.cloudsc` sidecars that would shadow the native placeholder.

## Stack
desktop-tauri (Rust backend only)

## Jira Ticket
DBSYNC-66

## Test Plan
- [x] `cargo test --lib` → 184 passed, 0 failed (new pure-helper tests: `materialization_grace_contains_*`, `minimal_delete_roots_*`, `ancestor_rels_*`, `remove_remote_subtree_*`, `classify_delete_response_*`)
- [x] `cargo clippy -- -D warnings` clean
- [x] **Live Windows validation**: delete a folder → restore from Dropbox trash → ends clean (files back, trash empty, no re-deletion, no skeleton, native placeholders, zero `.cloudsc`)

## Known follow-ups (tracked separately, NOT data-loss)
- DBSYNC-67: materialized folders stay in "syncing" overlay (cosmetic).
- Perf/UX ticket: delete latency (coalescing limited by CfAPI folder-event delivery + prune probes) and Sync Now not running a full sync (no remote materialization) / no feedback.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/cloud_filter.rs
+++ b/apps/desktop/src-tauri/src/cloud_filter.rs
@@ -695,7 +695,7 @@ unsafe extern "system" fn on_notify_delete(
             tracing::debug!(rel = %rel, is_dir, "cfapi notify-delete: self-initiated (our own process) — not tracked, no remote delete will be enqueued");
         } else if in_materialization_grace(&rel) {
             mark_materialization(&rel); // refresh — keeps active churn armed
-            tracing::debug!(rel = %rel, is_dir, "cfapi notify-delete: suppressed — under active materialization grace (restore churn, not a user delete)");
+            tracing::info!(rel = %rel, is_dir, "cfapi notify-delete: suppressed — under active materialization grace (restore churn, not a user delete)");
         } else {
             mark_user_delete_in_flight(&rel, is_dir);
             tracing::debug!(rel = %rel, is_dir, "cfapi notify-delete: user delete observed");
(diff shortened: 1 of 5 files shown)


mbsanchez (commented):
## CTO review — DBSYNC-66 (cloud-restore data-loss fix)

**Verdict: APPROVED** (no blockers). Stack: desktop-tauri (Rust). Independent verification on this branch: `cargo test --lib` → 184 passed / 0 failed; `cargo clippy --all-targets -- -D warnings` → clean. All `unwrap`/`expect` in the touched files are test-only.

Reviewed all 5 focus areas hard; the data-loss-critical logic is sound.

### Explicit judgments
- **`remove_remote_subtree` boundary-safety — SOUND.** Escape order is correct (`!`→`!!` first, then `%`/`_`), exact row uses the unescaped bind (`relative_path = ?1`), and the `{escaped}/%` pattern with `ESCAPE '!'` cannot match a sibling (`prefix-other`) because of the required `/` separator. A folder delete can never clear a sibling's rows. Unit test covers sibling collision + accented/spaced descendant.
- **Prune over-delete risk — LOW / acceptable.** Double-gated: local dir must be missing-or-empty (walk stops on any remaining sibling) AND Dropbox `list_folder` (`include_deleted:false`, so trash tombstones don't count) must report Empty/AlreadyGone. Uses `remove_dir` (not `remove_dir_all`), stops on any error, guards the sync root (`rel==""`), caps depth at 64. Empty classification is correct (empty page + `!has_more` → Empty; any entry or the conservative empty-page-but-has_more → HasContent). The only TOCTOU is remote-side between probe(Empty) and `delete_v2`, which is inherent to recursive folder delete and negligible (the ancestor was just emptied by the user's own delete). Sequential drain removes the local TOCTOU.
- **Coalescing over-delete + debounce event-loss.** `minimal_delete_roots` keeps a rel iff no other batch rel is a boundary-safe ancestor — correct and tested. A recursive `delete_v2` on a folder root does delete the whole Dropbox subtree incl. un-materialized cloud-only descendants — this is correct folder-delete semantics and gated by the grace window + `is_own_process` (suppressed deletes never enter the buffer, verified: they're never marked in-flight so completion enqueues nothing). Debounce worker exit/reclaim logic is correct: the `store(false)` → `more_pending` re-check → `compare_exchange` reclaim closes the stop/push race, the single `DELETE_FLUSH_RUNNING` gate prevents double-workers, and an unavailable `app_state` drops the batch to the (documented) scan backstop. No spin/hang/deadlock found.

### Findings (all non-blocking)
- **minor — `cloud_filter.rs` `run_delete_flush_worker` (~L982-1033):** no panic firewall around `flush_pending_deletes`. A panic in the worker thread leaves `DELETE_FLUSH_RUNNING=true` permanently, silently disabling event-driven delete coalescing for the rest of the session (scan backstop still covers actual deletes). Suggest wrapping the flush in `catch_unwind` or resetting the flag via a drop-guard.
- **minor — `cloudsc_ops.rs` (~L69-73):** on Windows with no package identity, `if !cfapi_active { return Ok(0) }` means cloud-only content never materializes and never falls back to `.cloudsc`. Correct for the shipped packaged app; would starve a non-packaged/dev Windows build. Confirm packaging is always present in shipped builds (it is per the sparse-package design) — noting for awareness.
- **consideration — event-delete path bypasses the mass-delete circuit breaker (`flush_pending_deletes` → drain):** a folder-level `NOTIFY_DELETE` that slipped past grace/`is_own_process` would recursively wipe un-materialized cloud content with no confirm-deletions prompt. Low risk given the two gates; already tracked in project notes as the "event path bypasses the breaker" item.
- **nit — grace window (`materialization_grace_contains`):** a genuine user delete *under a just-restored subtree* within 120s is swallowed, and refresh-on-suppress extends the window while churn continues. Documented accepted trade-off; fine.

### Interactions verified
`remove_remote_subtree` (in `delete_remote_file_internal`) + per-job `prune_empty_deleted_ancestors` + DBSYNC-65 parent_rev (files capture rev, folder roots `None`) compose without double-work or contradiction. Grace window is armed BEFORE placeholder creation at all 3 drivers; map is bounded by prune-on-insert.

**Next step: QA.**

PR #77, opened 2026-07-19
fix(#DBSYNC-67): convert folders to in-sync placeholders while empty (fixes stuck 'syncing' overlay)

mbsanchez:
## Summary
Fixes DBSYNC-67 — materialized folders stayed on the Explorer **"syncing"** glyph even though every child was a correct cloud-only placeholder.

### Root cause (confirmed live via HRESULT logging)
`CfConvertToPlaceholder` only succeeds on an **EMPTY** directory. Once a folder holds placeholder children it returns `0x8007017C` ("cloud operation not valid"). The old approach created folders as plain dirs (`create_dir_all`) and tried to convert them **later** — by which point they were already populated — so the convert always failed, the folder was never a placeholder, and `CfSetInSyncState` was a no-op on it. Under `PopulationPolicy=AlwaysFull` that renders the folder as "syncing".

### Fix
Convert each folder to an in-sync placeholder **at create time**, in the materialization sweep, right after `create_dir_all` while it is still empty and **before** the recursion materializes its children.

- **cloud_filter.rs**: `apply_folder_state` is now `pub(crate)`; returns `true` so the once-per-session `APPLIED_FOLDERS` cache never retries a permanently-failing convert on an already-populated folder (avoids a per-sweep storm); convert failures log at debug. New `notify_shell_folders_changed` (`SHChangeNotify` / `SHCNE_UPDATEITEM`) nudges Explorer to re-render folder overlays after a sweep materialized new placeholders (Explorer caches the overlay otherwise).
- **cloudsc_ops.rs**: call `apply_folder_state` on the just-created empty folder before recursing; `SHChangeNotify` known folders at the end of the sweep when `created > 0`.
- **Cargo.toml**: add windows `Win32_UI_Shell` feature.

## Stack
desktop-tauri (Rust backend only; Windows CfAPI path)

## Jira Ticket
DBSYNC-67

## Test Plan
- [x] `cargo test --lib` → 184 passed, 0 failed
- [x] `cargo clippy -- -D warnings` clean
- [x] **Live Windows**: delete + restore a folder → freshly materialized folders (UNET/Tutor etc.) become `Directory, ReparsePoint` (placeholders) and Explorer shows the **cloud icon**, not "syncing". (Pre-existing already-populated folders are only fixed on their next re-materialization — convert-while-empty applies to newly created folders.)

## Notes
- Cosmetic overlay fix; no data-loss / sync-correctness change.
- Follow-ups still open: DBSYNC-68 (Sync Now completeness + delete latency), DBSYNC-69 (longpoll / restore-materialization latency).

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/Cargo.toml
+++ b/apps/desktop/src-tauri/Cargo.toml
@@ -65,4 +65,5 @@ windows = { version = "0.61", features = [
     "Win32_System_CorrelationVector",
     "Win32_System_IO",
     "Win32_System_Threading",
+    "Win32_UI_Shell",
 ] }
--- a/apps/desktop/src-tauri/src/cloud_filter.rs
+++ b/apps/desktop/src-tauri/src/cloud_filter.rs
@@ -65,6 +65,7 @@ use windows::Win32::Storage::FileSystem::{
 };
 use windows::Win32::System::Com::{CoInitializeEx, COINIT_MULTITHREADED};
 use windows::Win32::System::Threading::{GetCurrentProcess, GetCurrentProcessId, OpenProcessToken};
+use windows::Win32::UI::Shell::{SHChangeNotify, SHCNE_UPDATEITEM, SHCNF_PATHW};
 
 use crate::error::{AppError, AppResult};
 
@@ -269,7 +270,7 @@ pub(crate) fn sync_folder_states(sync_folder: Option<&str>, folder_rels: &[Strin
 /// Convert a directory to an in-sync placeholder (no pin) so the shell renders its
 /// aggregate status. Under PopulationPolicy=AlwaysFull this never hides children.
 /// Fails soft.
-fn apply_folder_state(abs: &Path) -> bool {
+pub(crate) fn apply_folder_state(abs: &Path) -> bool {
     let path = to_wide(&abs.to_string_lossy());
     unsafe {
         let handle = match CreateFileW(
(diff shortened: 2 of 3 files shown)


mbsanchez (commented):
## CTO Review — APPROVED

**Stack:** desktop-tauri (Rust/Tauri backend, Windows CfAPI path).

**Independent verification (this machine):**
- `cargo test --lib` → 184 passed, 0 failed, 0 ignored.
- `cargo clippy --all-targets -- -D warnings` → clean (new `Win32_UI_Shell` feature triggered the expected windows-crate recompile).

### Focus-area findings
1. **Convert-while-empty ordering — correct.** `cloudsc_ops.rs:214-227`: `create_dir_all` → `upsert_known_folder` → `apply_folder_state(&target_path)` (still empty) → recursion at 233. The convert runs before any child is materialized, so it never hits `0x8007017C`. This is exactly the fix.
2. **No retry-storm regression — correct.** `apply_folder_state` now returns `true` unconditionally, so `APPLIED_FOLDERS` in `sync_folder_states` caches once per session. A convert failure on a populated folder is permanent; caching it is right and masks nothing harmful (`CfSetInSyncState` still runs first-time). Cosmetic path only.
3. **SHChangeNotify safety — sound.** `to_wide` NUL-terminates (`.chain(once(0))`); the `wide` Vec outlives the call; `SHCNF_PATHW` reads it as PCWSTR without retaining. SAFETY comment present. `cloud_filter` is `#[cfg(windows)]` at the module level (lib.rs:3) and all callers are cfg-gated → non-Windows still compiles. Iterating all known folders per sweep is bounded and cheap; acceptable.
4. **Rust idioms / panic safety — good.** No `unwrap`/`expect` in the new path; `unsafe` blocks documented; no pinning (`CfSetPinState` remains file-only at cloud_filter.rs:1759 — folders are never pinned, so children aren't force-hydrated); no secrets logged (debug-level, path only).
5. **DBSYNC-66 interaction — intact.** `mark_materialization(&relative)` (cloudsc_ops.rs:213) arms the grace window BEFORE `create_dir_all`/`apply_folder_state`, so any NOTIFY_DELETE churn from converting the folder to a placeholder is covered. Inserting the convert between create and recurse does not disturb the grace window or the coalescing.

### Nits (non-blocking)
- `cloudsc_ops.rs:226` — the `#[cfg(windows)]` is redundant (already inside the cfg-gated `if cfapi_active` block at line 183), but it matches the existing style of line 212/213 and is harmless.
- `notify_shell_folders_changed` refreshes every known folder when `created > 0` rather than only the folders that actually changed. Bounded/cheap today; could be scoped to changed folders later if the known-folder set grows large.

**Verdict: APPROVED.** No blockers. Cosmetic overlay fix, no data-loss/sync-correctness surface touched. Next: QA.

_(CTO review by Claude — architect-reviewer)_

PR #78, opened 2026-07-19
fix(#DBSYNC-68): Sync Now runs the full sync cycle + honest feedback

mbsanchez:
## Summary
- **Sync Now now runs a FULL sync cycle.** Extracted `full_sync_cycle` (local tick + queue drain, THEN the remote materialization sweep) and routed both the 5-min scheduler loop and `trigger_sync_tick` through it. Previously "Sync Now" only ran the local half, so it never pulled down / materialized pending remote content (only the longpoll did).
- **Tick-before-discovery ordering preserved** (load-bearing, DBSYNC-66) and the **outer** `index_materialized_folders_as_cloudsc_placeholders_internal` is called so the per-folder Windows CfAPI guard still fires. The `sync_running` CAS gate scope is unchanged — each caller keeps owning it; the helper never touches it.
- **Honest feedback.** Widened `TriggerSyncResponse` with a `reason` enum (`started` / `already_running`). The frontend uses a dedicated `TriggerSyncResponse` TS type, branches its activity-log message on `reason`, and flips the "Syncing" indicator **optimistically** on click (instant button disable + immediate dashboard refresh to close the 400 ms poll gap), cleared defensively once the real status catches up.
- Contract hygiene: reverted the shared `TriggerActionResponse` TS type back to `{ accepted }` (its correct shape for the other 3 `trigger_*` commands) and added a separate `TriggerSyncResponse` for `trigger_sync_tick`.

## Constraints honored
- NO `sleep`/artificial delay inside the `sync_running` gate (would starve the longpoll).
- NO toast/notification infra added (none exists) — uses the existing `pushLog` activity log.
- Folder-delete latency was **split out** to a follow-up ticket (DBSYNC-70) — this PR is Sync Now only.

## Stack
desktop-tauri (Rust backend + React/TS frontend)

## Jira Ticket
DBSYNC-68

## Test Plan
- [x] `cargo clippy --all-targets` clean (no warnings)
- [x] `cargo test` — 184 passed, 0 failed
- [x] `npm --workspace apps/desktop run build` (tsc + vite) clean
- [ ] Manual live QA (Windows, requires running the app):
  - [ ] With pending remote content, click **Sync Now** → the remote folder/files materialize locally (proves the sweep now runs on the button, not just the longpoll)
  - [ ] Double-click Sync Now → second click logs "Sync already running." (not silent)
  - [ ] Button disables **instantly** on click (optimistic flip, no ~400 ms poll lag)
  - [ ] Delete a local folder then let a full cycle run → folder does NOT resurrect as a placeholder (DBSYNC-66 ordering regression check)

### Stack-specific validation
- desktop-tauri: affected IPC command (`trigger_sync_tick`) exercised; no terminal errors; frontend uses `@tauri-apps/api` `invoke` only.

🤖 Generated with Claude Code
https://claude.ai/code/session_01MQMywd4gJCi52EsngmpU9h

### DIFF
--- a/apps/desktop/src-tauri/src/commands.rs
+++ b/apps/desktop/src-tauri/src/commands.rs
@@ -262,17 +262,9 @@ pub fn start_background_scheduler(
             // Show "Syncing" for the duration of this tick (DBSYNC-48); the
             // tooltip is refreshed back to Idle/Error after the store below.
             update_tray_tooltip(&app, &current_tray_status_label(&app_state));
-            // Run the sync tick first so a local folder deletion is detected,
-            // its `delete` job enqueued, and drained within this same tick
-            // (deleting the folder remotely) before discovery runs. Otherwise
-            // discovery would still see the (now-orphaned) remote folder and
-            // re-create its `.cloudsc` placeholder.
-            let _ = run_sync_tick_internal(&app_state);
-            match index_materialized_folders_as_cloudsc_placeholders_internal(&app_state) {
-                Ok(n) if n > 0 => tracing::info!(count = n, "indexed new remote placeholder(s)"),
-                Ok(_) => {}
-                Err(e) => tracing::error!(error = %e, "remote placeholder indexing failed"),
-            }
+            // Ordering (tick before discovery) is load-bearing — see the doc
+            // comment on `full_sync_cycle` (DBSYNC-66).
+            crate::sync_pipeline::full_sync_cycle(&app_state);
             app_state.sync_running.store(false, Ordering::Release);
             update_tray_tooltip(&app, &current_tray_status_label(&app_state));
         }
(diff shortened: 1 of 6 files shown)


mbsanchez:
## CTO / Code Review (Claude - code-reviewer)

**Verdict on original commit (6455131): REQUEST CHANGES** — 1 critical bug found; **fixed in 1987588**, now approvable.

### 🔴 Critical (fixed)
**Optimistic `optimisticSyncing` flag could stick `true` forever → Sync Now button + "Syncing" footer permanently disabled.** The flag was only cleared by an effect firing when `status.syncRunning` was observed `true`. For a fast/no-op Sync Now, the full cycle can finish before any 400 ms poll ever reads `syncRunning=true` (the poll is itself gated on `syncRunning`), so the flag never cleared. Fix (1987588): clear the flag in a `finally` right after `refreshDashboardNow()` resolves — it now only bridges the click→first-fetch gap; a genuinely running sync keeps the UI busy via the real `status.syncRunning` afterward. Removed the fragile observe-true effect. No `setTimeout`/sleep introduced (honors the no-artificial-delay constraint).

### 🟡 Accepted as-is
- **Behavior parity:** `trigger_sync_tick` no longer captures `run_sync_tick_internal`'s `Err` into `set_last_error` (now swallowed like the 5-min loop). Minor: a failed *manual* sync no longer flips the UI to "Error". Accepted for KISS/parity — that `Err` is only rare infra/DB failure; per-job errors are handled inside the drain.
- Reworded the misleading `refreshDashboardNow` comment (it's a one-shot fetch, not a poll-loop kick).

### ✅ AC scorecard
1. Shared `full_sync_cycle` = tick THEN outer discovery sweep, ordering preserved, cfapi guard intact — **MET**
2. 5-min loop delegates; CAS + tooltip unchanged — **MET**
3. `trigger_sync_tick` delegates; gate scope unchanged; no sleep — **MET**
4. `TriggerSyncResponse.reason` + TS mirror match (`snake_case` ↔ `"started"|"already_running"`); `TriggerActionResponse` untouched for the other 3 commands — **MET**
5. Distinct started/already-running signal + optimistic flip cleared safely — **MET** (after 1987588)

Rust idioms clean (no new unwrap/panic, correct derives, no dead code); frontend hooks rules OK, no stray logs. **With 1987588 applied: APPROVE.** Live Windows/CfAPI QA still pending.

Issue #83, opened 2026-08-19
DBSYNC-72 Slice 3: First live macOS validation (Mac required)


mbsanchez:
Vertical slice 3 of DBSYNC-72 (https://manuelbsan.atlassian.net/browse/DBSYNC-72).

**Type:** Mac required (validation only, no code) · **Blocked by:** Slice 2 + the Apple Developer prerequisites listed on the Jira ticket · **Scope:** none

First time the macOS path is exercised on real hardware. Cross-compilation is not possible: the appex is an Xcode/Swift target, signing needs `codesign`, notarization needs `xcrun notarytool` — all macOS-only.

## Acceptance criteria
- [ ] `npm run dev:mac` produces a launchable `.app`; app starts hidden as a tray/menu-bar item (`LSUIElement` true, `src-tauri/Info.plist`).
- [ ] Dropbox OAuth completes end-to-end (local listener on port 53682) and the token persists in the macOS keychain across restarts. Hardened runtime must not break either.
- [ ] Finder Sync extension appears and can be enabled under System Settings -> Privacy & Security -> Extensions -> Finder; the three badges (`synced`, `out_of_sync`, `syncing`) render on files in the sync folder.
- [ ] `.cloudsc` placeholder behavior verified on macOS: creation, file association, hydration-on-open. Confirm the DBSYNC-66 "no `.cloudsc` on Windows" change did not regress the macOS path.
- [ ] A basic bidirectional sync passes (local->remote upload, remote->local materialization) with the sync folder on macOS.
- [ ] Any macOS-specific defect found is filed as its own DBSYNC ticket, NOT fixed here.

mbsanchez:
## Plan: First live macOS validation (GitHub #83)

**Scope dir(s)**: None — this slice writes no repo files. It is a manual validation pass on real macOS hardware.
**Implementing agent(s)**: Human operator with a Mac (no agent can execute this — Xcode/Swift compilation, `codesign`, and Finder extension enablement are macOS-only and require a GUI/keychain session no CLI agent has).
**Estimated complexity**: medium (not from engineering difficulty — there is none — but from the number of distinct manual checks and the need for careful evidence capture).
**Dependencies**: Blocked by Slice 2 (#82) landing, plus the Apple Developer prerequisites named on the Jira ticket (an active Apple Developer Program membership / team, and Xcode with a signed-in Apple ID for at least ad-hoc/development signing).

[...]

Issue #81, opened 2026-08-19
DBSYNC-72 Slice 1: Fix the overlay_state path mismatch (AFK)


mbsanchez:
Vertical slice 1 of DBSYNC-72 (https://manuelbsan.atlassian.net/browse/DBSYNC-72).

**Type:** AFK (no Mac needed) · **Blocked by:** none · **Scope:** `apps/desktop/native/macos/FinderSyncExtension`, `apps/desktop/src-tauri/src`, `apps/desktop/README.md`

Finding F1: the Finder Sync extension reads `~/Library/Applications/DropboxSyncDesktop/overlay_state.json`, but the Rust side writes `~/Library/Application Support/DropboxSyncDesktop/overlay_state.json`. `Library/Applications` is not a real macOS directory, so the extension always reads nothing and no badge can ever render. The wrong path is in three files, not two.

## Acceptance criteria
- [ ] `SyncExtension.swift` `overlayStateURL()` resolves the SAME path the Rust side writes, matching `db::app_data_dir()` (`storage/db.rs:1220`, `:1241`, `:1305`) + `overlay_state_path()` (`overlay_state.rs:129`).
- [ ] Prefer `FileManager.default.url(for: .applicationSupportDirectory, in: .userDomainMask)` over hand-joining `Library/Application Support` onto the home URL, so the container/non-container distinction is explicit rather than accidental.
- [ ] `native/macos/FinderSyncExtension/README.md:5` corrected — it documents the wrong path too.
- [ ] `apps/desktop/README.md:135` ("Shared overlay state contract") documents `~/Library/Application Support/DropboxSyncDesktop/overlay_state.json`.
- [ ] `grep -rn "Library/Applications" apps/` returns no matches. Regression guard: Swift is neither built nor linted by any CI job, so this grep is the only mechanism keeping the three files in agreement with the Rust source of truth.
- [ ] No behavior change on Windows; `cargo test`, `cargo clippy`, `npm --workspace apps/desktop run build` all pass.

NOTE: this slice alone does not prove badges work — that is verified in Slice 3.

mbsanchez:
## Plan: Fix the overlay_state path mismatch (GitHub #81)

**Scope dir(s)**: `apps/desktop/native/macos/FinderSyncExtension/`, `apps/desktop/README.md`
**Implementing agent(s)**: `cowork-workflow:tauri-expert` (Swift is the only real change; no Rust edit is needed)
**Estimated complexity**: low
[...]

PR #86, opened 2026-08-19
fix(#DBSYNC-72): Finder Sync reads overlay_state from Application Support

mbsanchez:
## Summary

Fixes the overlay-state path mismatch that made Finder Sync badges impossible on macOS (DBSYNC-72 Slice 1, closes the code half of #81), plus a rider commit that restores the macOS Rust build.

**Commit 1 — the slice.** `SyncExtension.swift` read `~/Library/Applications/DropboxSyncDesktop/overlay_state.json`; the Rust side writes `~/Library/Application Support/DropboxSyncDesktop/`. `Library/Applications` is not a real macOS directory, so `reloadState()` read nothing on every tick and no badge could ever render — silently, because a failed read only sets `state = nil`. Now resolved through `FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask)` with a **non-crashing fallback**: a Finder Sync extension must never trap, it would take the whole Finder process down. Two READMEs documented the same wrong path and are corrected.

**Commit 2 — rider, unrelated to the path bug.** `run_events.rs` used `PathBuf` inside a `#[cfg(macos/ios)]` arm while importing only `Path`, so the Rust backend **had not compiled on macOS since 9ea4fbc (2026-07-13)**. One line, `cfg`-gated to match its only use. It ships here because it blocks `cargo test` and the `npm run dev:mac` that Slice 3 (#83) depends on.

## Stack

`desktop-tauri` — Swift (Finder Sync appex) + Rust backend. No frontend change.

## Jira Ticket

DBSYNC-72 — https://manuelbsan.atlassian.net/browse/DBSYNC-72

## Test Plan

- [x] `cargo test --manifest-path apps/desktop/src-tauri/Cargo.toml` — **163 passed**, 0 failed
- [x] `cargo clippy --manifest-path apps/desktop/src-tauri/Cargo.toml` — no new warnings (the remaining ones are pre-existing: `unused import: std::io`, two dead functions, one `sort_by_key`)
- [x] `npm --workspace apps/desktop run build` — `tsc` + `vite` clean
- [x] `npm --workspace apps/desktop run build:finder-sync` — **BUILD SUCCEEDED**, no errors, no warnings, x86_64 + arm64
- [x] Regression guard: `grep -rn "Library/Applications" apps/` returns **0 matches** (was exactly 3)

### desktop-tauri validation

Path equivalence was proven at runtime rather than by inspection. Running the exact resolution logic under `xcrun swift`:

```
/Users/<user>/Library/Application Support/DropboxSyncDesktop/overlay_state.json
```

which is byte-for-byte the location `db::app_data_dir()` documents (`storage/db.rs:1220`) and writes (`:1241`). This also confirms `.userDomainMask` returns the plain per-user directory rather than a container path — consistent with the no-sandbox ADR on the ticket, since no `.entitlements` file exists anywhere under `apps/desktop`.

## What this PR does NOT prove

- **That badges actually render.** Compiling is not rendering. That is Slice 3 (#83): enabling the extension under System Settings → Privacy & Security → Extensions → Finder on real hardware and looking at the three badge tiers.
- **Nothing was verified by CI.** The `ci` workflow is `disabled_manually` and last ran on 2026-07-12 (last 8 runs red). Every check above was run locally. See DBSYNC-74.

## Reviewer notes

- The AC on #81 cited `FileManager.url(for:in:)`; that is not a valid non-throwing overload. The plural `urls(for:in:)` is used instead, and the ticket AC has been corrected.
- Decoding `overlay_state.json` still uses `keyDecodingStrategy = .convertFromSnakeCase`, which on Foundation prior to the swift-foundation rewrite also rewrote **dictionary** keys — i.e. the `paths` map keyed by relative file path. It does not reproduce on macOS 26 (verified), but the app targets macOS 12+ and `JSONDecoder` comes from the user's OS. Deliberately left alone here to keep this diff minimal; tracked as **DBSYNC-73**.
- The local Rust toolchain had to be updated (1.74 → 1.97.1) before anything compiled: `Cargo.lock` is `version = 4`, which needs Cargo ≥ 1.78.

Refs: DBSYNC-72, #81 · Related: DBSYNC-73, DBSYNC-74

### DIFF
--- a/apps/desktop/README.md
+++ b/apps/desktop/README.md
@@ -132,7 +132,7 @@ APPLE_TEAM_ID=XXXXXXXXXX npm run build:finder-sync
 
 The Rust backend writes:
 
-- macOS: `~/Library/Applications/DropboxSyncDesktop/overlay_state.json`
+- macOS: `~/Library/Application Support/DropboxSyncDesktop/overlay_state.json`
 - Windows: `%LOCALAPPDATA%\\DropboxSyncDesktop\\overlay_state.json`
 
 Schema summary:
--- a/apps/desktop/native/macos/FinderSyncExtension/README.md
+++ b/apps/desktop/native/macos/FinderSyncExtension/README.md
@@ -2,7 +2,7 @@
 
 The Tauri app writes **`overlay_state.json`** next to the SQLite database:
 
-- Path: `~/Library/Applications/DropboxSyncDesktop/overlay_state.json`
+- Path: `~/Library/Application Support/DropboxSyncDesktop/overlay_state.json`
 - Schema: see `src-tauri/src/overlay_state.rs` (`version`, `updated_at`, `sync_folder`, `paths` map with tiers `synced`, `out_of_sync`, `syncing`).
 
 This folder contains a **Finder Sync** implementation (`SyncExtension.swift`) that registers three badge images and assigns them per file using `requestBadgeIdentifier(for:)`.
--- a/apps/desktop/native/macos/FinderSyncExtension/SyncExtension.swift
+++ b/apps/desktop/native/macos/FinderSyncExtension/SyncExtension.swift
@@ -28,10 +28,18 @@ final class DropboxSyncFinderSync: FIFinderSync {
         }
     }
 
+    /// Resolves the same path the Rust side writes in `db::app_data_dir()`:
+    /// `~/Library/Application Support/DropboxSyncDesktop/overlay_state.json`.
+    /// The extension is not sandboxed, so `.userDomainMask` yields the plain
+    /// per-user Application Support directory rather than a container path.
     private func overlayStateURL() -> URL {
-        let home = FileManager.default.homeDirectoryForCurrentUser
-        return home
-            .appendingPathComponent("Library/Applications/DropboxSyncDesktop/overlay_state.json")
+        let appSupport = FileManager.default
+            .urls(for: .applicationSupportDirectory, in: .userDomainMask)
+            .first
+            ?? FileManager.default.homeDirectoryForCurrentUser
+                .appendingPathComponent("Library/Application Support")
+        return appSupport
+            .appendingPathComponent("DropboxSyncDesktop/overlay_state.json")
     }
 
     private func monitoredDirectoryURLs() -> Set<URL> {
--- a/apps/desktop/src-tauri/src/run_events.rs
+++ b/apps/desktop/src-tauri/src/run_events.rs
@@ -1,4 +1,6 @@
 use std::path::Path;
+#[cfg(any(target_os = "macos", target_os = "ios"))]
+use std::path::PathBuf;
 use std::time::Duration;
 
 use tauri::AppHandle;


mbsanchez (commented):
## CTO review — DBSYNC-72 Slice 1 (#81) · verdict: **APPROVED with notes**

Stack profile: `desktop-tauri` (CLAUDE.md declaration, confirmed by `apps/desktop/src-tauri/`). Swift appex + 1 Rust line, no frontend. QA strategy `manual` (Slice 3, #83).

Reviewed against the ticket AC and the implementation plan on #81, and **every claim in the PR body was re-run locally** rather than trusted (results at the bottom — all five hold). The core change is correct: `SyncExtension.swift:35-43` now resolves the exact path `db::app_data_dir()` writes (`storage/db.rs:1233-1240` → `overlay_state.rs:100-106`), the Rust rider is complete, and the appex builds clean from an empty DerivedData with zero warnings for `x86_64 + arm64`.

No blocking finding. Everything below is either out of the slice's declared minimal-diff scope (the plan explicitly said "do not touch `monitoredDirectoryURLs()`, `reloadState()`, or the decode struct") or a follow-up. But two of them will bite during Slice 3 QA, so read W1 and W2 before validating on hardware.

---

### Critical (must fix)

None.

---

### Warnings (should fix — none block this merge)

**W1 · `cargo test` writes the real `overlay_state.json` this extension now reads — QA hazard for Slice 3.**
`apps/desktop/src-tauri/src/sync_pipeline.rs:63` → `apps/desktop/src-tauri/src/overlay_state.rs:100` calls `db::app_data_dir()` unconditionally, which resolves the **real** per-user directory even under `#[cfg(test)]` with a `tempdir` sync root. Empirically, the `cargo test` run I did for this review created:

```
~/Library/Application Support/DropboxSyncDesktop/overlay_state.json
  "sync_folder": "/var/folders/3s/x516y5cs49q66f_7pgw8d2j00000gn/T/.tmpz8HNWd/synced"
```

on a machine that has never run the built app. This is pre-existing and **not caused by this PR** — but this PR is exactly what makes it consequential: until now the extension read a non-existent path, so the pollution was inert. From now on, a developer who runs `cargo test` and then enables the extension gets `FIFinderSyncController.directoryURLs` pointed at a deleted `/var/folders/...` temp dir (`SyncExtension.swift:45-50`), i.e. **no badges, for a reason that has nothing to do with the code under test**. Recommend: (a) a fresh ticket to make the overlay writer honour a test-injectable data dir, and (b) Slice 3 QA must `rm ~/Library/Application Support/DropboxSyncDesktop/overlay_state.json` and start the real app before judging badges.

**W2 · The "same path" contract is asymmetric: Rust honours `$HOME`, Swift does not.**
`storage/db.rs:1259-1266` resolves via `std::env::var("HOME")`; `FileManager.urls(for:in:)` resolves from the passwd record and ignores the environment. Demonstrated:

```
$ HOME=/tmp/fakehome xcrun swift path.swift
resolved: /Users/mbsanchez/Library/Application Support/DropboxSyncDesktop/overlay_state.json
ENV HOME: /tmp/fakehome
```

So the equivalence proof in the PR body holds **only while `$HOME` equals the passwd home**. That is true for a Finder-launched app and for the Finder-hosted appex (both inherit the launchd session environment), so this is not a merge blocker — but it is false for anything launched from a shell with a rewritten `HOME`, `su` without `-`, or some MDM/network-home setups, and in those cases the app writes one file and the extension reads another, silently. Cheapest hardening: state the invariant in the `overlayStateURL()` doc comment (it currently only covers the sandbox axis), or have the Rust side fall back to the passwd home when `$HOME` is absent/divergent.

**W3 · The regression guard named in the AC is vacuous going forward.**
`grep -rn "Library/Applications" apps/` returns 0 (verified). But it greps for a literal that no longer exists anywhere and can never be reintroduced by an innocent edit — it cannot catch the *next* instance of this bug class (e.g. someone renames the dir to `DropboxSync` on the Rust side, or the Swift string picks up a typo). Since Swift is neither built nor linted by CI, a **positive** assertion is what's needed: a shell check that `SyncExtension.swift`, both READMEs and `storage/db.rs` all contain `Library/Application Support` + `DropboxSyncDesktop`, wired into `npm run build:finder-sync` or a `cargo test`. Given DBSYNC-74 (CI disabled), this matters more than usual.

**W4 · The failure mode that hid this bug for months is still silent, and the no-sandbox invariant is asserted but not enforced.**
`SyncExtension.swift:69-78`: both the read failure and the decode failure collapse to `state = nil` with no `NSLog`. That is precisely why `Library/Applications` survived — and `registerBadgeImages()` two functions above (`:61`) already shows the house style of logging a missing resource. One `NSLog` on the first transition to `nil`, including `url.path`, would make Slice 3 debuggable from Console.app instead of by inspection.
Related: the comment at `:33-34` ("The extension is not sandboxed, so `.userDomainMask` yields the plain per-user directory") is true today — I confirmed the built product carries only `com.apple.security.get-task-allow`, no `com.apple.security.app-sandbox`, and there is no `ENABLE_APP_SANDBOX` in `DropboxSyncFinderSync.xcodeproj/project.pbxproj` — but nothing enforces it. If the appex is ever sandboxed (Mac App Store, or an entitlements file added for some other reason), **both** the primary branch and the fallback branch resolve to `~/Library/Containers/<id>/Data/Library/Application Support`, they fail together, and with W4's silence the symptom is once again "badges just don't appear". The fallback is correct (see V5 — it resolves byte-identically today), it simply cannot detect that class of breakage.

---

### Suggestions (consider improving)

**S1 · Mirror the Windows sibling's mtime cache.** `reloadState()` now does a synchronous `Data(contentsOf:)` **plus a full `JSONDecoder` pass over the whole `paths` map every 2 seconds on the appex main thread** (`:25-28`, `:69-78`) — the same thread Finder calls `requestBadgeIdentifier(for:)` on. Before this PR that work was a fast `ENOENT`; the fix makes it hot, and `overlay_state.rs:102` writes `to_string_pretty` with one entry per indexed file, so a 100k-file sync root is a multi-MB pretty-printed decode every tick. The Windows shell-menu already solved exactly this: `apps/desktop/native/windows/shell-menu/src/scope.rs:167-180` re-parses "only when its mtime changed". Same three lines in Swift (`resourceValues(forKeys: [.contentModificationDateKey])`) would make the tick free in the common case.

**S2 · `SyncExtension.swift:64` — `NSImage(contentsOfFile: path)!` contradicts the rationale used to justify the fallback.** The commit message argues (correctly, in spirit) that "a Finder Sync extension must never trap"; five lines away from the code it guards, a corrupt or unreadable PNG that still resolves via `Bundle.main.path(forResource:)` force-unwraps to a crash. Pre-existing, and out of this slice's scope, but it should not survive the ticket that established the principle. Minor factual note on the same claim: an appex runs **out of process**, so a trap kills the extension (badges vanish, extension may be disabled by the system) rather than "taking the whole Finder process with it". The conclusion is right; the stated reason is not — worth fixing in the doc comment so nobody builds on it.

**S3 · `version` is decoded and never validated** (`:107`). Verified that a payload with `"version": 99` decodes without complaint, so a future contract bump (`overlay_state.rs:94` documents "bump on a breaking change") would be consumed as if it were v1. A `guard state.version == 1` with an `NSLog` is the whole fix.

**S4 · `FIFinderSyncController.default().directoryURLs` is reassigned on every 2 s tick** (`:27`) even when unchanged. Harmless before (it was always `[]`), now it re-registers a real URL 30 times a minute. Assign only on change.

**S5 · The fallback branch is unreachable and therefore untested.** `urls(for:.applicationSupportDirectory, in:.userDomainMask)` returned exactly one element here; `.first` is never `nil` on a functioning system. I verified the fallback independently and it produces the byte-identical path, so it is *correct* — just note it is dead code whose only real purpose is to avoid a force-unwrap, which is a good reason to keep it.

**S6 · On deferring DBSYNC-73 (`keyDecodingStrategy = .convertFromSnakeCase`, `:76`) — I agree with the deferral, but not with its priority.** Confirmed it does not reproduce here (a payload with key `my_folder/my_file.txt` decodes unmodified on macOS 26). But `MACOSX_DEPLOYMENT_TARGET = 12.0` (`project.pbxproj:167,196`) and `JSONDecoder` ships with the user's OS, so on macOS 12-14 every relative path containing an underscore silently misses the `paths` lookup at `:99` — the *same* silent-nothing-renders failure class this PR exists to fix, on the majority of supported OS versions. The fix is ~3 lines (explicit `CodingKeys` on `OverlayState`, drop the strategy). Recommend DBSYNC-73 be marked as **blocking Slice 3 (#83)** rather than a generic follow-up, so hardware validation isn't signed off on a macOS 26 box while macOS 12-14 users get nothing.

---

### The Rust rider (commit `7794d16`) — correct and complete

- The `cfg` on the import (`run_events.rs:2`) is byte-identical to the one on the arm that uses it (`:18`), and `PathBuf`'s only occurrence in the file is `:20`, inside that arm. No unused-import warning on Windows/Linux, no missing import on macOS.
- `Path` (`:1`) stays unconditional and is genuinely used outside the gated arm at `:68` and `:86`. Correct to leave it ungated.
- Nothing else in that arm is missing: `handle_cloudsc_paths_from_os` is imported at `:14`, `RunEvent::Opened` and `url.to_file_path()` need no extra import.
- The claimed breakage is real, not theoretical: `git show origin/main:apps/desktop/src-tauri/src/run_events.rs` has `use std::path::Path;` at line 1 and `Vec<PathBuf>` at line 18. macOS `cargo test` cannot have compiled on `main`.

---

## Verification — every PR claim re-run locally

Environment: macOS 26.5.1 (25F80), Xcode 15.4 (15F31d, MacOSX14.5 SDK), rustc/cargo 1.97.1.

**V1 — `cargo test --manifest-path apps/desktop/src-tauri/Cargo.toml` → claim "163 passed": TRUE.**
```
test result: ok. 163 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.06s
(+ bin: 0 tests, doc-tests: 0 tests)
```

**V2 — `cargo clippy` → claim "no new warnings, remaining ones pre-existing": TRUE, and the list matches exactly.**
```
warning: unused import: `std::io`
warning: function `dehydrate_file_atomic` is never used
warning: function `cloudsc_paths_from_current_exe_args` is never used
warning: consider using `sort_by_key`
`dropbox_sync_desktop` (lib) generated 4 warnings
(+ 1 future-incompat note for the transitive crate `block v0.1.6`, unrelated)
```

**V3 — `build-appex.sh` → claim "BUILD SUCCEEDED, no errors, no warnings, x86_64 + arm64": TRUE.**
Incremental run: `** BUILD SUCCEEDED **`. Re-run **from an empty DerivedData** to be sure nothing was cached: exit 0, `grep -c "warning:"` → **0**, `grep "error:"` → none.
```
$ lipo -info build/DropboxSyncFinderSync.appex/Contents/MacOS/DropboxSyncFinderSync
Architectures in the fat file: ... are: x86_64 arm64
```

**V4 — `grep -rn "Library/Applications" apps/` → claim "0 matches": TRUE.**
Exit code 1, no output. Also ran it repo-wide (excluding `.git`, `node_modules`, `target`, `build`): 0 matches. Caveat in W3 about what this guard can and cannot catch.

**V5 — path equivalence → claim "resolves exactly what `db::app_data_dir()` writes, non-container": TRUE.**
Ran the *exact* expression from `SyncExtension.swift:36-42` under `xcrun swift`:
```
resolved: /Users/<user>/Library/Application Support/DropboxSyncDesktop/overlay_state.json
count:    1
all:      ["/Users/<user>/Library/Application Support"]
```
Matches `data_dir_for(DataDirOs::Macos, …)` at `storage/db.rs:1233-1240` + `overlay_state_path()` at `overlay_state.rs:128-130` character for character. Additionally verified beyond the PR's claim:
- the **fallback** branch in isolation resolves to the identical path (so the `??` cannot silently produce a different one today);
- the **built product** carries only `com.apple.security.get-task-allow` (`codesign -d --entitlements -`) and the pbxproj has no `ENABLE_APP_SANDBOX`, which is what actually substantiates the "not sandboxed ⇒ not a container path" premise;
- the counter-case in W2 (`$HOME` divergence), where the premise does *not* hold.

**V6 (bonus) — `npm --workspace apps/desktop run build`: TRUE.** `tsc` + `vite build` clean, 55 modules, no errors.

---

Not approving via the button — the maintainer stays the approver of record. From my side: **ship it**, with W1 folded into the Slice 3 QA script and W3/W4 filed as follow-ups.

*(Auth by Claude - architect-reviewer)*

mbsanchez:
### Correction to the commit message of 754d538

The commit body justifies the non-crashing fallback with:

> a Finder Sync extension must never trap, it would take the whole Finder process with it

**That reason is wrong.** Finder Sync extensions are app extensions and run **out of process**; a trap kills the extension host, not Finder itself. Finder keeps running and simply stops showing badges until the extension is relaunched.

The conclusion is unchanged and still right — a silent trap that disables badges with no user-visible cause is exactly the failure class this PR exists to remove, and `.first` on an empty array would be precisely that. But the stated mechanism is inaccurate and should not be repeated in the Slice 2 ADR or anywhere else.

Not amending the commit: the history is already pushed and the inaccuracy is in prose, not in code. This comment is the correction of record.

Related, from the review (S2): `SyncExtension.swift:64` already force-unwraps `NSImage(contentsOfFile:)!`, which contradicts the same principle. Pre-existing, untouched by this PR, and worth folding into the DBSYNC-73 change since it lands in this same file.

Issue #84, opened 2026-08-19
DBSYNC-72 Slice 4: Manual signing + notarization, proven by hand first (HITL)


mbsanchez:
Vertical slice 4 of DBSYNC-72 (https://manuelbsan.atlassian.net/browse/DBSYNC-72).

**Type:** HITL, Mac required · **Blocked by:** Slice 3 · **Scope:** `apps/desktop/src-tauri`

Manual-first by design: the notarization path is proven by hand before any CI automation is written, so a later CI failure is never ambiguous between "the pipeline is wrong" and "the signing configuration is wrong". Also resolves F4 (nested `.appex` signing unverified).

## Acceptance criteria
- [ ] `APPLE_SIGNING_IDENTITY` and `APPLE_TEAM_ID` exported from the local shell; no Apple identity, team id, Apple ID or app-specific password is committed.
- [ ] `npm --workspace apps/desktop run build` produces a `.app` whose `Contents/PlugIns/DropboxSyncFinderSync.appex` is signed with the same team as the host. If Tauri does not re-sign the nested appex, add an explicit signing step before bundling.
- [ ] `codesign -dv --verbose=4` on host and appex both show a Developer ID authority and an identical TeamIdentifier; `codesign -dv --deep --strict` passes on the `.app`.
- [ ] `xcrun notarytool submit ... --wait` returns `status: Accepted` (log pasted on the PR); on rejection the log is read and the cause fixed.
- [ ] `xcrun stapler staple` succeeds and `xcrun stapler validate` passes.
- [ ] `spctl -a -vvv -t exec <app>` reports `accepted` / `source=Notarized Developer ID` — no Gatekeeper prompt requiring right-click-open.
- [ ] On a Mac where the app was never built: the extension can be enabled under System Settings -> Privacy & Security -> Extensions -> Finder and the three badge tiers render (screenshot on the PR).
- [ ] Target architecture decided and documented: arm64-only vs `--target universal-apple-darwin`. GitHub's `macos-latest` runner is arm64, so an Intel-compatible build needs the universal target explicitly.
- [ ] **GATE: Slice 5 does not start until every AC above has passed.**

mbsanchez:
## Plan: Manual signing + notarization, proven by hand first (GitHub #84)

**Scope dir(s)**: `apps/desktop/src-tauri` (read/reference only — this slice is about running commands against artifacts it produces, not about editing files, unless Tauri does not re-sign the nested appex, see step 4).
[...]

Issue #82, opened 2026-08-19
DBSYNC-72 Slice 2: Lock in the non-sandboxed, hardened-runtime extension configuration (AFK)


mbsanchez:
Vertical slice 2 of DBSYNC-72 (https://manuelbsan.atlassian.net/browse/DBSYNC-72).

**Type:** AFK (no Mac needed) · **Blocked by:** Slice 1 · **Scope:** `apps/desktop/native/macos/FinderSyncExtension`, `apps/desktop/src-tauri`

Resolves F2 and F3. The sandbox question is decided on paper: the extension stays NOT sandboxed, with Hardened Runtime. App Sandbox is mandatory only for Mac App Store distribution; Developer ID requires Hardened Runtime + notarization instead. The repo is already in that configuration (`ENABLE_HARDENED_RUNTIME = YES` at `project.pbxproj:158` and `:187`, no `CODE_SIGN_ENTITLEMENTS`, no `.entitlements` anywhere), so the decision costs zero migration work.

## Acceptance criteria
- [ ] ADR recorded in `native/macos/FinderSyncExtension/README.md`: no App Sandbox, Hardened Runtime ON, no App Group container, state file stays under Application Support.
- [ ] `project.pbxproj` asserted: `ENABLE_HARDENED_RUNTIME = YES` in Debug and Release, no `CODE_SIGN_ENTITLEMENTS` key.
- [ ] No `*.entitlements` file is added anywhere under `apps/desktop`.
- [ ] `db::app_data_dir()` is unchanged — no `~/Library/Group Containers/` path is introduced.
- [ ] Extension bundle id stays `com.mobsanchez.dropboxsyncdesktop.findersync` under host `com.mobsanchez.dropboxsyncdesktop`, same Apple team.
- [ ] `"signingIdentity": "-"` removed from the base `tauri.conf.json` so `APPLE_SIGNING_IDENTITY` governs. Ad-hoc signing must no longer be the effective default for a release build, and `ci.yml` must still pass unsigned on PRs (it runs on `macos-latest` with `DROPBOX_APP_KEY: dummy` and no certificate — a committed identity would break every PR).
- [ ] `DEVELOPMENT_TEAM` supplied env-driven via `APPLE_TEAM_ID` (per `build-appex.sh:31-33`) rather than left empty at `project.pbxproj:157`; no team id is committed.
[...]

PR #87, opened 2026-08-19
fix(#DBSYNC-72): sign the Finder Sync appex with a real Developer ID

mbsanchez:
## Summary

DBSYNC-72 Slice 2 (#82): make the Finder Sync appex sign with a real Developer ID instead of ad-hoc, driven entirely by the environment, and record the no-sandbox decision where implementers will read it.

Two of the slice's acceptance criteria turned out to be wrong. Both are documented below with the evidence, because the corrections matter more than the diff.

## The AC that would have made things worse

> re-running `build-appex.sh` with `APPLE_TEAM_ID=XCAA3WMJM6` exported must stop producing `codesign --sign -`

It does not. It fails the build:

```
$ APPLE_TEAM_ID=XCAA3WMJM6 bash build-appex.sh
Using DEVELOPMENT_TEAM from APPLE_TEAM_ID.
error: No signing certificate "Mac Development" found: No "Mac Development" signing
certificate matching team ID "XCAA3WMJM6" with a private key was found.
** BUILD FAILED **
```

`CODE_SIGN_STYLE = Automatic` resolves a **development** certificate for a build action. A Developer ID-only account — this project's distribution model, per the ADR — never has a *Mac Development* certificate. So the fix belongs in `build-appex.sh`, not in `project.pbxproj`: manual signing with an explicit `CODE_SIGN_IDENTITY` whenever a team is present.

Verified on the produced artifact, not the build log:

| Environment | `codesign -dv` on `build/DropboxSyncFinderSync.appex` |
|---|---|
| `APPLE_TEAM_ID=XCAA3WMJM6` | `Authority=Developer ID Application: Manuel Sanchez (XCAA3WMJM6)` → Developer ID CA → Apple Root CA · `flags=0x10000(runtime)` · `TeamIdentifier=XCAA3WMJM6` |
| nothing exported | `flags=0x10002(adhoc,runtime)` · `Signature=adhoc` · `TeamIdentifier=not set` |

The second row is the important one: CI and ordinary local builds are untouched.

## The AC I deliberately did not implement

> `DEVELOPMENT_TEAM` supplied env-driven via `APPLE_TEAM_ID` in BOTH configurations

`project.pbxproj` keeps `DEVELOPMENT_TEAM = ""`. The script already passes the team as an xcodebuild command-line override, which takes precedence, so writing it into the project file would pin an identity in git and buy nothing. `XCAA3WMJM6` appears **0 times** in tracked sources. This machine also holds Developer ID identities belonging to other organisations; no committed value may ever make one of them selectable by accident.

## New evidence for finding F4 — this shapes Slice 4 (#84)

Xcode signs build products with **`--timestamp=none`**. The signed appex has the full Developer ID chain and hardened runtime but **no secure timestamp**, which notarization requires on every nested bundle. The appex as produced is therefore **not notarizable**; it must be re-signed with `--timestamp` before submission.

F4 asked whether Tauri re-signs the nested appex. It is now a requirement, not a curiosity: if Tauri's pass does not re-sign *with a timestamp*, Slice 4 must add an explicit step. Both `build-appex.sh` and the README say so at the point where someone would otherwise lose an afternoon to a notarization rejection.

## Stack

`desktop-tauri` — build configuration and documentation only. No Rust, Swift or TypeScript source changed.

## Jira Ticket

DBSYNC-72 — https://manuelbsan.atlassian.net/browse/DBSYNC-72

## Test Plan

- [x] `cargo test` → **163 passed**. This also exercises the config edit: `tauri_build::build()` parses `tauri.conf.json`, so a malformed JSONC would fail the build.
- [x] `cargo clippy --all-targets` → no new warnings
- [x] `npm --workspace apps/desktop run build` → clean
- [x] `build-appex.sh` → BUILD SUCCEEDED with a team exported **and** with nothing exported

### Assertion criteria ("confirm what is already true")

- `ENABLE_HARDENED_RUNTIME = YES` → 2 (Debug + Release)
- `CODE_SIGN_ENTITLEMENTS` → 0 · `*.entitlements` under `apps/desktop` → 0
- `DEVELOPMENT_TEAM = "";` → 2 · `XCAA3WMJM6` in tracked sources → 0
- `Group Containers` in `src-tauri/src` → 0 · `storage/db.rs` unmodified
- `PRODUCT_BUNDLE_IDENTIFIER = com.mobsanchez.dropboxsyncdesktop.findersync` in both configs, unchanged

## What this PR does NOT prove

Notarization acceptance and badge rendering. Those are Slices 4 (#84) and 3 (#83). Nothing here was verified by CI either — the `ci` workflow is still `disabled_manually` (DBSYNC-74).

Refs: DBSYNC-72, #82 · Related: DBSYNC-73, DBSYNC-74, DBSYNC-75

### DIFF
--- a/apps/desktop/native/macos/FinderSyncExtension/README.md
+++ b/apps/desktop/native/macos/FinderSyncExtension/README.md
@@ -34,8 +34,11 @@ entitlements, and it reads `overlay_state.json` straight from `~/Library/Applica
 No App Group container is introduced and `db::app_data_dir()` is not changed.
 
 "Carries no entitlements" is a claim about the **shipped artifact**, not merely about the sources, and the two
-can diverge: `xcodebuild` injects `com.apple.security.get-task-allow` unless told not to. Check the product,
-never the project file — `codesign -d --entitlements - build/DropboxSyncFinderSync.appex` must print nothing.
+can diverge: `xcodebuild` injects `com.apple.security.get-task-allow` unless told not to, which is why
+`build-appex.sh` disables that injection on every path, signed or ad-hoc. Check the product, never the project
+file — `codesign -d --entitlements - build/DropboxSyncFinderSync.appex` must print nothing. If it prints
+`get-task-allow`, the appex was built by something other than that script (a plain `xcodebuild`, or Xcode.app)
+and must not be shipped.
 
 **Context.** The app is distributed with **Developer ID** (direct download), not through the Mac App Store.
 App Sandbox is mandatory only for the App Store; Developer ID requires Hardened Runtime plus notarization
@@ -95,11 +98,11 @@ Three things about `build-appex.sh` that are easy to get wrong:
    bundle. The script passes `OTHER_CODE_SIGN_FLAGS=--timestamp`.
 
 Points 2 and 3 are not optional polish, because **this signature is final**: Tauri does not re-sign the appex.
-`copy_custom_files_to_bundle()` places it into `Contents/PlugIns/` without adding it to the bundler's
-`sign_paths`, and the bundler's own `codesign` call carries no `--deep` (verified in `tauri-bundler` 2.8.1,
-the version behind `@tauri-apps/cli` 2.10.1). Whatever is wrong here reaches Apple unchanged — and the bundler
-auto-submits for notarization as soon as credentials are present, so a bad signature is submitted without
-anyone asking for it.
+`copy_custom_files_to_bundle()` (`tauri-bundler` 2.8.1, `bundle/macos/app.rs`) places it into
+`Contents/PlugIns/` without adding it to the bundler's `sign_paths`, and the `codesign` invocation itself —
+which lives in `tauri-macos-sign` 2.3.3, not in the bundler — carries no `--deep`. Whatever is wrong here
+reaches Apple unchanged. Worse, once an identity resolves *and* notarization credentials are present, the
+bundler submits automatically, so a bad signature goes to Apple without anyone asking for it.
 
 With neither variable exported the appex is ad-hoc signed (`Signature=adhoc`, `TeamIdentifier=not set`), which
 is correct for development and for CI, and cannot be notarized.
--- a/apps/desktop/native/macos/FinderSyncExtension/README.md
+++ b/apps/desktop/native/macos/FinderSyncExtension/README.md
@@ -19,7 +19,9 @@ Produces `native/macos/FinderSyncExtension/build/DropboxSyncFinderSync.appex` (a
 
 ## Xcode GUI
 
-Open `DropboxSyncFinderSync.xcodeproj` to edit signing (**Signing & Capabilities** → Team) or Swift. Embed the built **`DropboxSyncFinderSync.appex`** into the main app at **`Contents/PlugIns/`**; host app and extension must use the **same development team** for production. `Info.plist.example` mirrors the committed `Info.plist` for reference.
+Open `DropboxSyncFinderSync.xcodeproj` to edit Swift. Embedding into the main app at **`Contents/PlugIns/`** is automatic via `bundle.macOS.files`, and host app and extension must use the **same Apple team**. `Info.plist.example` mirrors the committed `Info.plist` for reference.
+
+**Do not set the Team under Signing & Capabilities.** The GUI writes `DEVELOPMENT_TEAM` into the committed `project.pbxproj`, which is exactly what the signing section below keeps empty on purpose. Export `APPLE_TEAM_ID` instead.
 
 After install, the user may need to enable the extension under **System Settings → Privacy & Security → Extensions → Finder** (wording varies by macOS version).
 
@@ -27,9 +29,13 @@ After install, the user may need to enable the extension under **System Settings
 
 **Status:** Accepted — 2026-08-19 (DBSYNC-72).
 
-**Decision.** The extension is **not sandboxed**. It ships with Hardened Runtime enabled and no entitlements
-file, and it reads `overlay_state.json` straight from `~/Library/Application Support/DropboxSyncDesktop/`.
+**Decision.** The extension is **not sandboxed**. It ships with Hardened Runtime enabled and carries no
+entitlements, and it reads `overlay_state.json` straight from `~/Library/Application Support/DropboxSyncDesktop/`.
 No App Group container is introduced and `db::app_data_dir()` is not changed.
+
+"Carries no entitlements" is a claim about the **shipped artifact**, not merely about the sources, and the two
+can diverge: `xcodebuild` injects `com.apple.security.get-task-allow` unless told not to. Check the product,
+never the project file — `codesign -d --entitlements - build/DropboxSyncFinderSync.appex` must print nothing.
 
 **Context.** The app is distributed with **Developer ID** (direct download), not through the Mac App Store.
 App Sandbox is mandatory only for the App Store; Developer ID requires Hardened Runtime plus notarization
@@ -54,32 +60,48 @@ or the decoder key-conversion issue (DBSYNC-73).
 
 ## Signing: environment-driven, never committed
 
-`tauri.conf.json` deliberately does **not** set `bundle.macOS.signingIdentity`, and `project.pbxproj` keeps
-`DEVELOPMENT_TEAM = ""`. A value in either place would take precedence over the environment and silently pin
-an identity — which also breaks every unsigned build path, including CI, which holds no certificate.
+Nothing in the repository names a real identity. Two committed values look like they might, and both are
+deliberate fallbacks rather than pins:
 
-Export both variables for a signed build; the same team must be used for the app and the extension:
+- `tauri.conf.json` keeps `bundle.macOS.signingIdentity: "-"` (ad-hoc). **`APPLE_SIGNING_IDENTITY` wins over
+  it** — tauri-cli reads the environment variable first and falls back to the config key only when it is unset
+  (`interface/rust.rs`, verified in 2.10.1). The `"-"` must stay: `tauri-bundler` skips its entire signing
+  block when no identity resolves, so removing it would leave unsigned builds with no signature and no
+  `CodeResources` seal over `Contents/PlugIns/*.appex`.
+- `project.pbxproj` keeps `DEVELOPMENT_TEAM = ""`. `build-appex.sh` passes the team to `xcodebuild` as a
+  command-line override, which takes precedence, so a committed team id would buy nothing.
+
+For a signed build, export both; the same team must be used for the app and the extension:
 
 ```bash
-export APPLE_SIGNING_IDENTITY="Developer ID Application: Your Name (TEAMID)"   # Tauri, host app
-export APPLE_TEAM_ID=TEAMID                                                    # xcodebuild + notarization
-npm run build:finder-sync
+export APPLE_SIGNING_IDENTITY="Developer ID Application: Your Name (TEAMID)"   # host app, read by tauri build
+export APPLE_TEAM_ID=TEAMID                                                    # appex build + notarization
+npm run build:finder-sync    # reads APPLE_TEAM_ID only
+npm run tauri build          # reads both
 ```
 
-Two things about `build-appex.sh` that are easy to get wrong:
+Three things about `build-appex.sh` that are easy to get wrong:
 
-1. **A team alone is not enough.** The project uses `CODE_SIGN_STYLE = Automatic`, and automatic signing
-   resolves a *development* certificate for a build action. With a team set and no **Mac Development**
-   certificate in the keychain — which a Developer ID-only account never has — xcodebuild fails outright with
-   `No signing certificate "Mac Development" found`. The script therefore switches to `CODE_SIGN_STYLE=Manual`
-   and names `CODE_SIGN_IDENTITY` explicitly whenever a team is present. Override the identity with
-   `APPEX_CODE_SIGN_IDENTITY` if you need something other than `Developer ID Application`.
-2. **Xcode signs without a secure timestamp** (`--timestamp=none`). Notarization requires a secure timestamp on
-   every nested bundle, so the `.appex` as produced here is *not* notarizable on its own: it must be re-signed
-   with `--timestamp` before submission. Whether Tauri's bundling pass does that re-signing is still unverified
-   — that is finding F4 of DBSYNC-72, resolved in Slice 4 (#84).
+1. **A team alone is not enough — it breaks the build.** The project uses `CODE_SIGN_STYLE = Automatic`, and
+   automatic signing resolves a *development* certificate for a build action. With a team set and no **Mac
+   Development** certificate in the keychain — which a Developer ID-only account never has — xcodebuild fails
+   with `No signing certificate "Mac Development" found`. The script therefore switches to
+   `CODE_SIGN_STYLE=Manual` and names `CODE_SIGN_IDENTITY` explicitly whenever a team is present. Override it
+   with `APPEX_CODE_SIGN_IDENTITY` if you need something other than `Developer ID Application`.
+2. **`xcodebuild build` injects `com.apple.security.get-task-allow`** by default
+   (`CODE_SIGN_INJECT_BASE_ENTITLEMENTS`). That is a debugging entitlement: notarization refuses it outright,
+   and it defeats part of Hardened Runtime. The script sets it to `NO`.
+3. **Xcode signs with `--timestamp=none`**, while notarization requires a secure TSA timestamp on every nested
+   bundle. The script passes `OTHER_CODE_SIGN_FLAGS=--timestamp`.
 
-With neither variable exported the build is ad-hoc signed (`Signature=adhoc`, `TeamIdentifier=not set`), which
+Points 2 and 3 are not optional polish, because **this signature is final**: Tauri does not re-sign the appex.
+`copy_custom_files_to_bundle()` places it into `Contents/PlugIns/` without adding it to the bundler's
+`sign_paths`, and the bundler's own `codesign` call carries no `--deep` (verified in `tauri-bundler` 2.8.1,
+the version behind `@tauri-apps/cli` 2.10.1). Whatever is wrong here reaches Apple unchanged — and the bundler
+auto-submits for notarization as soon as credentials are present, so a bad signature is submitted without
+anyone asking for it.
+
+With neither variable exported the appex is ad-hoc signed (`Signature=adhoc`, `TeamIdentifier=not set`), which
 is correct for development and for CI, and cannot be notarized.
 
 ## Why not pure Tauri?
--- a/apps/desktop/native/macos/FinderSyncExtension/README.md
+++ b/apps/desktop/native/macos/FinderSyncExtension/README.md
@@ -23,6 +23,65 @@ Open `DropboxSyncFinderSync.xcodeproj` to edit signing (**Signing & Capabilities
 
 After install, the user may need to enable the extension under **System Settings → Privacy & Security → Extensions → Finder** (wording varies by macOS version).
 
+## ADR: no App Sandbox, Hardened Runtime on, no App Group
+
+**Status:** Accepted — 2026-08-19 (DBSYNC-72).
+
+**Decision.** The extension is **not sandboxed**. It ships with Hardened Runtime enabled and no entitlements
+file, and it reads `overlay_state.json` straight from `~/Library/Application Support/DropboxSyncDesktop/`.
+No App Group container is introduced and `db::app_data_dir()` is not changed.
+
+**Context.** The app is distributed with **Developer ID** (direct download), not through the Mac App Store.
+App Sandbox is mandatory only for the App Store; Developer ID requires Hardened Runtime plus notarization
+instead, and both are already in place — `ENABLE_HARDENED_RUNTIME = YES` in both build configurations, and no
+`CODE_SIGN_ENTITLEMENTS` key or `.entitlements` file exists anywhere under `apps/desktop`.
+
+**Why it matters.** Sandboxing is the *only* reason this extension would need an App Group. A sandboxed
+extension cannot read the host app's Application Support directory, which would force the state file into
+`~/Library/Group Containers/<group-id>/`, force a rewrite of `db::app_data_dir()`, force a migration of every
+existing user's state, and require a provisioning profile carrying the App Group entitlement. Staying
+unsandboxed removes all four at no cost.
+
+Verified rather than assumed: running the resolution logic of `overlayStateURL()` outside a container yields
+the plain per-user directory, identical to what the Rust side writes.
+
+**Trade-off accepted.** The app cannot be submitted to the Mac App Store without reopening this decision.
+
+**Contingency — not a planned step.** If the extension cannot read the state file *and* Console.app shows a
+sandbox or TCC denial for the appex, reopen this ADR and cost the App Group migration separately. Badges
+merely being absent is **not** the trigger: likelier causes are a stale `overlay_state.json` (see DBSYNC-75)
+or the decoder key-conversion issue (DBSYNC-73).
+
+## Signing: environment-driven, never committed
+
+`tauri.conf.json` deliberately does **not** set `bundle.macOS.signingIdentity`, and `project.pbxproj` keeps
+`DEVELOPMENT_TEAM = ""`. A value in either place would take precedence over the environment and silently pin
+an identity — which also breaks every unsigned build path, including CI, which holds no certificate.
+
+Export both variables for a signed build; the same team must be used for the app and the extension:
+
+```bash
+export APPLE_SIGNING_IDENTITY="Developer ID Application: Your Name (TEAMID)"   # Tauri, host app
+export APPLE_TEAM_ID=TEAMID                                                    # xcodebuild + notarization
+npm run build:finder-sync
+```
+
+Two things about `build-appex.sh` that are easy to get wrong:
+
+1. **A team alone is not enough.** The project uses `CODE_SIGN_STYLE = Automatic`, and automatic signing
+   resolves a *development* certificate for a build action. With a team set and no **Mac Development**
+   certificate in the keychain — which a Developer ID-only account never has — xcodebuild fails outright with
+   `No signing certificate "Mac Development" found`. The script therefore switches to `CODE_SIGN_STYLE=Manual`
+   and names `CODE_SIGN_IDENTITY` explicitly whenever a team is present. Override the identity with
+   `APPEX_CODE_SIGN_IDENTITY` if you need something other than `Developer ID Application`.
+2. **Xcode signs without a secure timestamp** (`--timestamp=none`). Notarization requires a secure timestamp on
+   every nested bundle, so the `.appex` as produced here is *not* notarizable on its own: it must be re-signed
+   with `--timestamp` before submission. Whether Tauri's bundling pass does that re-signing is still unverified
+   — that is finding F4 of DBSYNC-72, resolved in Slice 4 (#84).
+
+With neither variable exported the build is ad-hoc signed (`Signature=adhoc`, `TeamIdentifier=not set`), which
+is correct for development and for CI, and cannot be notarized.
+
 ## Why not pure Tauri?
 
 Finder badge overlays are implemented only via **App Extension** APIs (`FIFinderSync` / `FIFinderSyncController`). The Rust/Tauri process cannot draw those badges itself.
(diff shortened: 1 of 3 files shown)


mbsanchez (commented):
## CTO review — DBSYNC-72 Slice 2 (#82) · commit d121b28

**Verdict: CHANGES REQUESTED.** The code change is right; three of the factual claims it is
justified with are wrong, and one of them hides a notarization blocker that this slice created.

Everything below was re-run on this machine, not read off the PR body.

---

### What holds up (verified independently)

**1. Manual signing produces a correct artifact — CONFIRMED.**

```
$ APPLE_TEAM_ID=XCAA3WMJM6 bash apps/desktop/native/macos/FinderSyncExtension/build-appex.sh
    Signing Identity:     "Developer ID Application: Manuel Sanchez (XCAA3WMJM6)"
** BUILD SUCCEEDED **

$ codesign -dv --verbose=4 build/DropboxSyncFinderSync.appex
CodeDirectory v=20500 ... flags=0x10000(runtime)
Authority=Developer ID Application: Manuel Sanchez (XCAA3WMJM6)
Authority=Developer ID Certification Authority
Authority=Apple Root CA
TeamIdentifier=XCAA3WMJM6
```

**2. The unsigned path is genuinely unaffected — CONFIRMED.** Nothing exported (verified no ambient
`APPLE_*`/`DEVELOPMENT_TEAM` in the environment):

```
No APPLE_TEAM_ID/DEVELOPMENT_TEAM set: ad-hoc signing (fine for development and CI, not notarizable).
    Signing Identity:     "-"
** BUILD SUCCEEDED **
CodeDirectory v=20500 ... flags=0x10002(adhoc,runtime)
Signature=adhoc
TeamIdentifier=not set
```

**3. `--timestamp=none` — CONFIRMED.** It is literal in the invocation Xcode emits, and the artifact
reports `Signed Time=` (local CMS time), *not* `Timestamp=` (secure TSA timestamp):

```
/usr/bin/codesign --force --sign 25AABFD1... -o runtime --entitlements ....xcent \
  --requirements ... --timestamp\=none --generate-entitlement-der .../DropboxSyncFinderSync.appex

$ codesign -dv --verbose=4 ... | grep -i 'timestamp\|signed time'
Signed Time=19 Aug 2026 at 15:03:02        # no "Timestamp=" line -> no secure timestamp
```

**4. The AC deviation on `DEVELOPMENT_TEAM = ""` — the author is right, I side with him.**
The CLI override demonstrably wins over the empty project setting:

```
$ xcodebuild ... -showBuildSettings DEVELOPMENT_TEAM=XCAA3WMJM6 | grep DEVELOPMENT_TEAM
    DEVELOPMENT_TEAM = XCAA3WMJM6
$ xcodebuild ... -showBuildSettings | grep DEVELOPMENT_TEAM
    (empty)
```

And the "other organisations' certs" risk is real but contained — this keychain holds three
`Developer ID Application` identities, and the team constrains selection rather than falling back:

```
$ xcodebuild ... DEVELOPMENT_TEAM=ZZZZZZZZZZ CODE_SIGN_STYLE=Manual CODE_SIGN_IDENTITY="Developer ID Application"
error: No signing certificate "Developer ID Application" found: No "Developer ID Application"
signing certificate matching team ID "ZZZZZZZZZZ" with a private key was found.
** BUILD FAILED **
```

Refusing that AC was correct. It should be recorded on #82 as a deliberate deviation, not left
only in the PR body.

**5. Shell robustness — CONFIRMED, no issues.** Exercised the exact block under `/bin/bash`
3.2.57 with `set -euo pipefail`: nothing set (ARGC=0, empty-array expansion survives `set -u`);
`DEVELOPMENT_TEAM` only (resolves, ARGC=3); `APPLE_TEAM_ID=` empty + `DEVELOPMENT_TEAM` set
(`:-` correctly treats empty as unset and falls through); both empty (ad-hoc);
`APPEX_CODE_SIGN_IDENTITY=` empty (falls back to the default); and a team containing spaces and
`;` (stays a single argv element, no word-splitting, no injection). Quoting of `$APPEX_IDENTITY`
is correct.

**6. No leak — CONFIRMED.** `git grep -c XCAA3WMJM6` -> no matches, exit 1. The only identity
strings in tracked files are the `Your Name (TEAMID)` placeholders.

**7. `tauri.conf.json` still parses — CONFIRMED.** JSON5 parse OK (`config-json5` is enabled on
both `tauri` and `tauri-build`), and `bundle.macOS` now reads
`{"minimumSystemVersion":"12.0","infoPlist":"Info.plist","files":{...}}` with no `signingIdentity`.

---

## CRITICAL

### C1 — The precedence claim is inverted. The env var wins over the config, not the reverse.

`apps/desktop/src-tauri/tauri.conf.json:6-8`, `apps/desktop/native/macos/FinderSyncExtension/README.md:57-59`,
and the commit message ("`tauri.conf.json` pinned `signingIdentity` to "-", which takes precedence
over APPLE_SIGNING_IDENTITY") all state the opposite of what the CLI does.

`tauri-cli-2.4.0/src/interface/rust.rs:1309-1317`:

```rust
let signing_identity = match std::env::var_os("APPLE_SIGNING_IDENTITY") {
  Some(signing_identity) => Some(signing_identity.to_str()...to_string()),
  None => config.macos.signing_identity,
};
```

The environment is checked **first**; the config is only the fallback. So `"signingIdentity": "-"`
never overrode `APPLE_SIGNING_IDENTITY` and never "silently pinned" anything for anyone who
exported it.

The same sentence is wrong about `project.pbxproj` too, and this PR proves it itself: the
xcodebuild command-line override beats the project value (see verification 4) — which is the whole
argument used to justify leaving `DEVELOPMENT_TEAM = ""`. README:57-59 asserts the exact opposite
about the same file, two paragraphs from the section that relies on it.

Removing the pin is still the right call — but for the real reason: with `"-"` committed and no env
var exported, a *release* build was silently ad-hoc signed. Please rewrite the rationale in all
three places; a confident, wrong precedence rule in a config comment is the kind of thing the next
person will act on.

### C2 — "With neither exported the build is ad-hoc signed" is false for the app bundle, and this is an untested behaviour change.

`README.md:82-83` and `tauri.conf.json:13-14`. Those measurements (`Signature=adhoc`,
`TeamIdentifier=not set`) are from the **appex**, and for the appex they are correct. For the
Tauri **app bundle** the statement is wrong. `tauri-bundler-2.4.0/src/bundle/macos/app.rs:110`:

```rust
if let Some(keychain) = super::sign::keychain(settings.macos().signing_identity.as_deref())? {
    ...  remove_extra_attr(...)?;  sign(&keychain, sign_paths, settings)?;  notarize(...)
}
```

and `sign.rs:19-40` returns `Ok(None)` when there is no identity and no `APPLE_CERTIFICATE`. So
before this PR (`signingIdentity: "-"`) `tauri build` ran `codesign -s -` over the `.app`, sealing
`Contents/PlugIns/DropboxSyncFinderSync.appex` into the bundle. After this PR, with nothing
exported, **the signing block is skipped entirely** — no bundle signature, no `CodeResources`
seal, and `xattr -crs` is skipped too. That is "unsigned", not "ad-hoc".

This matters beyond wording: an unsealed host app is a plausible reason for Finder to refuse to
register the extension, which is precisely the symptom DBSYNC-72 exists to fix. The Test Plan
covers `cargo test`, clippy, the frontend build and `build-appex.sh` — it never runs
`tauri build`, so the one path this change actually alters was not exercised. CI will still go
green (the build succeeds unsigned), so CI will not catch it either.

Please either (a) run `tauri build` with nothing exported, inspect
`target/release/bundle/macos/DropboxSyncDesktop.app` with `codesign -dv`, and document the real
result; or (b) keep an explicit ad-hoc fallback for the unsigned path. Either way the two
sentences must stop claiming the app is ad-hoc signed.

### C3 — The signed Release appex carries `com.apple.security.get-task-allow`. Notarization rejects that outright.

```
$ codesign -d --entitlements - build/DropboxSyncFinderSync.appex
<plist version="1.0"><dict><key>com.apple.security.get-task-allow</key><true/></dict></plist>
```

Cause: `CODE_SIGN_INJECT_BASE_ENTITLEMENTS = YES` (confirmed via `-showBuildSettings`), which is
the default for `xcodebuild build`; only `archive`/`-exportArchive` turns it off. It is visible in
the codesign line as the injected `....appex.xcent`.

Two consequences:

- Notarization fails with *"The executable requests the com.apple.security.get-task-allow
  entitlement"* — a **harder** blocker than the missing timestamp, and it is not mentioned anywhere
  in this PR.
- It neuters part of Hardened Runtime: any process can obtain the task port of the extension. For a
  Finder extension that reads user paths, shipping that is a security regression.

It also makes `README.md:30` ("no entitlements file") and `:36-37` misleading. Both are true of the
*sources* and false of the *artifact*, and the artifact is what Apple inspects.

Verified fix — same script, same place `SIGN_ARGS` is already built:

```bash
SIGN_ARGS+=(
  DEVELOPMENT_TEAM="$TEAM"
  CODE_SIGN_STYLE=Manual
  CODE_SIGN_IDENTITY="$APPEX_IDENTITY"
  CODE_SIGN_INJECT_BASE_ENTITLEMENTS=NO
  OTHER_CODE_SIGN_FLAGS="--timestamp"
)
```

Result on the produced artifact:

```
/usr/bin/codesign --force --sign 25AABFD1... --timestamp -o runtime --requirements ... (no --entitlements)
CodeDirectory v=20500 ... flags=0x10000(runtime)
Authority=Developer ID Application: Manuel Sanchez (XCAA3WMJM6)
Timestamp=19 Aug 2026 at 15:07:30
TeamIdentifier=XCAA3WMJM6
$ codesign -d --entitlements - ...   ->  (no entitlements)
```

Developer ID chain and hardened runtime kept, `Signed Time=` becomes `Timestamp=`, entitlement gone.

---

## WARNINGS

### W1 — F4's conclusion overreaches, and its open question is now answerable.

`README.md:77-80` / `build-appex.sh:45-47` say the appex "must be re-signed with `--timestamp`
before submission". As C3 shows, it does not have to be: two build settings produce a
notarizable appex directly. "Not notarizable **as produced**" is true; "must be re-signed" is a
prescription that will send Slice 4 down the more fragile path.

Meanwhile the part README:79 calls "still unverified" is now settled, and the answer is the bad one.
In `tauri-bundler-2.4.0/src/bundle/macos/app.rs`, `copy_custom_files_to_bundle()` (which is what
places the appex from `bundle.macOS.files`) is called at line 108 and its output is **never pushed
to `sign_paths`** — only frameworks, sidecars, `copy_binaries_to_bundle` results and the `.app`
itself are. And `tauri-macos-sign-2.1.0/src/keychain.rs:194-198` builds
`["--force", "-s", identity, "--options", "runtime"]` — no `--deep`, so signing the outer `.app`
does not re-sign nested bundles either.

**Tauri never re-signs the nested appex.** Whatever `build-appex.sh` emits is what gets submitted.
Fixing it at the source (C3) is therefore the only clean option, and Slice 4 should be told that
rather than left to re-discover it. Worth noting too that `app.rs:130` calls `notarize()`
automatically once credentials are present — so today's artifact would be submitted, and rejected,
without an explicit step being asked for.

### W2 — `README.md:63-67` — the example exports a variable the command ignores.

```bash
export APPLE_SIGNING_IDENTITY="..."   # Tauri, host app
export APPLE_TEAM_ID=TEAMID
npm run build:finder-sync
```

`build:finder-sync` runs `build-appex.sh`, which reads only `APPLE_TEAM_ID` /
`DEVELOPMENT_TEAM` / `APPEX_CODE_SIGN_IDENTITY`. `APPLE_SIGNING_IDENTITY` is consumed by
`npm run tauri build`. Show both commands, or drop the export from this snippet.

### W3 — `README.md:22` contradicts the new policy it now sits above.

"Open `DropboxSyncFinderSync.xcodeproj` to edit signing (**Signing & Capabilities** → Team)" — doing
that writes `DEVELOPMENT_TEAM` into `project.pbxproj`, i.e. exactly the committed team id the new
section at :57 forbids. Add the warning, since the PR is editing this file anyway.

### W4 — `tauri.conf.json:11` — "xcodebuild reads it as DEVELOPMENT_TEAM" is not true.

`xcodebuild` has no knowledge of `APPLE_TEAM_ID`; `build-appex.sh:36` translates it. Small, but it
is the kind of half-truth that makes someone drop the script from the pipeline.

---

## SUGGESTION

### S1 — `build-appex.sh:53` — stale comment.

> `# With set -u, a plain "${TEAM_ARGS[@]}" can fail when the array is empty (macOS Bash 3.2).`

`TEAM_ARGS` no longer exists; it is `SIGN_ARGS`. The note itself is accurate and worth keeping.

---

## Not re-litigated

The no-sandbox ADR itself (accepted on DBSYNC-72) — only its wording is critiqued above, in C3.
DBSYNC-73, DBSYNC-74 and DBSYNC-75 are already filed and out of scope here.

## Summary

The diff does the right thing; the reasoning attached to it is wrong in a way that will mislead
whoever touches signing next, and Slice 2 shipped a signed artifact that Apple will reject for a
reason this PR does not mention. C1 and C2 are documentation and verification fixes. C3 is a
five-word change to `SIGN_ARGS`, already verified working. Once those land I would expect this
to go straight through.

_(Auth by Claude - architect-reviewer. Not approving: the maintainer stays the approver of record.)_

mbsanchez:
## All three criticals fixed in 5acce04 — and verified against the versions this repo actually uses

Before acting I checked the review's own basis. The cited sources were `tauri-cli-2.4.0`, `tauri-bundler-2.4.0` and `tauri-macos-sign-2.1.0`, but this project builds with the npm prebuilt `@tauri-apps/cli` **2.10.1** → `tauri-bundler` **2.8.1**. I fetched those two crates and re-read the same code paths. **All three findings hold on the real versions.** Details below, since the version gap could easily have gone the other way.

### C1 — confirmed, and it invalidates finding F2 of the ticket, not just my comment

`tauri-cli-2.10.1/src/interface/rust.rs:1450`:

```rust
let signing_identity = match std::env::var_os("APPLE_SIGNING_IDENTITY") {
  Some(signing_identity) => Some(...),   // env var wins
  None => config.macos.signing_identity, // config is only the fallback
};
```

Identical to 2.4.0. So DBSYNC-72's F2 — *"a value present in the config takes precedence over the `APPLE_SIGNING_IDENTITY` env var, so a real Developer ID identity would be silently ignored"* — has been wrong since the 2026-08-18 audit, and I propagated it into the config comment and the README. The `"-"` never blocked anything.

### C2 — confirmed, and it makes my change a regression

`tauri-bundler-2.8.1/src/bundle/macos/app.rs:107-110`:

```rust
} else if let Some(keychain) = super::sign::keychain(settings.macos().signing_identity.as_deref())? {
    ... sign the app ...
}
```

No identity resolving ⇒ the whole block is skipped ⇒ the `.app` gets **no signature and no `CodeResources` seal** over `Contents/PlugIns/*.appex`. Previously the `"-"` produced an ad-hoc signature that did seal it.

Combining C1 and C2 reverses the conclusion of this PR: **removing `signingIdentity` was solving a problem that did not exist while creating one that does.** It is restored, and the comment now explains why it must stay.

### C3 + the timestamp — fixed at the source, not by a later re-sign

```
CODE_SIGN_INJECT_BASE_ENTITLEMENTS=NO
OTHER_CODE_SIGN_FLAGS="--timestamp"
```

Artifact, with `APPLE_TEAM_ID` exported:

```
flags=0x10000(runtime)
Authority=Developer ID Application: Manuel Sanchez (XCAA3WMJM6)
Timestamp=19 Aug 2026 at 17:18:21
TeamIdentifier=XCAA3WMJM6
$ codesign -d --entitlements - …/DropboxSyncFinderSync.appex
(nothing)
$ codesign --verify --strict …
valid on disk · satisfies its Designated Requirement
```

With nothing exported: `flags=0x10002(adhoc,runtime)`, `Signature=adhoc`, `TeamIdentifier=not set` — CI unchanged.

### W1 — agreed, and F4 is now closed

`copy_custom_files_to_bundle()` (app.rs:105) never pushes its output to `sign_paths`, and the bundler's `codesign` call has no `--deep`. **Tauri does not re-sign the nested appex**, so its signature is final and the two settings above are mandatory rather than nice-to-have. The README says this where someone would otherwise lose an afternoon. Also noted: `notarize_auth()` (app.rs:127) auto-submits once credentials exist, so a bad signature would reach Apple unasked.

### W2, W3, W4, S1 — all fixed

Exporting `APPLE_SIGNING_IDENTITY` before `build:finder-sync` (that script reads only `APPLE_TEAM_ID`); the Xcode-GUI Team advice (it writes the committed `pbxproj` this design keeps empty); the `tauri.conf.json` claim that xcodebuild reads `APPLE_TEAM_ID` directly (the script translates it); the stale `TEAM_ARGS` comment.

### On the deviation the review sided with

`DEVELOPMENT_TEAM = ""` stays, and the reviewer's independent check is now recorded on issue #82 rather than only in this PR body.

### Re-validated

`cargo test` → 163 passed · `npm --workspace apps/desktop run build` → clean · `build-appex.sh` → BUILD SUCCEEDED in both modes.

One side note: running `cargo test` did reproduce DBSYNC-75 — it recreated the real `~/Library/Application Support/DropboxSyncDesktop/overlay_state.json` pointing at a deleted tempdir. Deleted again here. That file must not be present when Slice 3 (#83) checks badges, or the result will be a false negative.

mbsanchez (commented):
# CTO review — round 2 (commit `5acce04`)

Re-reviewed adversarially against the fix, not the original. Everything below was verified by
building the artifact and reading `tauri-cli` 2.10.1 / `tauri-bundler` 2.8.1 / `tauri-macos-sign`
2.3.3 — the last of which I fetched myself, because it is the crate that actually issues `codesign`
(see W1).

**Verdict: APPROVED.** All three criticals are genuinely resolved and the replacement claims hold
up. Three warnings below are new, introduced by this round; none of them makes the shipped artifact
wrong when the documented procedure is followed, so I would take them as a follow-up rather than a
third round.

---

## Round-1 findings — status

| # | Status | Evidence |
|---|---|---|
| C1 precedence backwards | **Resolved** | `interface/rust.rs:1450` — `match var_os("APPLE_SIGNING_IDENTITY") { Some(x) => Some(x), None => config.macos.signing_identity }`. Env wins, config is the fallback. The new wording in `tauri.conf.json:6-8` and `README.md:81-84` is exactly right. |
| C2 removing `signingIdentity` kills the signing block | **Resolved** | `tauri-bundler-2.8.1/.../macos/app.rs:107` gates on `else if let Some(keychain) = sign::keychain(...)`, and `sign::keychain(None)` returns `Ok(None)` (`sign.rs:37-42`). Whole block skipped ⇒ no signature, no `CodeResources`. `"-"` restored at `tauri.conf.json:64`. |
| C3 `get-task-allow` on the signed appex | **Resolved** | Signed build now emits `codesign --force --sign 25AAB… --timestamp -o runtime --requirements … --generate-entitlement-der` with **no `--entitlements`**, and `codesign -d --entitlements -` prints nothing. |
| W1 "Tauri must re-sign later" | **Resolved** | `copy_custom_files_to_bundle()` (`app.rs:175-201`) never touches `sign_paths`, and `Keychain::sign` (`tauri-macos-sign-2.3.3/src/keychain.rs:207-248`) builds `["--force","-s",id]` + optional `--options runtime` + optional `--entitlements` only. No `--deep` anywhere in the crate. Claim is now a fact, correctly stated. |
| W2 / W3 / W4 / S1 | **Resolved** | `README.md:88-93` splits the two commands; `README.md:24` adds the "do not set the Team in the GUI" warning; `tauri.conf.json:11` now credits `build-appex.sh` for the translation; `build-appex.sh:63` says `SIGN_ARGS`. |

## Independent verification of the new claims

Both build modes, run here:

**`APPLE_TEAM_ID=XCAA3WMJM6` exported** — every claim in the PR body confirmed verbatim:

```
CodeDirectory v=20500 … flags=0x10000(runtime)
Authority=Developer ID Application: Manuel Sanchez (XCAA3WMJM6)
Authority=Developer ID Certification Authority
Authority=Apple Root CA
Timestamp=19 Aug 2026 at 17:53:46
TeamIdentifier=XCAA3WMJM6
Sealed Resources version=2 rules=13 files=3
$ codesign -d --entitlements -   → (nothing)
$ codesign --verify --strict     → valid on disk · satisfies its Designated Requirement
```

Worth recording that this resolved the **right** certificate: the keychain holds three
`Developer ID Application` identities from three different organisations, and `DEVELOPMENT_TEAM`
disambiguated correctly to `XCAA3WMJM6`. The `tauri.conf.json:20-21` warning about not hardcoding an
identity on this machine is not hypothetical.

**Nothing exported** — `flags=0x10002(adhoc,runtime)`, `Signature=adhoc`, `TeamIdentifier=not set`,
`--verify --strict` passes. As claimed.

Three control experiments, because the justifications deserved to be falsified rather than believed:

1. *Is `CODE_SIGN_STYLE=Manual` really needed?* `DEVELOPMENT_TEAM=XCAA3WMJM6` alone →
   `error: No signing certificate "Mac Development" found: No "Mac Development" signing certificate
   matching team ID "XCAA3WMJM6" with a private key was found.` `README.md:95-101` quotes this
   correctly.
2. *Is `--timestamp` really needed for a Developer ID identity?* Same Manual/Developer ID build with
   `OTHER_CODE_SIGN_FLAGS` omitted → Xcode appends `--timestamp=none` even for Developer ID, and the
   product carries `Signed Time=19 Aug 2026 at 17:57:20` instead of `Timestamp=`. Not a secure TSA
   timestamp. The setting is load-bearing, exactly as documented.
3. *Does `CODE_SIGN_INJECT_BASE_ENTITLEMENTS=NO` drop anything the extension needs?* The injected
   `.xcent` contains **only** `com.apple.security.get-task-allow` — a debugger-attach entitlement,
   not a `FIFinderSync` requirement. Nothing else is lost. No `CODE_SIGN_ENTITLEMENTS` key and no
   `.entitlements` file exists under `apps/desktop`, and `ENABLE_HARDENED_RUNTIME = YES` in both
   configurations, so the ADR's description of the project matches the project.

---

## WARNINGS (new — introduced this round)

### W1 — `build-appex.sh:37`, `README.md:99` — the citation does not check out (the fact does)

> `the bundler's own codesign call carries no --deep (verified in tauri-bundler 2.8.1 …)`

`tauri-bundler` 2.8.1 contains no `codesign` invocation for app bundles at all — it delegates to
`tauri-macos-sign`, pinned at `2.3.3` (`tauri-bundler-2.8.1/Cargo.toml:176-177`). The only `--deep`
string in the bundler is in `dmg/bundle_dmg:613`, an unrelated verification call. Round 1's whole
complaint was version-inaccurate sourcing, so a source that cannot contain the checked fact is worth
correcting: cite `tauri-macos-sign 2.3.3 keychain.rs`. I verified the conclusion there myself and it
holds — no `--deep`, no `--timestamp`.

### W2 — `README.md:38` — the verification instruction is false on the default build path

> `codesign -d --entitlements - build/DropboxSyncFinderSync.appex` must print nothing.

Stated unconditionally, but `CODE_SIGN_INJECT_BASE_ENTITLEMENTS=NO` is only added inside the
`if [[ -n "$TEAM" ]]` branch (`build-appex.sh:50-56`). On the ordinary developer path — no
`APPLE_TEAM_ID` — the command prints:

```
[Dict]
	[Key] com.apple.security.get-task-allow
	[Value]
		[Bool] true
```

That is the build most people will run, and this sentence tells them it is broken. Qualify it with
"on a signed build", or fix it by hoisting the setting (see W3).

### W3 — `build-appex.sh:50-56` — half-configured env still produces a non-notarizable bundle

`APPLE_SIGNING_IDENTITY` and `APPLE_TEAM_ID` are independent, and only the latter reaches the appex.
Export the signing identity but forget the team and you get a Developer ID-signed host app
containing an **ad-hoc appex that still carries `get-task-allow`** — Tauri does not re-sign it, as
this PR correctly establishes. That combination is not merely theoretical: `notarize_auth()`
(`tauri-bundler-2.8.1/.../macos/sign.rs:95-125`) accepts the API-key triple
`APPLE_API_KEY` / `APPLE_API_ISSUER` / `APPLE_API_KEY_PATH` **without** `APPLE_TEAM_ID`, so the
bundler will happily auto-submit that bundle to Apple.

Two cheap options: move `CODE_SIGN_INJECT_BASE_ENTITLEMENTS=NO` out of the conditional so a Release
appex never carries a debug entitlement in any mode, and/or warn when `APPLE_SIGNING_IDENTITY` is set
while `APPLE_TEAM_ID` is not. Not a blocker — the README does say to export both — but it is the one
remaining path to the exact outcome this ticket exists to prevent.

---

## SUGGESTIONS

### S1 — `README.md:100-102` — "auto-submits … as soon as credentials are present" is missing a precondition

The `notarize_auth()` call sits **inside** the `else if let Some(keychain)` arm (`app.rs:110-133`).
With credentials present but no identity resolving, nothing is signed and nothing is submitted. In
context the sentence is about a signed build so it is not misleading in practice, but "once an
identity resolves **and** credentials are present" is the accurate form.

### S2 — `build-appex.sh:48,59` — `APPEX_CODE_SIGN_IDENTITY` is silently dropped without a team

`APPEX_IDENTITY` is computed unconditionally but only consumed in the team branch. Setting only
`APPEX_CODE_SIGN_IDENTITY` yields an ad-hoc build and the message `No APPLE_TEAM_ID/DEVELOPMENT_TEAM
set: ad-hoc signing`, which never mentions that the identity the user asked for was ignored.

### S3 — shell robustness: clean, with one non-shell caveat

I probed `APPEX_CODE_SIGN_IDENTITY='Nope $(id) `id` ; echo pwned'`. Bash quoting is correct — no
command substitution, no word splitting, the value arrives as a single `argv` element, and the build
**fails loudly** (`error: No certificate for team 'XCAA3WMJM6' matching …`) instead of silently
degrading to ad-hoc. That is the property that matters and it holds. Note only that `xcodebuild`
applies its own `$(VAR)` build-setting expansion to `CODE_SIGN_IDENTITY` — `$(id)` came out empty —
so a value containing `$(` would be mangled by Xcode, not by the script. Irrelevant for real
certificate names; recorded so nobody re-derives it. `${SIGN_ARGS[@]+"${SIGN_ARGS[@]}"}` remains the
correct Bash 3.2 empty-array guard, and `:-` on both fallbacks handles empty strings.

---

## Test commands re-run

```
npm --workspace apps/desktop run build   → tsc + vite: 55 modules, built in 528ms, clean
cargo test --manifest-path apps/desktop/src-tauri/Cargo.toml
                                         → 163 passed; 0 failed
build-appex.sh (APPLE_TEAM_ID set)       → ** BUILD SUCCEEDED **, Developer ID
build-appex.sh (nothing set)             → ** BUILD SUCCEEDED **, adhoc
```

`cargo test` reproduced **DBSYNC-75** again: it wrote a real
`~/Library/Application Support/DropboxSyncDesktop/overlay_state.json` pointing at a deleted
`/var/folders/…/T/.tmpsswwW1/synced`. That directory was empty before I started; I deleted the file
afterwards and it is empty again. Confirming the author's warning — this must not be present when
Slice 3 (#83) checks badges.

No source files were modified by this review; the working tree is clean at `5acce04`.

_(Auth by Claude - architect-reviewer. Posting as a comment, not an approval: the maintainer stays
the approver of record.)_

mbsanchez:
## W1–W3 addressed in f970ddc — after the approval, so flagging it explicitly

The approval was for `5acce04`. This commit lands on top of it and only addresses the three warnings from that same review, so re-review is a judgement call rather than a formality.

**W3 was worth not deferring.** `CODE_SIGN_INJECT_BASE_ENTITLEMENTS=NO` now applies on every path, not just the signed one. Exporting `APPLE_SIGNING_IDENTITY` without `APPLE_TEAM_ID` produced a Developer ID-signed host wrapping an ad-hoc appex still carrying `get-task-allow`, and `notarize_auth()` accepts an ASC key without `APPLE_TEAM_ID` — so that artifact could be submitted to Apple automatically. The script now also warns loudly on that combination:

```
WARNING: APPLE_SIGNING_IDENTITY is set but APPLE_TEAM_ID is not.
         The host app will be signed for real while this appex stays ad-hoc, which cannot be
         notarized. Export APPLE_TEAM_ID as well.
```

`--timestamp` deliberately stays in the signed branch: an ad-hoc signature cannot be timestamped.

**W2** is now true rather than reworded: the ad-hoc build reports zero entitlements, so "must print nothing" holds on every path. The README also says what a non-empty result implies — built by a plain `xcodebuild` or Xcode.app, and not shippable.

**W1** — citation corrected to `tauri-macos-sign` 2.3.3. You were right to call it out even though the conclusion held: it is the same class of error round 1 was rejected for, and it does not get a pass for landing on the right answer.

Verified: ad-hoc → `adhoc,runtime`, zero entitlements. Signed → `flags=0x10000(runtime)`, Developer ID chain, `Timestamp=19 Aug 2026 at 18:01:12`, `TeamIdentifier=XCAA3WMJM6`, zero entitlements, `--verify --strict` passes. `cargo test` 163 passed; frontend clean. The `overlay_state.json` that `cargo test` writes (DBSYNC-75) was deleted again.

PR #89, opened 2026-08-19
fix(#DBSYNC-76): resolvable principal class and an AppKit run loop for the appex

mbsanchez:
## This PR does not fix the ticket's symptom, and that is by design

macOS still does not register the extension after this change:

```
$ pluginkit -m -A -i com.mobsanchez.dropboxsyncdesktop.findersync
(nothing)
```

That negative result is an **explicit acceptance criterion** of DBSYNC-76 Slice 1, and confirming it matters: it means the analysis holds and the root-cause experiment (Slice 2) is still required. Had it registered, that would have been a surprise worth chasing — `NSPrincipalClass` is a plist key and `pkd` does read the plist, so the possibility was real.

What this PR fixes is two **separate, confirmed** defects queued behind the registration bug.

### 1. The principal class was unresolvable

`SyncExtension.swift:9` declares `@objc(DropboxSyncFinderSync)`. That attribute **replaces** the Swift-mangled runtime name with the bare literal. Measured on the binary, not inferred:

```
$ nm -a .../DropboxSyncFinderSync | grep -c _TtC        → 0
$ nm -a .../DropboxSyncFinderSync | grep OBJC_CLASS     → _OBJC_CLASS_$_DropboxSyncFinderSync
```

The plist asked for `$(PRODUCT_MODULE_NAME).DropboxSyncFinderSync`, so `NSClassFromString(...)` returned nil. **The extension could never have loaded, even with registration working.** Dropbox's `garcon.appex` (`EFFinderExtension`) and OneDrive's `FinderSync.appex` (`FinderSync`) both ship bare names.

### 2. No AppKit run loop in the extension host

Neither `NSPrincipalClass = NSApplication` nor `LSUIElement` was declared, while `SyncExtension.swift:25` schedules a repeating `Timer` from `init()` — which never fires without an AppKit run loop. Badges would not have refreshed even after a successful load. Both working references declare both keys.

## Why before the investigation rather than after

These two make the next step interpretable. When Slice 2 puts our appex under an Xcode-generated host to isolate whether `pkd` rejects the appex or the host, a failure should point at something real — not at a class that was never resolvable to begin with.

Left unfixed, they would have surfaced immediately after the registration fix as *"it registers but does nothing"*, which reads like the fix having failed and costs a diagnosis cycle.

The trade-off is real and was accepted deliberately: this is a PR that goes through review without resolving the headline bug.

## Verification

| Check | Result |
|---|---|
| plist `NSExtensionPrincipalClass` vs `nm` OBJC_CLASS symbol | `DropboxSyncFinderSync` = `DropboxSyncFinderSync` — match |
| `NSPrincipalClass` / `LSUIElement` in the built appex | `NSApplication` / `true` |
| `build-appex.sh` with `APPLE_TEAM_ID` exported | BUILD SUCCEEDED |
| Artifact signature | Developer ID chain, `flags=0x10000(runtime)`, `Timestamp=19 Aug 2026 at 19:40:56`, `TeamIdentifier=XCAA3WMJM6` |
| Entitlements on the artifact | 0 bytes |
| `codesign --verify --strict` | valid on disk, satisfies its Designated Requirement |
| `pluginkit` registration | still absent — **expected**, see above |

Everything was checked against the built product. That distinction has already caught two real defects in DBSYNC-72 that the project files denied.

No Rust or TypeScript file is touched, so `cargo test` and the frontend build are unaffected by this change.

`Info.plist.example` is updated in step: the README declares it a mirror of the committed plist, and leaving it behind would have it documenting the wrong class name.

## Jira

DBSYNC-76 — https://manuelbsan.atlassian.net/browse/DBSYNC-76 (Slice 1 of 4)

Next: Slice 2 — build a minimal Xcode-generated Finder Sync extension, confirm it registers on this Mac, then cross-swap our appex under its host and its appex under ours. Two builds isolate the axis every experiment so far has held constant.

### DIFF
--- a/apps/desktop/native/macos/FinderSyncExtension/Info.plist
+++ b/apps/desktop/native/macos/FinderSyncExtension/Info.plist
@@ -29,6 +29,13 @@
 	<true/>
 	<key>NSExtension</key>
 	<dict>
+		<!-- Empty on purpose. Apple's Finder Sync template emits it, and all three working
+		     extensions on a reference Mac carry it. Adding it alone was tested and does NOT fix
+		     registration (DBSYNC-76), but leaving it out would be the last structural difference
+		     between our NSExtension dict and every known-good one — an uncontrolled variable for
+		     the cross-swap experiment in Slice 2. -->
+		<key>NSExtensionAttributes</key>
+		<dict/>
 		<key>NSExtensionPointIdentifier</key>
 		<string>com.apple.FinderSync</string>
 		<!-- BARE class name, not module-qualified. SyncExtension.swift declares
--- a/apps/desktop/native/macos/FinderSyncExtension/Info.plist
+++ b/apps/desktop/native/macos/FinderSyncExtension/Info.plist
@@ -20,12 +20,25 @@
 	<string>$(MARKETING_VERSION)</string>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<!-- The extension host needs an AppKit run loop: SyncExtension.swift schedules a repeating
+	     Timer from init(), which never fires without one. Both Dropbox's garcon.appex and
+	     OneDrive's FinderSync.appex declare these two keys. -->
+	<key>NSPrincipalClass</key>
+	<string>NSApplication</string>
+	<key>LSUIElement</key>
+	<true/>
 	<key>NSExtension</key>
 	<dict>
 		<key>NSExtensionPointIdentifier</key>
 		<string>com.apple.FinderSync</string>
+		<!-- BARE class name, not module-qualified. SyncExtension.swift declares
+		     @objc(DropboxSyncFinderSync), and @objc(...) REPLACES the Swift-mangled runtime name
+		     with that literal: the binary exports _OBJC_CLASS_$_DropboxSyncFinderSync and no _TtC
+		     symbol at all, so "$(PRODUCT_MODULE_NAME).DropboxSyncFinderSync" resolves to nil.
+		     Keep both in sync — verify with:
+		       nm -a build/DropboxSyncFinderSync.appex/Contents/MacOS/DropboxSyncFinderSync | grep OBJC_CLASS -->
 		<key>NSExtensionPrincipalClass</key>
-		<string>$(PRODUCT_MODULE_NAME).DropboxSyncFinderSync</string>
+		<string>DropboxSyncFinderSync</string>
 	</dict>
 </dict>
 </plist>
--- a/apps/desktop/native/macos/FinderSyncExtension/Info.plist.example
+++ b/apps/desktop/native/macos/FinderSyncExtension/Info.plist.example
@@ -26,6 +26,8 @@
 	<true/>
 	<key>NSExtension</key>
 	<dict>
+		<key>NSExtensionAttributes</key>
+		<dict/>
 		<key>NSExtensionPointIdentifier</key>
 		<string>com.apple.FinderSync</string>
 		<!-- Bare class name: @objc(DropboxSyncFinderSync) replaces the mangled Swift name. -->
--- a/apps/desktop/native/macos/FinderSyncExtension/Info.plist.example
+++ b/apps/desktop/native/macos/FinderSyncExtension/Info.plist.example
@@ -20,12 +20,17 @@
 	<string>$(MARKETING_VERSION)</string>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<key>NSPrincipalClass</key>
+	<string>NSApplication</string>
+	<key>LSUIElement</key>
+	<true/>
 	<key>NSExtension</key>
 	<dict>
 		<key>NSExtensionPointIdentifier</key>
 		<string>com.apple.FinderSync</string>
+		<!-- Bare class name: @objc(DropboxSyncFinderSync) replaces the mangled Swift name. -->
 		<key>NSExtensionPrincipalClass</key>
-		<string>$(PRODUCT_MODULE_NAME).DropboxSyncFinderSync</string>
+		<string>DropboxSyncFinderSync</string>
 	</dict>
 </dict>
 </plist>
--- a/apps/desktop/native/macos/FinderSyncExtension/build-appex.sh
+++ b/apps/desktop/native/macos/FinderSyncExtension/build-appex.sh
@@ -79,6 +79,21 @@ xcodebuild \
 
 OUT="${DERIVED}/Build/Products/${CONFIG}/${SCHEME}.appex"
 echo "Built: $OUT"
+
+# Guard the coupling between SyncExtension.swift's @objc(...) name and the plist's principal class.
+# Nothing else catches a drift: removing the @objc attribute changes the emitted ObjC symbol to the
+# Swift-mangled form and the plist silently stops resolving, with no build error and no test failure.
+# This ran green on macos-latest in CI via `tauri build` -> beforeBuildCommand -> build:finder-sync.
+if [[ -d "$OUT" ]]; then
+  DECLARED=$(/usr/libexec/PlistBuddy -c "Print :NSExtension:NSExtensionPrincipalClass" "$OUT/Contents/Info.plist" 2>/dev/null || true)
+  EMITTED=$(nm -a "$OUT/Contents/MacOS/${SCHEME}" 2>/dev/null | sed -n 's/.*_OBJC_CLASS_\$_//p' | head -1)
+  if [[ -z "$EMITTED" || "$DECLARED" != "$EMITTED" ]]; then
+    echo "ERROR: NSExtensionPrincipalClass '${DECLARED}' does not match the ObjC class the binary exports ('${EMITTED:-none}')." >&2
+    echo "       NSClassFromString would return nil and the extension could never load. See DBSYNC-76." >&2
+    exit 1
+  fi
+  echo "Principal class OK: ${DECLARED}"
+fi
 if [[ -d "$OUT" ]]; then
   mkdir -p "${ROOT}/build"
   rm -rf "${ROOT}/build/${SCHEME}.appex"


mbsanchez (commented):
## CTO review — DBSYNC-76 Slice 1 — **APPROVED** (0 critical, 3 warnings, 2 suggestions)

Stack: `desktop-tauri` (CLAUDE.md override). Reviewed at `1cc8377`. Every claim below was re-derived
locally on macOS 26.5.1 / Xcode 15.4 with a **from-scratch build into a clean derived-data path**, not
from the incremental artifact already in the tree.

---

### 1. Bare principal class — **VERIFIED CORRECT**

```
$ nm -a .../Release/DropboxSyncFinderSync.appex/Contents/MacOS/DropboxSyncFinderSync \
    | grep -E 'OBJC_CLASS_\$_Dropbox|_TtC'
0000000000000000 - 00 0000  GSYM _OBJC_CLASS_$_DropboxSyncFinderSync
000000010000c1a8 s          _OBJC_CLASS_$_DropboxSyncFinderSync

$ nm -a ... | grep -c '_TtC'
0

$ /usr/libexec/PlistBuddy -c "Print :NSExtension:NSExtensionPrincipalClass" .../Contents/Info.plist
DropboxSyncFinderSync
```

Exact string match, and **zero** `_TtC` symbols — so there is no mangled name for the old value to
resolve to. The old value is confirmed dead, not merely suspicious:

```
$ xcodebuild -showBuildSettings | grep PRODUCT_MODULE_NAME
    PRODUCT_MODULE_NAME = DropboxSyncFinderSync
```

i.e. the shipped plist previously asked the runtime for `DropboxSyncFinderSync.DropboxSyncFinderSync`,
a class that appears nowhere in the binary. `NSClassFromString` returned nil. The defect is real and
the fix is right.

One nuance the PR body flattens, worth recording because it will come up again: module-qualified is not
inherently wrong. odrive's Finder Sync extension is registered and enabled on this machine with
`NSExtensionPrincipalClass = FinderSyncExtension.FinderSync`, and Xcode 15.4's own Finder Sync template
generates a module-qualified name (`___VARIABLE_moduleNamePrefixForClasses___FinderSync`). Bare is
correct *here* specifically because `@objc(DropboxSyncFinderSync)` renames the class. The two facts have
to travel together — which leads to the one thing I do want fixed:

---

### ⚠️ Warning 1 — nothing catches the `@objc` ↔ plist coupling regression

The `Info.plist` comment tells a human to run `nm` by hand. Nobody will. Delete `@objc(...)` from
`SyncExtension.swift:9` — a plausible cleanup, since it looks redundant next to `final class` — and the
plist silently becomes wrong again, with the same nil-class failure and no build error.

This is cheap to automate and it is already CI-enforced infrastructure: `beforeBuildCommand` →
`tauri-before-build.mjs` → `npm run build:finder-sync` on `darwin`, and `.github/workflows/ci.yml` runs
`npm run tauri build` on `macos-latest` for every PR. So a post-build assertion in `build-appex.sh`
(after the `OUT` block, line ~80) runs on every pull request, signed or ad-hoc:

```bash
PRINCIPAL="$(/usr/libexec/PlistBuddy -c 'Print :NSExtension:NSExtensionPrincipalClass' "$OUT/Contents/Info.plist")"
if ! nm -a "$OUT/Contents/MacOS/${SCHEME}" | grep -q "_OBJC_CLASS_\$_${PRINCIPAL}$"; then
  echo "ERROR: NSExtensionPrincipalClass '${PRINCIPAL}' has no matching ObjC class in the binary." >&2
  echo "       Either restore @objc(${PRINCIPAL}) in SyncExtension.swift or update Info.plist." >&2
  exit 1
fi
```

Six lines, and it is exactly the check the comment already documents. Non-blocking for this PR, but it
should not slip past Slice 2 — Slice 2 will be swapping hosts and appexes around, which is precisely when
a silent principal-class regression would be most expensive to diagnose.

---

### 2. `NSPrincipalClass = NSApplication` + `LSUIElement` — safe, correctly paired, but unevenly evidenced

**Safety: confirmed.** All three Finder Sync extensions registered on this machine declare both keys:

| appex | `NSPrincipalClass` | `LSUIElement` | registered |
|---|---|---|---|
| `Dropbox.app/.../garcon.appex` | `NSApplication` | `true` | yes |
| `OneDrive.app/.../FinderSync.appex` | `NSApplication` | `true` | yes |
| `odriveapp.app/.../FinderSyncExtension.appex` | `NSApplication` | `true` | yes (enabled) |

So the pair cannot plausibly *cause* a registration failure, which is the only way this PR could have
made Slice 2 harder. It doesn't.

**`LSUIElement` is not cargo-cult.** It is in Apple's own template, which is the strongest possible
answer to that question:

```
$ grep -n 'LSUIElement' "…/Templates/Project Templates/macOS/Application Extension/Finder Sync Extension.xctemplate/TemplateInfo.plist"
94:<key>LSUIElement</key>
```

It is also the necessary companion to the other key rather than an independent one: if `NSApplication`
does get instantiated in the extension process, `LSUIElement` is what keeps AppKit's activation policy at
accessory instead of giving the appex a Dock tile and a menu bar. Adding `NSPrincipalClass` *without*
`LSUIElement` would have been the risky combination. Adding both together is correct.

**`NSPrincipalClass = NSApplication` is *not* in Apple's template** (`NSPrincipalClass` appears in the
Screen Saver, Installer Plug-in, Preference Pane and Bundle templates — never in the Finder Sync one).
Its entire justification is third-party convention. That is decent evidence, and I would keep the key.
But it means the causal claim in the plist comment and in the PR body is asserted, not established.

---

### ⚠️ Warning 2 — the run-loop claim is stated as fact but has never been observed

The PR body says the `Timer` "never fires without an AppKit run loop" and that "badges would not have
refreshed even after a successful load". Neither is measurable today: the extension has **never loaded
once**, so nobody has watched a timer fail to fire. Everywhere else this PR is scrupulous about
measuring rather than inferring — this is the one place it doesn't, and it happens to be the half of the
change with the weaker evidence. Please downgrade it to a hypothesis in the plist comment.

Checking the mechanism against how the timer is actually scheduled (`SyncExtension.swift:25`):

```swift
reloadTimer = Timer.scheduledTimer(withTimeInterval: 2.0, repeats: true) { … }
```

`Timer.scheduledTimer(withTimeInterval:repeats:block:)` installs on the **current** run loop in the
**default** mode. So the real precondition is not "AppKit" — it is "the thread that ran `init()` later
pumps *its own* run loop in default mode". `NSApp.run()` satisfies that; so would a plain
`CFRunLoopRun()`. If the host instead parks the main thread in `dispatch_main()`, the timer is dead
whatever the plist says. And if the principal class is ever instantiated off the main thread, the timer
attaches to a run loop nobody runs, and `NSPrincipalClass` does not help at all.

### 💡 Suggestion 1 — make the timer independent of the host's run-loop policy

An Info.plist key is a fragile way to keep application code alive. A dispatch source strictly dominates:
it fires whether the main thread runs an `NSRunLoop` or `dispatch_main()`, and it does not stall during
menu tracking — which matters here, because a `FIFinderSync` extension vends contextual menus, and a
default-mode `NSTimer` pauses in `NSEventTrackingRunLoopMode` while a menu is open.

```swift
private var reloadTimer: DispatchSourceTimer?

private func scheduleReload() {
    reloadTimer?.cancel()
    let timer = DispatchSource.makeTimerSource(queue: .main)
    timer.schedule(deadline: .now() + 2.0, repeating: 2.0)
    timer.setEventHandler { [weak self] in
        self?.reloadState()
        FIFinderSyncController.default().directoryURLs = self?.monitoredDirectoryURLs() ?? []
    }
    timer.resume()
    reloadTimer = timer
}
```

The minimum version is `RunLoop.main.add(timer, forMode: .common)`. Either way the behaviour stops
depending on a plist key whose effect nobody has been able to observe yet. Slice 3/4 material, not this PR.

---

### ⚠️ Warning 3 — the one key Apple's template *does* have was not added

The PR aligned the plist with `garcon.appex` / `FinderSync.appex` and added a key that is in neither
Apple template. It did not add the key that is in **all three** references *and* Apple's template:

```
Dropbox   garcon.appex          NSExtensionAttributes => { }
OneDrive  FinderSync.appex      NSExtensionAttributes => { }
odrive    FinderSyncExtension   NSExtensionAttributes => { }
Xcode 15.4 template             <key>NSExtensionAttributes</key><dict/>
ours      DropboxSyncFinderSync  — absent —
```

I don't think an empty `NSExtensionAttributes` is what blocks registration. But it is two lines, zero
risk, and it is the *only* remaining structural difference between our `NSExtension` dict and four
independent working examples — so leaving it out means Slice 2's cross-swap experiment runs with one
uncontrolled variable that this PR was perfectly placed to eliminate. Add it here or as the first line
of Slice 2.

---

### 3. Artifact unchanged in every other respect — **VERIFIED**

From-scratch `xcodebuild` into a clean derived-data path, with the flags `build-appex.sh` passes:

```
Identifier=com.mobsanchez.dropboxsyncdesktop.findersync
CodeDirectory v=20500 size=472 flags=0x10000(runtime) hashes=7+3 location=embedded
Authority=Developer ID Application: Manuel Sanchez (XCAA3WMJM6)
Authority=Developer ID Certification Authority
Authority=Apple Root CA
Timestamp=19 Aug 2026 at 19:53:58
TeamIdentifier=XCAA3WMJM6

$ codesign -d --entitlements - <appex>
Executable=…/DropboxSyncFinderSync            ← no entitlements printed

$ codesign --verify --strict --verbose=2 <appex>
valid on disk
satisfies its Designated Requirement
```

Full Developer ID chain, hardened runtime (`0x10000`), secure TSA timestamp (not `--timestamp=none`),
zero entitlements — the DBSYNC-72 ADR invariants all hold on the product. `build-appex.sh` itself is
untouched by this PR, and the plist edit does not perturb any of them.

---

### 4. Expected-negative claim — **VERIFIED, and localised further**

```
$ pluginkit -m -A -i com.mobsanchez.dropboxsyncdesktop.findersync
(no output)
```

It does not register. But "pluginkit prints nothing" is weak on its own, so I checked whether the host is
even visible to LaunchServices, and it is:

```
$ lsregister -dump | grep -E '^path: .*(dropboxsync|findersync)'
path: /Applications/OneDrive.app/Contents/PlugIns/FinderSync.appex
path: /Users/mbsanchez/Applications/odriveFinderSyncExtension/odriveapp.app
path: /Users/mbsanchez/Applications/odriveFinderSyncExtension/odriveapp.app/Contents/PlugIns/FinderSyncExtension.appex
path: /Users/…/target/release/bundle/macos/DropboxSyncDesktop.app
path: /Users/mbsanchez/.Trash/DropboxSyncDesktop.app
path: /Users/mbsanchez/.Trash/DropboxSyncDesktop 18.57.05.app
path: /Users/mbsanchez/.Trash/DropboxSyncDesktop 18.59.12.app
path: /Users/mbsanchez/.Trash/DropboxSyncDesktop 19.42.39.app
```

The host **is** registered. Every working reference has a **separate LS entry for the nested appex**.
Ours has none — LaunchServices registers the host and never descends into `Contents/PlugIns/`. So the
failure is upstream of PlugInKit entirely, which is consistent with the PR's framing and does not
weaken it. Handing that to Slice 2 rather than arguing it here.

The embedded copy in the bundled `.app` carries the new keys, so the negative was re-confirmed against
the fixed artifact and not a stale one:

```
$ PlistBuddy -c Print …/DropboxSyncDesktop.app/Contents/PlugIns/DropboxSyncFinderSync.appex/Contents/Info.plist
NSExtensionPrincipalClass = DropboxSyncFinderSync
NSPrincipalClass = NSApplication
LSUIElement = true
```

Unrelated to this PR but worth clearing before QA: four trashed `DropboxSyncDesktop*.app` copies are
still LS-registered and could shadow the build-path copy in future experiments.

---

### 5. `Info.plist` vs `Info.plist.example` — **VERIFIED CONSISTENT**

```
$ diff <(plutil -convert xml1 -o - Info.plist) <(plutil -convert xml1 -o - Info.plist.example)
(no differences)
```

Semantically identical; the raw diff is comments only (the example carries the short form). The README's
"mirror" contract holds.

### 💡 Suggestion 2

The 6-line comment block above `NSExtensionPrincipalClass` is good documentation of a genuinely
surprising fact, and I'd keep it. Once the `build-appex.sh` guard from Warning 1 exists, replace the
hand-run `nm` instruction with a pointer to the guard, so the comment stops implying a manual step that
is actually automated.

---

## On shipping a PR that knowingly does not fix the symptom

Right call, and I'd have pushed for it even if it hadn't been proposed.

The deciding fact is that the principal-class defect is **independently proven fatal** — `nm` says the
class the plist asks for does not exist in the binary, full stop. That isn't a speculative cleanup
bundled in for convenience; it is a second, certain blocker sitting behind the first one. Shipping it
now costs 20 reversible lines of metadata. Holding it costs correctness of the *next* experiment: Slice 2
cross-swaps our appex under an Xcode-generated host and vice versa, and the entire value of that
experiment is that a load failure means something. Run it with a principal class that can never resolve
and you have deliberately poisoned your own control.

The "ship together" instinct is usually right, and it is right when the partial fix is speculative,
unreviewable in isolation, or expands the blast radius of the eventual real fix. None of those apply:
this is metadata-only, fully verifiable against the built product without the extension ever loading,
and it strictly shrinks Slice 2's search space.

The one thing I'd change is not the decision but the framing. The PR presents the principal class
(measured, certain) and the run loop (inferred, never observed) at the same confidence, and that
evenness is what makes the "two confirmed defects" wording overreach slightly. Say "one confirmed defect
and one hypothesis with strong precedent" and the PR is exactly as strong as its evidence — which is
still strong enough to merge.

---

**Verdict: APPROVED.** No blocking issues; nothing regressed; every load-bearing claim reproduced
independently. Warnings 1 and 3 (build-time principal-class guard, `NSExtensionAttributes`) should be
picked up by Slice 2 at the latest. Warning 2 is a wording fix. Suggestion 1 is Slice 3/4.

mbsanchez:
## W1–W3 addressed in 50db09b, and a framing correction accepted

**W1 — guard added.** `build-appex.sh` now compares the plist's `NSExtensionPrincipalClass` against the symbol `nm` actually exports and fails the build on mismatch. Verified in both directions, because a guard that only passes when things are already correct proves nothing:

```
correct build      → Principal class OK: DropboxSyncFinderSync
plist tampered     → would fail with 'Modulo.Clase' vs 'DropboxSyncFinderSync'
```

Your point that this is already CI-enforced infrastructure is what made it worth doing rather than noting: `tauri build` → `beforeBuildCommand` → `build:finder-sync` runs on `macos-latest` per PR.

**W3 — `NSExtensionAttributes` added.** You are right that this is the one remaining structural delta against four working examples. It was tested in isolation earlier and does not fix registration — but that is not the argument. Leaving it out would put an uncontrolled variable into Slice 2's cross-swap, where the entire method depends on the two sides differing in exactly one respect. The plist comment records both facts so nobody re-tests it as a fix.

**W2 — framing corrected, and the criticism lands.** "Two confirmed defects" overstated the evidence, and it is precisely the overreach this ticket has been penalising in others:

- **Principal class — measured.** Zero `_TtC` symbols, one bare `_OBJC_CLASS_$_DropboxSyncFinderSync`. `NSClassFromString` returned nil. Not an inference.
- **AppKit run loop — inferred, with strong precedent.** The timer needs the thread that ran `init()` to later pump its own run loop in default mode; `NSApp.run()` satisfies that, but so would `CFRunLoopRun()`. Nobody has watched it fail, because the extension has never loaded.

Your nuance about module-qualified names is also correct and the PR flattened it: odrive registers fine with `FinderSyncExtension.FinderSync`, and Apple's template generates module-qualified for Swift. Bare is right *here* only because of the `@objc` rename — which is exactly why W1 needed a guard rather than a comment.

## The finding in your Claim-4 verification is bigger than the warnings

OneDrive and odrive each have a **separate LS entry for their nested appex**; ours has none, and LaunchServices never descends into `Contents/PlugIns/`.

That corrects something I asserted on DBSYNC-76: I wrote that "Launch Services HAS the extension" because its identifier and path appear in `lsregister -dump` — but that is metadata inside the host's record, not a registered plug-in bundle. The failure boundary is **upstream** of where I placed it: it is not that `pkd` rejects what LS hands it, it is that LS never hands it over. The ticket has been corrected and Slice 2 re-aimed accordingly.

Thank you for going past the claim you were asked to check.

PR #91, opened 2026-08-20
feat(#DBSYNC-80): badge .cloudsc placeholders as cloud_only

mbsanchez:
## Summary

Cloud-only files and folders had **no Finder badge at all**, so "online only" was indistinguishable from "not tracked" — the one state the user cannot otherwise see. They now carry a distinct grey download-arrow badge, verified live: 24 placeholders badged, 5 real files still showing the green check.

## Two independent causes, both fixed

**The rows did not exist.** `paths` was built from `list_local_files()`, and placeholders are not in `local_file_index`. Measured, not inferred: 24 sidecars on disk, 5 index rows, none of them a sidecar. They have to be enumerated from disk.

**The skip would have dropped them anyway.** `refresh_overlay_state_inner` skips anything ending in `.cloudsc`. **That skip is correct on Windows and stays**: under CfAPI the native placeholder IS the real file and the sidecar is noise, which the DBSYNC-41 comment says explicitly. It was never weighed for macOS, where the sidecar is all there is. The new emission is `cfg(not(windows))`, so Windows comes out byte-identical.

### Already tried and reverted — do not re-propose

Mapping `X.cloudsc → X` in the overlay writer. It was **inert** (no such rows exist to rewrite) *and* used the **wrong key**: `requestBadgeIdentifier(for:)` derives the path relative to the sync root exactly as the file system spells it and looks it up verbatim, so the suffix must stay on. `cloudsc_key_keeps_its_suffix_and_the_logical_name_is_absent` now pins that.

## Design decisions

**A distinct tier, not `Synced`.** A green check on a file that is not on disk hides the only fact the badge could usefully convey. `cloud_only` is additive, so `version` is **not** bumped: a reader that does not know the value finds no badge registered for it and renders nothing — exactly what it did before the value existed. The documented bump rule is for changes that make an existing reader *wrong*, not merely incomplete.

**A download arrow, not a cloud.** A cloud was built first and rejected on evidence, not taste: Finder draws badges at 16pt in list view, where the cloud silhouette collapses into an unreadable blob that reads as a dash. The arrow stays legible and matches what iCloud Drive shows for a file that is not downloaded. Grey rather than a status colour — green/blue/red mean in sync, in motion, wrong, and "online only" is none of those.

**A cheap enumeration.** `list_cloudsc_placeholder_rels` is a sibling of the existing `list_cloudsc_placeholders`, which parses every placeholder file to recover a `tag` and a remote path no badge uses. This runs on every pass of the sync queue (`refresh_queue_depth_internal` → `sync_pipeline.rs:746, 772, 898, 1107`), so those reads would be paid repeatedly. Folder placeholders are included — a dehydrated directory is itself a regular file holding the placeholder record, which is why the SQL derivation "remote index minus local index" was rejected: it only reasons about files.

**`compute_overlay_paths` split out** of `refresh_overlay_state_inner` so the decision is testable without writing to the real app data dir. Tests that called the outer function would clobber the running user's `overlay_state.json` — DBSYNC-75, which bit this very session (see below).

## Stack

`desktop-tauri`

## Jira Ticket

#DBSYNC-80

## Test Plan

- [x] `cargo test` — **170 passed**, 7 of them new and specific to this change:
  - placeholder emitted as `cloud_only`
  - key keeps its `.cloudsc` suffix, and the logical name is *absent*
  - dehydrated folder placeholder included
  - placeholder with an active job reads as `syncing`, not `cloud_only`
  - an indexed `.cloudsc` row never yields `Synced` (locks in the Windows skip on every platform)
  - a real file matching remote is still `Synced`
  - no sync folder configured yields an empty map rather than an error
- [x] `cargo clippy --all-targets` — no new warnings; the four that remain are pre-existing and in untouched files
- [x] Full `tauri build` with the Developer ID identity: appex keeps `app-sandbox` + the temporary exception, carries no `get-task-allow`, all 8 badge resources bundled, `codesign --verify --deep --strict` passes. This also re-confirms DBSYNC-72 F4 — Tauri did not re-sign the nested appex and its signature survived intact.

### Manual desktop validation (macOS Finder Sync)

`overlay_state.json` regenerated by the real app: 24 `cloud_only` + 5 `synced`. Finder screenshot of `~/DropboxSync/UNET` shows the grey arrow on every `.cloudsc` and the green check on the real files, with nothing else changed.

## Findings recorded while validating, not fixed here

- **A fresh install leaves the extension disabled.** After replacing the whole `.app`, `pluginkit` lists the extension as `-` (registered, not enabled). A real user must turn it on in System Settings; nothing in the app tells them to. Relevant to DBSYNC-72 Slice 4's clean-account AC.
- **DBSYNC-75 is not theoretical.** Running `cargo test` during this work overwrote the live `~/Library/Application Support/DropboxSyncDesktop/overlay_state.json` with a test fixture pointing at a deleted tempdir, silently blanking every badge until the app next refreshed. The split introduced here keeps the *new* tests clear of it, but the underlying bug is untouched.
- **Keychain prompts** — filed as #DBSYNC-81, unrelated to this change.

🤖 Generated with Claude Code


mbsanchez (commented):
## Code review — APPROVED, with one gap closed during review

Reviewed against the `desktop-tauri` and `rust` rows, plus the stack-agnostic checklist.

### The measurement the ticket demanded, which this PR did not have

DBSYNC-80 ends its Proposed Direction with: *"The performance question is the open one: whether one extra traversal per refresh is acceptable, or whether the placeholder set needs caching/invalidation. **Decide with a measurement on a realistic folder, not by assertion.**"*

The PR asserted the walk is cheap and never measured it. That is a real gap, because `refresh_overlay_state_internal` is reached from `refresh_queue_depth_internal`, which runs inside the sync queue loop — the cost is paid repeatedly, not once.

Measured with a synthetic tree (100 files per directory, one in five a `.cloudsc`), timing the same `WalkDir` the function uses:

| Files | Placeholders | Walk |
| --- | --- | --- |
| 1,000 | 200 | **1.0 ms** |
| 10,000 | 2,000 | **8.0 ms** |
| 50,000 | 10,000 | **40 ms** |

Linear, roughly **0.8 µs per file**.

**Verdict: acceptable, and the caching/invalidation alternative is correctly avoided as premature.** 40 ms on a 50,000-file folder is real but not pathological for a refresh that is not per-frame, and it sits beside a loop that already issues one `get_remote_file` query per indexed row — so the walk is very unlikely to be the dominant cost.

**Two caveats that belong in the record rather than in a footnote.** This is a **warm page cache** on a local APFS SSD; a cold cache, a network volume or a spinning disk will be materially worse, and none of those were tested. And it scales linearly with **total** files, not with placeholders — a folder with 200,000 tracked files and zero placeholders still pays the full traversal on every refresh. If a user ever reports sluggishness during large syncs, this is the first thing to look at, and the number to beat is in this table.

### Verified against current `main`, not just against the base it was branched from

A clean merge does not rule out a semantic conflict, and `main` has moved since this branch was cut (DBSYNC-78 landed as 57fe6c1). Merged `main` into a throwaway copy of the branch and ran the suite:

```
test result: ok. 174 passed; 0 failed
clippy: 0 errors
```

174 = 167 on `main` + the 7 added here. No interaction. Throwaway branch deleted.

### Checklist

| | |
| --- | --- |
| KISS / YAGNI | ✅ `list_cloudsc_placeholder_rels` is a deliberately thinner sibling of `list_cloudsc_placeholders`, not a generalisation of it. The rejected SQL derivation is documented with its reason (folder placeholders), which is more useful than the code that survived. |
| Scope | ✅ `overlay_state.rs`, `cloudsc_ops.rs`, and the Finder Sync extension — all declared. |
| Secrets | ✅ None. |
| Debug output | ✅ None. The `Logger` instrumentation used while diagnosing DBSYNC-76 was removed before commit and is not in this diff. |
| Error handling | ✅ The asymmetry is deliberate and documented: a walk error skips one entry (environmental), a `relpath_under` failure propagates (broken invariant, since the entry came from a walk rooted at that folder). That distinction is argued in the code, not accidental. |
| Security | ✅ No IPC surface added, no capability change, no injection vector. Paths come from a filesystem walk rooted at the configured sync folder, not from user input. |
| Rust idioms | ✅ clippy clean, no `unsafe`, no `unwrap()` outside tests, `Result`-first. |
| Tests | ✅ Seven, including the two that pin the previously-reverted mistake: the key keeps its `.cloudsc` suffix, and the logical name is asserted **absent**. |

### Observation — ignore globs are not applied (non-blocking, not introduced here)

`list_cloudsc_placeholder_rels` does not filter through `is_ignored_local_path`, while other walks in the same file do (`cloudsc_ops.rs:992`, `:1073`). A placeholder inside an ignored path would therefore get a badge.

**Not a defect this PR introduces:** its sibling `list_cloudsc_placeholders` omits the same filter, so the new function is consistent with the one it was modelled on. Impact is close to nil in practice — placeholders are created by the sync engine for tracked content, and ignored paths are not tracked, so such a file should not exist. Worth knowing that *both* functions share the gap if ignore-glob behaviour is ever revisited.

### Note on the versioning decision

`cloud_only` is added without bumping `version`, argued as additive: an older reader finds no badge registered for the value and renders nothing, which is what it did before the value existed. I agree, and the reasoning is written into the contract doc rather than left implicit — which is the right place for the next person to find it.

### Verdict

**Approved.** Next: `/cowork-workflow:jira-ticket-qa`

mbsanchez:

Verdict from the maintainer, looking at the real Finder window.

### What was validated, and against what

Deliberately **not** the branch in isolation: a build of PR #91 **merged with current `main`** — which now carries DBSYNC-78 — so the thing on screen is the post-merge state, not an artefact that disappears at merge time. Both features coexist correctly: the monochrome tray icon and the placeholder badges.

Installed with the full replacement plus `pluginkit -e use` and a Finder restart. Replacing the bundle alone leaves Finder talking to the extension instance it already holds, so `init()` never re-runs and badges silently stop — the trap documented in the extension README from DBSYNC-76.

### Evidence

`overlay_state.json`, regenerated by the real app after install:

```
sync_folder: /Users/mbsanchez/DropboxSync
Counter({'cloud_only': 23, 'synced': 6})
```

Finder, `~/DropboxSync/UNET`: nine `.cloudsc` entries carrying the grey download arrow, five real files carrying the green check, plainly distinct in the list-view badge column.

**The detail worth pointing at:** `referencias.pdf` shows the **green check**. Yesterday it was `referencias.pdf.cloudsc` — it was hydrated while validating DBSYNC-78, and the badge followed it across. The counts moved from 24 `cloud_only` + 5 `synced` to 23 + 6, which reconciles exactly with that one file. That is the state transition working end to end, not a static snapshot.

### Acceptance criteria

| AC | Result |
| --- | --- |
| Placeholders carry a badge at all | ✅ 23 of them |
| Placeholders and real files are visually distinct | ✅ Grey arrow vs green check |
| Real files unaffected | ✅ 6 still `synced` |
| Windows path unchanged | ⚠️ Not testable on this machine — see below |

**Windows was not exercised**, and cannot be here. Its arm rests on the `cfg(not(windows))` gate, a unit test asserting the skip still holds on every platform, and the review's cfg-pivot compile check. That is real evidence about the code, not about the platform. Anyone with a Windows box should confirm the status column is unchanged before this is treated as verified there.

Ready for PO approval.

Issue #93, opened 2026-08-20
DBSYNC-78 Slice 2: Validate the menu-bar icon live (manual, dev Mac)


mbsanchez:
Vertical slice of **DBSYNC-78**. Type: **manual / HITL** — requires a human looking at a real menu bar.

* Scope dir(s): none (validation only)
* Blocked by: Slice 1 (#92)
* User stories: 1, 2

## Why this is its own slice

Whether an icon *looks* native is a human judgement; no `cargo test` can assert it. Keeping it separate also stops the automated half from being reported as done when only the mechanical part was checked — which is exactly how badges were called working on 2026-08-19 when they were not visible on screen.

## Acceptance criteria

- [ ] The icon appears in the menu bar and is tinted by the system, matching the weight and colour of neighbouring items (Dropbox, OneDrive, odrive) rather than standing out.
- [ ] Verified on a **light** and a **dark** menu bar — the point of a template image is that the app does not choose.
- [ ] Verified with the menu **open**, i.e. in the highlighted state, which a non-template icon does not follow.
- [ ] All three states seen live and told apart **by shape alone**: Idle, Syncing, Error.
- [ ] Screenshot evidence attached to the PR for each of the above.

## Known difficulty, stated up front

Only **Idle** has ever been observed live. `Syncing` and `Error` need to be provoked, and the same gap is still open on DBSYNC-76 Slice 4, where the `out_of_sync` and `syncing` badges were never seen either. If they cannot be triggered in a reasonable way, say so and leave the box unticked — do not infer them from the Idle case, and do not close this slice on a partial check.

**Windows cannot be validated here.** The dev machine is a Mac. The Windows arm is asserted by a unit test and by the assets being untouched, which is not the same as having seen it. That limit belongs in the PR, not hidden behind a ticked box.

## Reminder

macOS hides status items when the frontmost app's menus need the room. Before reporting the icon as absent, bring a small-menu app forward and look again.

Jira: DBSYNC-78

mbsanchez:
## Implementation Plan

**Scope dir(s)**: none — validation only, no file changes
**Implementing agent(s)**: human (manual desktop QA, per the project's qa_strategy)
**Estimated complexity**: medium — low effort, but two ACs have no known way to trigger them
**Dependencies**: blocked by Slice 1 (#92)

### Steps

[...]

Issue #92, opened 2026-08-20
DBSYNC-78 Slice 1: Native macOS menu-bar icon — template artwork + platform-aware resolver


mbsanchez:
Vertical slice of **DBSYNC-78**. Type: **AFK**.

* Scope dir(s): `apps/desktop/src-tauri`
* Blocked by: none
* User stories: 1, 2, 3

## What

macOS gets monochrome template artwork; Windows keeps its colour icon. Both the initial icon (`TrayIconBuilder` in `lib.rs`) and the live state swap (`update_tray_tooltip` in `auth_session.rs`) resolve through one function so they cannot disagree.

This reverses the DBSYNC-32 ADR **for macOS only** — see the Jira ticket for the decision and its reasoning.

## Acceptance criteria

- [ ] `tray_icon_for_label(label) -> (&'static [u8], bool)` is the single source of both the asset and the `icon_as_template` flag; neither call site carries its own `include_bytes!` match or its own boolean literal any more.
- [ ] macOS resolves to monochrome template assets and `true`; every other platform resolves to the existing colour assets and `false`. The Windows arm is byte-identical to today — same three files, same flag.
- [ ] The three states stay distinguishable **without colour**: cloud+check, cloud+circular-arrow, cloud+exclamation. DBSYNC-32 decides *when* the icon changes and is untouched.
- [ ] No `#[tauri::command]` added, no capability declaration added — the tray helper is `pub(crate)` and never crosses the IPC boundary.
- [ ] `cargo test` green, including: every state resolves to different artwork; an unknown label falls back to idle; the template flag is uniform across states; the flag is `true` on macOS and `false` elsewhere (each compiled only on its own platform).
- [ ] `cargo clippy --all-targets` introduces no new warnings.
- [ ] Bundle-size impact recorded — the assets are `include_bytes!`-linked into the binary.

## Notes for the implementer

Template images are read by macOS **for their alpha only**; the system tints the shape. That is why colour cannot survive templating and why the states differ by shape. Render at medium symbol weight — hairlines disappear when a 128 px asset is scaled to the ~18 pt menu bar.

**Do not treat "the icon is missing from the menu bar" as a defect without checking overflow first.** macOS hides status items when the frontmost app's menus need the room; an app with nine menu titles pushed this icon off the bar entirely on a 1280-point-wide display, and it reappeared as soon as a smaller-menu app came forward. That cost real debugging time on 2026-08-20.

Jira: DBSYNC-78

[...]

PR #94, opened 2026-08-20
fix(#DBSYNC-78): monochrome template menu-bar icon on macOS, colour kept on Windows

mbsanchez:
## Summary

The macOS menu-bar item was not merely a colourful glyph — it was a **rounded-rectangle app-icon tile** (blue gradient, white cloud, blue check, 128×128). Dock artwork in the menu bar. It adapted to nothing the system does there and read as foreign beside Dropbox, OneDrive and odrive, which all ship templates.

macOS now ships monochrome template artwork. **Windows keeps its colour icon** — the notification area has no template convention, so the original reasoning still holds there. Two platforms whose conventions genuinely differ, not two ways of doing one thing.

## This reverses an ADR, deliberately

DBSYNC-32 decided *"Not a template — we keep the brand blue in every state"*, stated twice in prose beside the code it described. That traded native appearance for brand recognition. Reversed for macOS only, on the maintainer's explicit call (recorded on the Jira ticket).

**Resolving the conflict with DBSYNC-32:** templating **discards colour**, so a green/red state swap cannot survive it. The three states are distinguished by **shape** inside a common cloud silhouette — check, circular arrow, exclamation. DBSYNC-32 decides *when* the icon changes and is untouched; only *how* it is drawn changes.

Both stale comments are deleted along with the code they described. A comment asserting the opposite of what the code does is worse than either alone.

## Key changes

- `tray_icon_for_label(label) -> (&'static [u8], bool)` is the **single source** of both the asset and the `icon_as_template` flag. Each call site previously carried its own `include_bytes!` match and its own hardcoded `false` — the shape of duplication that lets a startup icon and a state icon drift apart with nothing to catch it.
- Three new assets from Apple's SF Symbols (`checkmark.icloud`, `arrow.clockwise.icloud`, `exclamationmark.icloud`), **medium** weight — hairlines vanish when 128 px is scaled to the ~18 pt bar.
- A tuple rather than a state enum: an enum would be tidier in the abstract and would ripple through DBSYNC-32's call sites for no gain today, since the label is already a status string everywhere it is produced.

## Stack

`desktop-tauri` (Rust only — no frontend, no IPC, no capabilities)

## Jira Ticket

#DBSYNC-78 — slices [#92](https://github.com/mbsanchez/DropboxSync/issues/92) (AFK) and [#93](https://github.com/mbsanchez/DropboxSync/issues/93) (manual).

## Test Plan

- [x] `cargo test --manifest-path apps/desktop/src-tauri/Cargo.toml` — **167 passed**, 4 new. Each pins a specific *silent* failure:
  - every state resolves to **different** artwork — a collision would stop macOS reporting state while the flag and both call sites still looked correct
  - an unknown label falls back to **idle** — the label is a status string, not an enum
  - the template flag is **uniform** across states — a mixed pair would visibly change weight mid-sync
  - the flag is `true` on macOS and `false` elsewhere, each assertion `cfg`-gated to its own platform
- [x] `cargo clippy --all-targets` — no new warnings; the four that remain are pre-existing and in untouched files.
- [x] **No IPC surface, no capabilities.** Measured: 0 `#[tauri::command]` added, `capabilities/default.json` untouched.
- [x] **Bundle size measured, not predicted.** Searching the built binary for a byte signature from each PNG:

  ```
  tray-idle.png            linked=False     tray-idle-template.png   linked=True
  tray-sync.png            linked=False     tray-sync-template.png   linked=True
  tray-error.png           linked=False     tray-error-template.png  linked=True
  ```

  The `#[cfg]` arms do strip the discarded `include_bytes!`. Net **−9,469 bytes** on macOS.

### Manual desktop validation (macOS menu bar)

- [x] **System-tinted.** The shipped asset is **black** on transparent and renders **white** on screen. That inversion is the proof — a white asset would have looked identical while proving nothing.
- [x] **Highlighted state.** Opening the menu makes macOS draw its own rounded highlight behind the icon, as it does for system items.
- [x] **`Syncing` captured live.** A capture loop over the menu-bar region at 0.4 s, started *before* triggering a real hydration so the opening frames are a known-Idle baseline. 60 frames collapsed by hash to exactly two images: 57 baseline, and 3 consecutive frames beginning right after the trigger — cloud + **circular arrow** against the baseline's cloud + **check**, plainly different at real size. The window lasted ~1.2 s, consistent with DBSYNC-48's note that a 60 s poll misses it.

### Two checks NOT met — stated, not skipped

- [ ] **Light vs dark menu bar.** Toggled both ways and captured both; the captures are **identical**, because this machine's menu bar is translucent over a light wallpaper and its own appearance did not change. **Inconclusive as evidence** — it neither confirms nor refutes. The black-asset/white-render inversion above already establishes tinting; this specific comparison needs a machine whose bar visibly changes.
- [ ] **`Error` state never seen.** Triggering it means deliberately breaking the token or a transfer; not done. The glyph is structurally the same cloud with a different inner symbol, so the risk is low — but low risk is not evidence, and the box stays open. The identical gap is open on DBSYNC-76 Slice 4.

**Note:** these boxes are unticked on purpose. On 2026-08-19 the Finder badges were reported working twice before anyone had looked, and both reports were wrong.

## Disclosure

Validating `Syncing` required a real transfer, authorised by the maintainer: `UNET/referencias.pdf` (921,678 bytes) is now materialised and its `.cloudsc` is gone. Reversible by dehydrating from the app.

🤖 Generated with Claude Code


mbsanchez (commented):
## Code review — APPROVED, with two non-blocking observations

Reviewed against the `desktop-tauri` and `rust` rows, plus the stack-agnostic checklist.

### The check this PR was missing, now run

**The non-macOS arm had never been compiled by anyone.** CI does not run (DBSYNC-74) and the dev machine is a Mac, so a syntax or type error in the Windows branch would have shipped undetected — and note the trap: the `cfg`-gated unit test asserting `false` lives *inside* the untested arm, so it proves nothing about a platform it never compiles on.

Cross-compilation was not available (`rustup target list --installed` returns only `aarch64-apple-darwin`), so the arm was exercised by **pivoting the `cfg` predicate**: `target_os = "macos"` → `"ios"` on both arms and both platform tests, making the non-macOS branch the live one on this host. Result:

```
test other_platforms_keep_their_colour_icons ... ok
test every_state_has_its_own_artwork ... ok
test unknown_label_falls_back_to_idle ... ok
test the_template_flag_is_the_same_for_every_state ... ok
test result: ok. 4 passed
```

The Windows branch compiles, type-checks, resolves three distinct colour assets and returns `false`. File restored afterwards; working tree clean and the macOS tests pass again. This proves the code *shape*, not Windows *behaviour* — the notification area still has to be seen by someone on Windows.

### Checklist

| | |
| --- | --- |
| KISS / YAGNI | ✅ A consolidation, not a new abstraction. The tuple over an enum is argued in the ticket and I agree: an enum would ripple through DBSYNC-32's call sites for no gain today. |
| Scope | ✅ Entirely inside `apps/desktop/src-tauri`. |
| Secrets | ✅ None. |
| Debug output | ✅ None left behind. |
| Security | ✅ No IPC surface (`pub(crate)`, never a `#[tauri::command]`), no capability change, no injection surface. Verified, not assumed: 0 commands added, `capabilities/default.json` untouched. |
| Rust idioms | ✅ clippy `--all-targets` clean, no `unsafe`, no `unwrap()` outside tests, no panics on the tray path. |
| Tests | ✅ Four, and each names the silent failure it prevents rather than chasing coverage. |

### Observation 1 — the comment under-describes the code (non-blocking)

The doc comment says *"**Windows keeps the colour.** The notification area has no template convention…"*, but the predicate is `not(target_os = "macos")`, which also covers **Linux**. Linux is a considered platform in this repo — `overlay_state.rs` documents the contract as feeding *"macOS Finder Sync badges, Linux file-manager emblems, and the Windows status column"*.

Behaviour is right; the prose names one of the two platforms it governs. Worth fixing precisely because this PR's own thesis is that a comment disagreeing with its code is worse than either alone — it deleted two such comments. Suggest: *"Every non-macOS platform keeps the colour icon — Windows' notification area has no template convention, and Linux tray implementations vary."*

### Observation 2 — a pre-existing silent failure, not introduced here

`update_tray_tooltip` swallows a decode failure: `if let Ok(img) = Image::from_bytes(bytes)` with no `else`. If an asset ever failed to decode, the tray would quietly keep its previous icon with nothing logged. That behaviour predates this PR and the diff does not make it worse, so it is **not** a blocker. Flagging it because this codebase has now lost real time to exactly this failure mode three times in two days — the Finder Sync state file, the extension registration, and the hidden status item. `lib.rs` propagates the same error properly with `?`.

### Verdict

**Approved.** The evidence discipline in the PR body is the right standard: two acceptance criteria left explicitly unticked with the reason stated, and the bundle-size claim measured against the built binary rather than inferred from the source.

Next: `/cowork-workflow:jira-ticket-qa`

mbsanchez:

Verdict from the maintainer, looking at the real menu bar: **the icon reads as native and the menu behaves correctly.**

### Acceptance criteria

| AC | Result | Evidence |
| --- | --- | --- |
| Appears, system-tinted, matches neighbours | ✅ | Confirmed by the maintainer; capture beside Dropbox / OneDrive / odrive |
| Highlighted state (menu open) | ✅ | macOS draws its own rounded highlight |
| `Syncing` seen live, distinguishable by shape | ✅ | Cloud + circular arrow vs cloud + check, 3 frames during a real hydration |
| Light vs dark menu bar | ⛔ **Not testable on this display** | See below |
| `Error` seen live | ❌ **Not triggered** | Maintainer chose not to break the token for it |

### Light vs dark — upgraded from "inconclusive" to "not testable here", with proof

The whole menu bar was captured in both modes and hashed:

```
light hash: 0f08dba26a3ac02153338344c502b9c7
dark  hash: 0f08dba26a3ac02153338344c502b9c7
```

**Byte-identical.** Nothing in the bar changes — Dropbox, OneDrive and odrive included, all known-good template images. The bar's appearance on this display is driven by the wallpaper behind its translucency, not by the Light/Dark setting, so the test has no signal here for any icon. Our icon is indistinguishable from the control group under the same conditions.

That is not evidence of correctness, and it is not being counted as such. Tinting is established separately and more strongly: the shipped asset is **black** and it renders **white**.

### `Error` — left open on purpose

Triggering it means deliberately breaking a token or a transfer. The maintainer declined, which is the right call for a cosmetic AC. The glyph is the same cloud silhouette with a different inner symbol, as with the two states already seen, so the residual risk is low — but low risk is not evidence and the box stays unticked on #93. The identical gap is still open on DBSYNC-76 Slice 4.

### Defect found during QA — filed separately, NOT a regression from this PR

The maintainer reported that **the flyout takes about three clicks to appear**, and that the intended behaviour is left click → window, right click → menu.

Root cause located and **not caused by this PR** — the diff here touches only `.icon(...)` and `.icon_as_template(...)`, never the click path:

`lib.rs` attaches `.menu(&menu)` and never configures `show_menu_on_left_click`. In `tray-icon 0.21.3` the default is `menu_on_left_click: true` (`src/lib.rs:209`), so a left click opens the **menu** *and* fires `on_tray_icon_event`, which toggles the flyout at the same time. Two behaviours contending for one click.

The comment above that handler claims to be the *"single source of truth for the flyout"*; the builder default contradicts it. Filed as **DBSYNC-82** with the fix direction.

Ready for PO approval.

Issue #97, opened 2026-08-20
Tray: the 250 ms recently_toggled suppression never altered an outcome — candidate for removal


mbsanchez:
Split out of **DBSYNC-82** Slice 2 (#96), which measured this but was forbidden from acting on it. Filed so the evidence is not lost.

## What was measured

`run_events.rs` hides the `main` flyout when it loses focus, gated by a 250 ms suppression:

```rust
if visible && !crate::recently_toggled(Duration::from_millis(250)) {
    let _ = w.hide();
}
```

`recently_toggled` (`lib.rs`) exists so a tray click that both focuses an already-visible flyout and re-triggers the toggle does not race a blur-driven hide against its own show. Its doc comment describes that race explicitly.

After DBSYNC-82 removed the menu from the left click, both branches of the gate were instrumented and exercised: 10 alternating show/hide rounds, a right-click round, and a genuine click-away.

| Scenario | `visible` | `suppressed` | Outcome |
| --- | --- | --- | --- |
| 12 events during tray-click sequences | `false` | `true` | Suppression changed nothing — the hide was already a no-op |
| 1 event on a genuine click-away | `true` | `false` | Hide ran correctly and the flyout closed |

**`visible && suppressed` was true 0 times out of 13.**

## What that does and does not establish

**Does:** in every observed tray click, the window was already hidden by the time the blur arrived, so the guard was protecting a call that would not have run. The race it was written for did not occur.

**Does not:** prove it is dead. 13 samples across two interaction patterns is a small window, all on one machine, one macOS version, one display. Timing workarounds fail on the machine you did not test — a slower or more loaded system could reorder the very events that lined up cleanly here.

**Explicitly not in question:** the blur-hide itself. The click-away case shows it doing exactly its job. This ticket is about the 250 ms guard around it, nothing else.

## Why this was not done inside DBSYNC-82

Deleting a timing workaround on the same day its suspected cause was removed, on 13 samples, is the same move that created it: fixing what you can see without measuring what you cannot. DBSYNC-82's plan forbade it in writing, and that held.

## Direction to decide

1. **Widen the sample** before deciding — different load, a second machine if one is available, and the interaction paths not yet exercised (dragging onto the flyout, Mission Control, a display change).
2. Or **argue it from the code** rather than from samples: enumerate what can now emit `Focused(false)` on `main` while the window is visible and a tray toggle happened within 250 ms. If that set is provably empty, the guard is dead and the sample size stops mattering.
[...]

Issue #96, opened 2026-08-20
DBSYNC-82 Slice 2: Validate the click behaviour live, and measure the 250 ms workaround


mbsanchez:
Vertical slice of **DBSYNC-82**. Type: **manual / HITL** — a human has to click a real menu-bar icon.

* Scope dir(s): none (validation + measurement only)
* Blocked by: Slice 1 (#95)
* User stories: 1, 2

## Why this is separate

The bug is a race between three mechanisms. Whether it is gone is a question about *timing and focus on a live desktop*, which no unit test reaches. Keeping it out of Slice 1 also stops "the flag is set" being reported as "the bug is fixed" — the two are not the same claim.

## Acceptance criteria — behaviour

- [ ] **One** left click shows the flyout. Not two, not three.
- [ ] A second left click hides it.
- [ ] Right click shows the menu, and **both** items still work (Settings opens the setup window; Exit quits).
- [ ] A right click does **not** toggle the flyout as a side effect.
- [ ] Repeated alternation (left, left, right, left…) stays deterministic — the original symptom was non-determinism, so a single successful click proves little. Exercise it ten times or so.

## Acceptance criteria — measurement

- [ ] **Determine whether `recently_toggled(250 ms)` still fires after the fix**, and record the answer with evidence.

  It exists to suppress the blur-hide when a tray click races it (`run_events.rs:58`, `lib.rs:53`). If the menu stealing focus on left click was the only thing creating that race, the suppression may now be dead code. If other paths still trigger it, it is load-bearing and must stay.

  Suggested method: a temporary `tracing` line where the suppression takes effect, exercised across the click sequence above. Instrumentation must be **removed before commit** — the DBSYNC-76 diagnostics were, and the same standard applies.

- [ ] **Do not remove the suppression in this ticket** whatever the measurement says. If it turns out to be dead, open a follow-up with this evidence attached. Deleting a timing workaround on the same day its cause was removed, without a measurement in hand, is how the original 3-click bug got its band-aid in the first place.

## Known limits, stated up front

* **Windows and Linux are not covered.** `menu_on_left_click` has its own platform implementation on each, and the source reading that settles the right-click question above is the **macOS** file specifically. The flag is cross-platform in the API, but its behaviour elsewhere is unverified from this machine.
* Evidence for every ticked box goes on the PR. On 2026-08-19 the Finder badges were reported working twice before anyone had looked; both reports were wrong.

Jira: DBSYNC-82

[...]

Issue #95, opened 2026-08-20
DBSYNC-82 Slice 1: Stop the left click from opening the menu


mbsanchez:
Vertical slice of **DBSYNC-82**. Type: **AFK**.

* Scope dir(s): `apps/desktop/src-tauri`
* Blocked by: none
* User stories: 1, 2

## What

`TrayIconBuilder` never configures what a left click should do with the attached menu, so `tray-icon`'s default takes over and opens it — while `on_tray_icon_event` toggles the flyout for the same click.

One builder call fixes it: `show_menu_on_left_click(false)`.

## The right-click path needs no code — verified in the dependency source

`tray-icon-0.21.3/src/platform_impl/macos/mod.rs:477`:

```rust
if button == MouseButton::Right || (menu_on_left_click && button == MouseButton::Left) {
```

The menu is shown on **right click unconditionally**; only the left click is gated by the flag. So disabling it leaves the right-click menu fully intact, and no handler change is required.

The right click also emits a `Click` event with `MouseButton::Right` (`:361`). The existing handler matches `MouseButton::Left` only, so it correctly ignores it — a right click will not toggle the flyout as a side effect.

## Acceptance criteria

- [ ] `show_menu_on_left_click(false)` on the `TrayIconBuilder` (the current API; `menu_on_left_click` is deprecated in tauri 2.10.3 and calls the same underlying setter).
- [ ] The `on_tray_icon_event` handler is **unchanged** — it already matches `MouseButton::Left` only, and the right-click event it now also receives is correctly ignored.
- [ ] The comment above the handler is corrected. It currently claims to be the *"single source of truth for the flyout"*, which was false precisely because the builder default was competing with it. Leaving that sentence in place beside the fix reproduces the failure that caused this bug: prose asserting something the configuration contradicts.
- [ ] **The 250 ms `recently_toggled` suppression stays.** Removing it here would be speculative — see Slice 2, which measures whether it still fires.
- [ ] No `#[tauri::command]` added, no capability change, no new dependency.
- [ ] `cargo test` and `cargo clippy --all-targets` green, with no new warnings.

## Note on testability

`cargo test` cannot press a mouse button, and the builder is constructed inside `run()`'s `setup` closure with no seam to assert against. Do **not** invent an abstraction just to make the flag assertable — that would be a worse trade than the manual check in Slice 2. If a cheap assertion presents itself, take it; otherwise say so plainly rather than adding a test that only restates the source line.

Jira: DBSYNC-82

[...]

PR #98, opened 2026-08-20
fix(#DBSYNC-82): give the tray's left click to the flyout, the right click to the menu

mbsanchez:
## Summary

Clicking the menu-bar icon took **about three attempts** to show the window, and which attempt worked was not predictable. One builder call fixes it.

## Three behaviours on one gesture, not two

The ticket originally said two. The architecture review found a third, which is what makes the symptom non-deterministic:

1. **The menu opened on left click.** `tray-icon 0.21.3` defaults `menu_on_left_click` to `true` (`src/lib.rs:209`) and the builder never overrode it.
2. **The flyout toggled on the same click** via `on_tray_icon_event`.
3. **The blur-hide then fired**, because the menu had taken key focus from the flyout that had just been shown (`run_events.rs`).

The third one is the tell: that blur-hide **already carries a 250 ms `recently_toggled` suppression**, and its own doc comment describes this exact race — *"a click that both focuses the (already-visible) window and re-triggers the tray toggle doesn't race a blur-driven hide immediately followed by a click-driven show"*. Someone met this symptom before and put a timer over it instead of removing its cause.

## The fix

`show_menu_on_left_click(false)`. **The menu is not lost, and this is settled from the dependency source rather than assumed** — `tray-icon-0.21.3/src/platform_impl/macos/mod.rs:477`:

```rust
if button == MouseButton::Right || (menu_on_left_click && button == MouseButton::Left) {
```

Right click shows the menu **unconditionally**; only the left click is gated by the flag. The right click also reaches our handler carrying `MouseButton::Right`, which the existing match already ignores — so no handler change was needed, and none was made.

Both comments are rewritten. The old one called itself the *"single source of truth for the flyout"* while the builder default was quietly competing with it. The new text **names the precondition** that makes the claim true rather than restating the claim — a comment that only became correct by accident is worth replacing, not preserving.

## Stack

`desktop-tauri` (Rust only — no IPC, no capabilities, no new dependency)

## Jira Ticket

#DBSYNC-82 — slices [#95](https://github.com/mbsanchez/DropboxSync/issues/95) (AFK) and [#96](https://github.com/mbsanchez/DropboxSync/issues/96) (manual).

## Test Plan

- [x] `cargo test` — **174 passed**. Their role here is regression only.
- [x] `cargo clippy --all-targets` — no new warnings, and **zero deprecation warnings**, which is the one automated signal that actually distinguishes the correct API from the deprecated `menu_on_left_click`.
- [x] **No new unit test, and none was invented.** `cargo test` cannot press a mouse button, and the builder lives inside `run()`'s `setup` closure with no seam. Creating one purely to assert a boolean would add indirection to production code to buy a test that restates a source line. Stated rather than papered over.

### Manual desktop validation (macOS menu bar)

Clicks synthesised with `CGEvent` — accessibility's `click` only performs a menu-bar item's *default* action, so the right-click half cannot be exercised through AppleScript. Visibility read via System Events' window count.

- [x] **One** left click shows the flyout: 0 windows → 1.
- [x] A second hides it: → 0.
- [x] **10 alternating rounds: 10 passed, 0 failed** — `show=1 hide=0` every time. This is the criterion that matters; the bug was non-deterministic, so a single success would have proved nothing.
- [x] Right click shows the menu — **Settings** and **Exit** both render.
- [x] Right click does **not** toggle the flyout — window count stayed at 0 across the gesture.

## The 250 ms suppression: measured, and deliberately left alone

Both branches of the gate were instrumented and exercised:

| Scenario | `visible` | `suppressed` | Outcome |
| --- | --- | --- | --- |
| 12 events during tray-click sequences | `false` | `true` | Suppression changed nothing — the hide was already a no-op |
| 1 event on a genuine click-away | `true` | `false` | Hide ran correctly and the flyout closed |

**`visible && suppressed` was true 0 times out of 13.**

Two separate readings, and conflating them would be wrong:

- **The blur-hide is load-bearing.** The click-away case shows it doing its job. Nothing here argues against it.
- **The guard around it never altered an outcome** in these samples.

**It was not removed.** 13 samples across two interaction patterns is evidence, not proof, and timing workarounds fail on the machine you did not test. Deleting one on the same day its suspected cause disappeared is precisely the move that created it. Filed as [#97](https://github.com/mbsanchez/DropboxSync/issues/97) with the data and a better route than more sampling: prove from the code that the set of events which can still trigger the race is empty.

### A discarded measurement, recorded

The first click-away run was **invalid and its data thrown out**. It clicked at (640, 500) intending to land outside the flyout — but the flyout occupies **(331, 30) 360×600**, so that point is *inside the window*. The results were incoherent, and they looked enough like a finding that treating them as one would have produced a confident wrong answer. The window geometry was queried and the test redone at (100, 700).

- [x] Instrumentation removed before commit — `grep DBSYNC82 run_events.rs` → 0; the diff is one file, +13/−2.
- [x] Clean build reinstalled and re-checked: opens on the first click, zero instrumentation lines in the log.

## Not covered

**Windows and Linux.** `menu_on_left_click` has a separate platform implementation on each, and the source reading above is the **macOS** file specifically. Unverified from this machine.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/lib.rs
+++ b/apps/desktop/src-tauri/src/lib.rs
@@ -264,10 +264,21 @@ pub fn run() {
                 .icon(tray_image)
                 .icon_as_template(idle_as_template)
                 .menu(&menu)
+                // DBSYNC-82. `tray-icon` defaults `menu_on_left_click` to TRUE, so without
+                // this the left click opened the menu as well as running the handler below,
+                // and the menu stealing focus made the flyout blur-hide itself. Three
+                // behaviours on one gesture; it took about three clicks to land the window.
+                // Off, the menu belongs to the right click — which `tray-icon` shows
+                // unconditionally, so nothing is lost and the handler needs no change.
+                .show_menu_on_left_click(false)
                 .tooltip("DropboxSyncDesktop - Idle")
                 .on_tray_icon_event(move |_tray, event| {
-                    // Single source of truth for the flyout: a left-click toggles it
-                    // (shows+positions if hidden, hides if visible). No more DoubleClick.
+                    // The flyout's only trigger — true because `show_menu_on_left_click`
+                    // above takes the menu off this gesture, not because nothing else
+                    // could claim it. A left-click toggles the window (shows+positions if
+                    // hidden, hides if visible). No DoubleClick. The right click also
+                    // arrives here carrying `MouseButton::Right` and is ignored by the
+                    // match below, so it opens the menu without touching the flyout.
                     if let TrayIconEvent::Click {
                         button: MouseButton::Left,
                         button_state: MouseButtonState::Up,


mbsanchez (commented):
## Code review — APPROVED, with one framing correction

Reviewed against the `desktop-tauri` and `rust` rows plus the stack-agnostic checklist. A 13-line diff, most of it comment — so the review effort went into the claims the PR makes rather than the lines it changes.

### Two claims re-verified, because both were reachable by a wrong route

**1. The crate version.** The PR's central argument rests on reading `tray-icon-0.21.3/src/platform_impl/macos/mod.rs:477`. That source was located by listing the cargo registry and taking the highest version present — which is *not* evidence it is the version this project links. Checked against `Cargo.lock`:

```
name = "tray-icon"
version = "0.21.3"
```

It is. The reading stands, but it stood on a coincidence until now.

**2. Builder chain order.** The new call is placed next to `.menu(&menu)`, and a reader could reasonably infer the order is load-bearing. It is not — `menu()` calls `with_menu` and `show_menu_on_left_click` calls `with_menu_on_left_click`, two independent fields on the same attrs struct (`tauri-2.10.3/src/tray/mod.rs:241` and `:319`); neither resets the other. The placement is a readability choice. Worth knowing before someone "tidies" the chain and worries they have broken something.

### The one thing the PR under-states

**This change is not platform-gated, and the PR's framing implies otherwise.**

`show_menu_on_left_click(false)` is a plain builder call with no `cfg`. It applies on **Windows and Linux too**. The PR files that under a "Not covered" heading alongside the macOS-only source reading, which reads as *"unverified elsewhere"* — but the more important fact is that behaviour on those platforms genuinely **changes**.

The contrast with DBSYNC-78 is the point: that change was `cfg`-gated, so Windows came out byte-identical by construction and "unverified" meant "untouched". Here Windows users' left click stops opening the menu. That is almost certainly an improvement — the notification-area convention is left-click for the app, right-click for the context menu, the same as macOS — and nothing in the repo documents or depends on the old behaviour (checked: no reference to left-click semantics in any markdown or frontend source). But "very likely correct" and "unchanged" are different claims, and only the second was true of the previous ticket.

**Non-blocking.** The change should ship; the PR body should say *behaviour changes on Windows and Linux and was not validated there*, rather than listing them as merely uncovered.

### Checklist

| | |
| --- | --- |
| KISS / YAGNI | ✅ One builder call. The plan explicitly forbade inventing a seam to make the flag unit-testable, and that held — no abstraction was added. |
| Scope | ✅ `lib.rs` only, +13/−2. |
| Secrets | ✅ None. |
| Debug output | ✅ Clean. The measurement instrumentation was removed before commit and I re-checked the merge target: `grep DBSYNC82` across `src/` returns nothing. |
| Security | ✅ No IPC command, no capability change, no new dependency, no input surface. |
| Rust idioms | ✅ clippy `--all-targets` clean, **zero deprecation warnings** — which is the real signal that `show_menu_on_left_click` was used rather than the deprecated `menu_on_left_click`. |
| Tests | ✅ 174 pass as regression. No new unit test, correctly, and the PR says so plainly instead of manufacturing one. |

### On the comment rewrite

The old comment asserted it was the *"single source of truth for the flyout"* while the builder default silently competed with it. The replacement names the precondition that makes the claim true rather than repeating the claim. That is the right correction: the failure here was never the code, it was prose that could not be falsified by reading the file it sat in.

### On the discarded measurement

The PR volunteers that its first click-away run was invalid — the click at (640, 500) landed *inside* a flyout occupying (331, 30) 360×600 — and that its incoherent output looked enough like a finding to be dangerous. Publishing a thrown-out experiment alongside the one that replaced it is the correct standard, and it is what makes the 0-of-13 number trustworthy rather than merely stated.

Leaving the 250 ms suppression alone and filing #97 with the data is the right call for the reason the ticket gives: that guard exists because someone once fixed a symptom without measuring.

### Verdict

**Approved.** Next: `/cowork-workflow:jira-ticket-qa`

mbsanchez:

Verdict from the maintainer on the real menu-bar icon. The installed build already carried the fix, so no reinstall was needed to validate.

### Acceptance criteria

| AC | Result | Evidence |
| --- | --- | --- |
| **One** left click shows the flyout | ✅ | 0 windows → 1 |
| A second left click hides it | ✅ | → 0 |
| **Deterministic over 10 alternating rounds** | ✅ | `show=1 hide=0` ten times, **0 failures** |
| Right click shows the menu | ✅ | Settings + Exit rendered |
| Right click does **not** toggle the flyout | ✅ | Window count stayed at 0 across the gesture |
| Menu item **Settings** works | ✅ | Opened `DropboxSync — Setup` |
| Menu item **Exit** works | ✅ | pid 41748 before → no process after |

The 10-round result is the one that carries weight. The original bug was **non-deterministic**, so a single successful click would have proved nothing.

### A false result caught and corrected

The first Exit test reported *"process terminated — Exit works"* and **that conclusion was wrong**. Its `pid before` was empty: the app was already not running when the click was attempted, and the AppleScript errored saying the process did not exist. The script only inspected state *after* the action, so "nothing is running" satisfied it regardless of cause.

Redone with a proper before/after guard — app relaunched, confirmed running as pid 41748, Exit clicked, process confirmed gone. That is the result in the table.

Worth recording because the failure mode is subtle: a check that only looks at the end state will happily confirm any outcome that happens to match it.

### An unexplained app exit, investigated and not turned into a ticket

Between the Settings test and the first Exit attempt the app exited on its own. Investigated:

- No crash report in `~/Library/Logs/DiagnosticReports/`.
- The log shows normal activity either side of a gap, with no error at the exit.
- **Hypothesis tested and refuted:** that closing the `Setup` window quits the app — plausible, since `run_events.rs` lets that window close genuinely rather than hiding it. Reproduced deliberately: opened Settings, closed the window, watched the pid at +6 s and +14 s. **The app survived.**

The likeliest explanation is mundane: the maintainer was asked to exercise the tray menu during this very QA, and that menu has an Exit item. No reproduction, no crash, no log evidence — so no ticket is being filed on a phantom.

### Not covered

**Windows and Linux.** As raised in review, this change is **not** `cfg`-gated — unlike DBSYNC-78, behaviour there genuinely changes rather than merely being unverified. The convention on both matches (left click for the app, right click for the menu), so it is very likely an improvement, but nobody has seen it.

Ready for PO approval.

Issue #101, opened 2026-08-20
DBSYNC-81 Slice 3: Validate the prompt count from a clean state (manual)


mbsanchez:
Vertical slice of **DBSYNC-81**. Type: **manual / HITL**.

* Scope dir(s): none — validation only
* Blocked by: Slice 2 (#100)
* User stories: 1, 2, 3

## Acceptance criteria

- [ ] **First run after a clean install prompts exactly once.** Not "fewer times" — once. The count is the deliverable.
- [ ] **A second launch of the same binary prompts zero times.** This already holds today and must not regress: measured before the fix as `SecurityAgent` absent across 25 one-second samples of a quit-and-relaunch cycle.
- [ ] **A full `cargo test` run adds no new application entry** to the item's ACL. Slice 1's guarantee, re-checked here end to end.
- [ ] The user is still signed in after updating from the old five-item layout — **no re-authentication**.
- [ ] Evidence attached: the ACL dump before and after, and the observed prompt count.

## How to establish a clean state

The current keychain is polluted: `dropbox-session-expires` carries **seven** trusted applications, six of them dead `cargo test` binaries with invalidated signatures. Measuring "how many prompts does a new user get" against that store measures nothing.

Getting to a clean state means deleting the app's keychain items and signing in again. **That is destructive to the maintainer's session and must be confirmed by them before it happens** — it is not a step to take unilaterally, and the prompt count is not worth a surprise logout.

If they would rather not, say so and leave the AC unticked. An unmeasured box is honest; a box ticked against a polluted store is not.

## Detecting prompts without asking a human to count

`SecurityAgent` is the process macOS spawns to draw the authorization dialog. Polling `pgrep -x SecurityAgent` once a second across a launch gives a mechanical count that does not depend on someone watching the screen — this is how the zero-prompt relaunch result above was obtained.

Note it detects *a* dialog, not *which item* it names. If the count needs attributing per item, the ACL dump before and after is the better instrument.

## Reminder

The failure mode to guard against on this ticket is the same one that has recurred all session: reporting a fix as working before looking. On 2026-08-19 the Finder badges were called working twice before anyone had looked, and both reports were wrong.

Jira: DBSYNC-81

mbsanchez:
## Implementation Plan

**Scope dir(s)**: none — validation only
**Implementing agent(s)**: human (manual desktop QA), with the prompt counting automated
**Estimated complexity**: **medium** — the counting is mechanical, the clean state is not
[...]

Issue #100, opened 2026-08-20
DBSYNC-81 Slice 2: One keychain item on macOS instead of five


mbsanchez:
Vertical slice of **DBSYNC-81**. Type: **AFK**.

* Scope dir(s): `apps/desktop/src-tauri`
* Blocked by: Slice 1 (#99) — while `cargo test` still touches the real keychain, any before/after prompt count here is contaminated
* User stories: 1, 2

## The measurement this slice acts on

The store holds **five** items, not the three the ticket originally claimed. Enumerated with `security dump-keychain` (attributes only, no prompt):

```
dropbox-session-access      ← chunk marker
dropbox-session-access.0
dropbox-session-access.1
dropbox-session-refresh
dropbox-session-expires
```

`get_session()` reads all five: `load_value_chunked(SESSION_ACCESS)` walks the marker plus both chunks, then `SESSION_REFRESH`, then `SESSION_EXPIRES`. macOS prompts **per item**, and *Always Allow* grants the one item in the dialog. **Five items, five prompts** — which matches the maintainer's report exactly, not approximately.

## Why the access token is split at all

`SAFE_UTF16_PAYLOAD_BYTES = 2400`, because Windows' `CRED_MAX_CREDENTIAL_BLOB_SIZE` is 2560 bytes and `keyring` stores passwords as UTF-16 (`secure_store.rs:15`). **macOS has no such limit.** A Windows constraint is being paid on macOS, and it triples the prompt count for the access token alone.

## Acceptance criteria

- [ ] On macOS the session is stored so that a single `get_session()` touches **one** keychain item, not five.
- [ ] **Windows is untouched.** Its chunking and its 2560-byte ceiling remain exactly as they are. A macOS-only layout is acceptable; silently breaking Windows is not, and it cannot be verified from this machine — so the Windows arm must be provably unchanged by reading the diff, not by testing it.
- [ ] **Existing users migrate without re-authenticating.** Anyone already signed in has the five-item layout on disk. Reading it, rewriting it in the new shape and removing the old entries must happen transparently — a user who updates and is asked to log in again is a failure of this slice, not an acceptable cost.
- [ ] The migration is exercised by a test, not only by reasoning about it.
- [ ] No IPC command added, no capability change. `SecureStore` is not exposed over IPC today (verified: zero `#[tauri::command]` reaches it) and must stay that way.
- [ ] `cargo test` and `cargo clippy --all-targets` green.

## Deliberately not specified

[...]

Issue #99, opened 2026-08-20
DBSYNC-81 Slice 1: Stop the test suite from reading the real keychain


mbsanchez:
Vertical slice of **DBSYNC-81**. Type: **AFK**.

* Scope dir(s): `apps/desktop/src-tauri`
* Blocked by: none
* User stories: 3

## The evidence

Dumping the keychain ACLs (`security dump-keychain -a`, attributes only — no prompt) for `dropbox-session-expires` shows **seven trusted applications**, and six of them are the same file:

```
entry 1, applications (7):
  0..5: .../target/debug/deps/dropbox_sync_desktop_lib-fb105e315db65b93
        six entries, each a DIFFERENT cdhash
        five of them status -2147415734 (signature no longer matches)
  6:    .../target/release/bundle/macos/DropboxSyncDesktop.app  (OK)
        requirement: identifier "com.mobsanchez.dropboxsyncdesktop" and anchor apple generic
                     and certificate leaf[subject.OU] = XCAA3WMJM6
```

That first path is the **`cargo test` lib harness**. It is in the ACL because it *asked for the secret and was granted it*. Every `cargo test` run compiles a fresh binary with a fresh cdhash, which macOS treats as a different application — so it prompts again and leaves another dead entry behind.

This is the sibling of **DBSYNC-75** (`cargo test` writing the live `overlay_state.json`), and worse in kind: that one clobbers state, this one reads production credentials.

## Why this slice goes first

It is the actual source of the maintainer's day-to-day pain. Relaunching the installed app produces **zero** prompts — measured: `SecurityAgent` seen in 0 of 25 one-second samples across a quit-and-relaunch cycle. The app's ACL entry is keyed on the designated requirement, which survives rebuilds signed with the same identity.

So the recurring dialogs come overwhelmingly from test runs, not from using the app. Until that stops, any measurement taken for Slices 2 and 3 is contaminated by it.

## Acceptance criteria

- [ ] **Name the test.** Which test, through which call path, reaches `SecureStore`? `refresh_token_guarded`'s own test is already isolated — it takes an injected refresher and never touches the store, verified by reading its signature — so it is **not** the culprit. Five `AppState` builders construct a `SecureStore`, but constructing one is a no-op; something must actually call `get_session`, `get_token` or `store_session`. Find it, do not guess.
- [ ] `cargo test` completes without any keychain access. Verified by evidence, not by assertion: no new application entry appears in the item's ACL after a full run.
[...]

PR #102, opened 2026-08-20
fix(#DBSYNC-81): keep cargo test out of the real keychain, and stop chunking off Windows

mbsanchez:
## Summary

Two defects behind "the keychain asks me about five times", both **measured**. The reported symptom turned out to have a cause nobody had suspected — and the ticket's original hypothesis was refuted along the way.

## 1. One test was reading production credentials

`sync_pipeline::tests::scan_blocks_mass_deletion_until_overridden` — one test out of 174 — called the full scan and discarded the Result, on an assumption it stated in a comment:

> *"The scan's later remote-refresh step **errors here with no token**…"*

That holds on CI, where the keychain is empty. On a developer's machine it is **false**: there *is* a token, so the step does not error — it prompts and reads the real session. Every `cargo test` compiles a fresh binary with a fresh cdhash, which macOS treats as a different application, so it prompts again and leaves a dead ACL entry behind. **Six had accumulated.**

**The fix**: `scan_local_changes_internal` was doing two jobs. Split so both are named — `scan_local_changes_only` (walk, decide, enqueue; no credentials) and the original composing it with the remote refresh and bookkeeping. The test calls the local half and now **asserts the Result** instead of swallowing it; it was only ignoring that error because of the wrong assumption.

### How it was found — the part worth stealing

The planned measurement (poll `SecurityAgent`, diff the ACL) gave a **contradiction**: the agent was up for 34 samples, yet the ACL looked unchanged. That was **my measurement error** — I counted one of the five stored items. Recounting all five showed `dropbox-session-access` going 7 → 8 while the rest held at 7.

It still did not name the test. A `panic!` at the top of `get_session` under `#[cfg(test)]` did: it aborts **before** the keychain call, so the harness reports the offending test through its thread name and **costs zero dialogs**. For anything like this it beats watching for prompts.

## 2. Chunking off Windows bought nothing and cost prompts

`SAFE_UTF16_PAYLOAD_BYTES = 2400` exists for `CRED_MAX_CREDENTIAL_BLOB_SIZE` — a **Windows** limit. macOS has none, and macOS prompts per keychain **item**, so splitting the access token into a marker plus two chunks turned one authorization into three. Measured: five stored items, five dialogs on first run.

The constant is now platform-conditional. `MAX / 2` rather than `MAX` so the accumulator in `split_utf16_chunks` cannot overflow.

**No migration code was needed and none was written.** `store_value_chunked` already inlines a single-chunk value into the base key and calls `clear_overflow_parts(base, 0)`, deleting `.0`/`.1`. The next session write collapses the layout and cleans up after itself — nobody is asked to sign in again.

## Stack

`desktop-tauri` (Rust only — no IPC, no capabilities, no new dependency)

## Jira Ticket

#DBSYNC-81 — slices [#99](https://github.com/mbsanchez/DropboxSync/issues/99), [#100](https://github.com/mbsanchez/DropboxSync/issues/100); [#101](https://github.com/mbsanchez/DropboxSync/issues/101) still open.

## Test Plan

- [x] `cargo test` — **177 passed** (174 + 3 new).
- [x] `cargo clippy --all-targets` — no new warnings; the four that remain are pre-existing and in untouched files.
  **This line was false when first written** and CTO review caught it: the run it quoted predated the chunking tests, and the final diff emitted `unused import: SAFE_UTF16_PAYLOAD_BYTES`. Fixed in 609325c and re-run against the final diff with a forced rebuild (`touch src/lib.rs`) so nothing came from cache. The four survivors are in `cloudsc_ops.rs`, `dropbox_transfer.rs` and `open_handlers.rs` — the branch touches neither file.
- [x] **No test reaches the keychain.** Proven by re-installing the probe on **both** `get_session` *and* `get_token` and running the full suite: 174 pass, **zero panics**. Before the fix exactly one panicked. Stronger than an ACL diff — it shows no path *reaches* the call, not merely that no entry appeared.
- [x] **The Windows arm was compiled and executed**, by pivoting its `cfg` predicate so the Windows branch goes live on this host. `an_oversized_secret_is_still_split_on_windows ... ok` — the 2400 constant still splits and every chunk respects the ceiling. Pivot reverted; `grep cfg(unix)` → 0. This matters because that test lives *inside* the arm that never compiles here, so writing it proves nothing on its own.
- [x] Probe removed before commit: `grep DBSYNC81` → 0.

### On the new tests

`split_utf16_chunks` is pure, so chunking is testable without a keychain. Four tests, and one of them is a guard: `the_test_secret_really_is_over_the_windows_ceiling` exists because without it the two platform tests could pass **vacuously** on a secret that never needed splitting.

## What this does NOT do

- **It reaches three prompts on first run, not one.** Collapsing `access`/`refresh`/`expires` into a single item is a genuine schema migration — runs once per user, no second chance — and buys the smaller half of the win. Left for its own change.
- **The item-count drop is predicted from the code, not observed.** Confirming it means writing a session and re-reading the store, which is #101's job.
- ~~**`security dump-keychain` now hangs** on this machine past 20 s~~ — it responded normally on the post-review run, so the hang was transient rather than a standing limitation. The probe evidence stands either way.

## And a hypothesis this refutes

The ticket suspected repeated `codesign --force` re-signing of causing the recurrence. **Refuted**: the app's ACL entry is keyed on the designated requirement (identifier + Apple anchor + team `XCAA3WMJM6`), not a cdhash, so it survives rebuilds with the same identity. Confirmed by quitting and relaunching the installed app: `SecurityAgent` absent across 25 one-second samples — **zero prompts**.

The recurring dialogs were test runs, not app use.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/storage/secure_store.rs
+++ b/apps/desktop/src-tauri/src/storage/secure_store.rs
@@ -13,7 +13,22 @@ const SESSION_EXPIRES: &str = "dropbox-session-expires";
 
 /// Windows `CRED_MAX_CREDENTIAL_BLOB_SIZE` is 2560 bytes; keyring stores passwords as UTF-16 LE,
 /// so the check is `utf16_units * 2 <= 2560`. We stay under that per chunk.
+#[cfg(windows)]
 const SAFE_UTF16_PAYLOAD_BYTES: usize = 2400;
+
+/// **No equivalent ceiling exists off Windows** (DBSYNC-81). Chunking there bought
+/// nothing and cost a keychain prompt per part: macOS asks per ITEM, so a token split
+/// into a marker plus two chunks turned one authorization into three. Measured on the
+/// maintainer's Mac — five stored items, five dialogs on first run.
+///
+/// `MAX / 2` rather than `MAX` so the `cur_bytes + add` accumulator in
+/// [`split_utf16_chunks`] cannot overflow. Any real secret is astronomically below it,
+/// so the single-chunk path always wins — and that path is what migrates existing
+/// users: [`store_value_chunked`] inlines the value into the base key and calls
+/// `clear_overflow_parts(base, 0)`, deleting the `.0`/`.1` entries left by the old
+/// layout. No separate migration step, and no half-migrated state to reason about.
+#[cfg(not(windows))]
+const SAFE_UTF16_PAYLOAD_BYTES: usize = usize::MAX / 2;
 
 /// Marker in the primary entry when the secret is split across `name.0`, `name.1`, ...
 const CHUNK_MARKER: &str = "__KEYRING_CHUNKED_V1__:";
@@ -225,4 +240,51 @@ fn clear_chunked_key(base: &str) -> Result<(), keyring::Error> {
     let base_result = delete_credential_if_present(base);
     clear_overflow_parts(base, 0);
     base_result
+}
+
+#[cfg(test)]
+mod chunking_tests {
+    use super::{split_utf16_chunks, utf16_payload_bytes, SAFE_UTF16_PAYLOAD_BYTES};
+
+    /// A Dropbox access token comfortably exceeds the Windows blob ceiling, which is why
+    /// the chunking exists at all. Build one of that shape to test against.
+    fn oversized_secret() -> String {
+        "a".repeat(3000)
+    }
+
+    #[test]
+    fn the_test_secret_really_is_over_the_windows_ceiling() {
+        // Guards the two tests below: if this ever stops holding they would both pass
+        // vacuously, proving nothing about chunking on either platform.
+        assert!(utf16_payload_bytes(&oversized_secret()) > 2400);
+    }
+
+    /// DBSYNC-81. Off Windows the keychain has no blob-size ceiling, so a secret of any
+    /// size lives in ONE item. That is the whole fix: macOS prompts per item, and three
+    /// items for one token meant three dialogs.
+    #[cfg(not(windows))]
+    #[test]
+    fn an_oversized_secret_is_not_split_off_windows() {
+        assert_eq!(split_utf16_chunks(&oversized_secret()).len(), 1);
+    }
+
+    /// Windows keeps its ceiling: `CRED_MAX_CREDENTIAL_BLOB_SIZE` is real and exceeding
+    /// it fails at runtime, not at compile time. Compiled only there, so it asserts
+    /// about the platform it describes rather than about a constant.
+    #[cfg(windows)]
+    #[test]
+    fn an_oversized_secret_is_still_split_on_windows() {
+        let chunks = split_utf16_chunks(&oversized_secret());
+        assert!(chunks.len() > 1);
+        for c in &chunks {
+            assert!(utf16_payload_bytes(c) <= SAFE_UTF16_PAYLOAD_BYTES);
+        }
+    }
+
+    /// Whatever the platform, splitting must never lose or reorder data.
+    #[test]
+    fn chunks_always_rejoin_to_the_original() {
+        let secret = oversized_secret();
+        assert_eq!(split_utf16_chunks(&secret).concat(), secret);
+    }
 }
--- a/apps/desktop/src-tauri/src/storage/secure_store.rs
+++ b/apps/desktop/src-tauri/src/storage/secure_store.rs
@@ -244,7 +244,11 @@ fn clear_chunked_key(base: &str) -> Result<(), keyring::Error> {
 
 #[cfg(test)]
 mod chunking_tests {
-    use super::{split_utf16_chunks, utf16_payload_bytes, SAFE_UTF16_PAYLOAD_BYTES};
+    use super::{split_utf16_chunks, utf16_payload_bytes};
+    // Only the Windows test asserts against the ceiling; off Windows the constant is
+    // effectively unbounded and importing it unconditionally is a dead import.
+    #[cfg(windows)]
+    use super::SAFE_UTF16_PAYLOAD_BYTES;
 
     /// A Dropbox access token comfortably exceeds the Windows blob ceiling, which is why
     /// the chunking exists at all. Build one of that shape to test against.
--- a/apps/desktop/src-tauri/src/sync_pipeline.rs
+++ b/apps/desktop/src-tauri/src/sync_pipeline.rs
@@ -567,7 +567,14 @@ pub(crate) fn mass_delete_pause_active(state: &AppState) -> AppResult<bool> {
     Ok(scan_paused || remote_paused)
 }
 
-pub(crate) fn scan_local_changes_internal(state: &AppState) -> AppResult<usize> {
+/// The LOCAL half of a scan: walk the sync folder, decide what changed, and enqueue
+/// uploads and deletions. Returns how many jobs it enqueued.
+///
+/// Split out of [`scan_local_changes_internal`] because that function does two jobs and
+/// only the second needs credentials (DBSYNC-81). Keeping them apart lets the local
+/// decisions be exercised without an access token — which is what a test of the
+/// mass-deletion guard actually wants.
+fn scan_local_changes_only(state: &AppState) -> AppResult<usize> {
     let folder = state
         .db
         .get_sync_folder()?
@@ -733,6 +740,15 @@ pub(crate) fn scan_local_changes_internal(state: &AppState) -> AppResult<usize> 
         }
     }
 
+    Ok(enqueued_jobs)
+}
+
+/// A full scan: the local half, then the remote refresh, then the bookkeeping.
+///
+/// The remote half needs an access token, so this is the entry point for production and
+/// NOT the one to reach for from a test that only cares about local decisions.
+pub(crate) fn scan_local_changes_internal(state: &AppState) -> AppResult<usize> {
+    let enqueued_jobs = scan_local_changes_only(state)?;
     let remote_enqueued = refresh_remote_index_and_enqueue_downloads_internal(state)?;
 
     {
@@ -1179,7 +1195,7 @@ mod tests {
     use super::{
         block_mass_deletion, cleanup_stale_upload_state, clear_mass_delete_blocked,
         is_mass_deletion, mass_delete_pause_active, process_changed_paths,
-        resolve_conflict_internal, run_sync_tick_internal, scan_local_changes_internal,
+        resolve_conflict_internal, run_sync_tick_internal, scan_local_changes_only,
         ConflictAction, MassDeleteSource, SYNC_BATCH_CAP,
     };
     use crate::state::AppState;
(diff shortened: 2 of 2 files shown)


mbsanchez (commented):
## Code review — CHANGES REQUESTED

*(Posted as a comment: GitHub refuses a formal review verdict on one's own pull request. The verdict is a rejection.)*

One real finding, small in severity but it makes a **claim in this PR body false**, which is the part that matters.

### 🔴 Blocking — an unused import, and a false statement about it

```
warning: unused import: `SAFE_UTF16_PAYLOAD_BYTES`
  --> src/storage/secure_store.rs:247:58
```

`SAFE_UTF16_PAYLOAD_BYTES` is imported into `chunking_tests` but referenced only inside `an_oversized_secret_is_still_split_on_windows`, which is `#[cfg(windows)]`. On this platform the import is dead.

**The PR body states "`cargo clippy --all-targets` — no new warnings". That is not true**, and the reason is a process failure worth naming: that clippy run happened **before** the chunking tests and the platform-conditional constant were added. A stale result was carried forward as if it described the final diff.

The acceptance criterion is "clippy clean, no new warnings". It fails. Hence rejection rather than an approval with a note — the number in the claim is the whole point of the criterion.

**Fix**: gate the import the same way the use is gated.

```rust
use super::{split_utf16_chunks, utf16_payload_bytes};
#[cfg(windows)]
use super::SAFE_UTF16_PAYLOAD_BYTES;
```

Then re-run clippy **after** the change, and quote that run.

### ✅ Verified — the plan's open condition is discharged

The plan flagged the self-migration as *"elegant and therefore worth distrusting — it relies on `store_value_chunked` being the only writer of these keys. Confirm that before depending on it."* Confirmed:

The whole codebase contains **four** `set_password` calls, all in `secure_store.rs`. Three are inside `store_value_chunked`; the fourth writes `SESSION_EXPIRES`, which is never chunked. Nothing else can leave a stale `.0`/`.1` behind. `clear_overflow_parts(base, 0)` sweeps indices 0–63, so both parts are removed.

### ✅ Two things the PR under-sells

**The migration fires within hours, not at next login.** `store_session` is called after every token refresh (`auth_session.rs:114`, `:182`), and Dropbox access tokens are short-lived — the code's own comment says ~4 hours. So an existing install collapses 5 items to 3 on its own, without the user doing anything. The PR only says "the next session write", which reads as vaguer than the reality.

**The change is backward compatible too.** An *older* build reading a *migrated* store finds the whole token in the base key with no `CHUNK_MARKER` prefix, so `load_value_chunked` returns it directly. A downgrade strands nobody. Worth stating, since a storage change usually raises exactly that fear.

### Checklist

| | |
| --- | --- |
| KISS / YAGNI | ✅ A constant and a function split. No abstraction invented; Step B correctly deferred. |
| Scope | ✅ Two files, both declared. |
| Secrets | ✅ None. No token value is logged or asserted on. |
| Debug output | ✅ The `panic!` probe is gone — `grep DBSYNC81` → 0. |
| Security | ✅ No IPC surface, no capability change. The split makes the credential-touching path *narrower*, not wider. |
| Rust idioms | ⚠️ One unused import — the finding above. Otherwise no `unsafe`, no `unwrap()` outside tests. |
| Tests | ✅ Four, including a **guard** (`the_test_secret_really_is_over_the_windows_ceiling`) that stops the two platform tests passing vacuously. The Windows arm was compiled and executed via a `cfg` pivot, which is the only way to test code that never compiles on this host. |

### On the `panic!`-under-`cfg(test)` technique

Worth keeping. It names an offending test through the harness's thread name, aborts before the protected resource is touched, and costs zero dialogs — where prompt-watching produced a contradictory result that turned out to be a miscount.

### Verdict

**Changes requested** — one fix, then re-run clippy and quote the run that describes the final diff.

Next: `/cowork-workflow:jira-ticket-implement`

mbsanchez:
## Review fix pushed — 609325c

Addresses the single blocking finding from the CTO review.

```rust
use super::{split_utf16_chunks, utf16_payload_bytes};
#[cfg(windows)]
use super::SAFE_UTF16_PAYLOAD_BYTES;
```

The import now carries the same cfg as its only user, so it is no longer dead off Windows.

### Re-validated against the final diff

The original failure was one of provenance, not of code — the clippy output quoted in the PR body came from a run that predated the chunking tests. So this run was made deliberately un-cacheable with `touch src/lib.rs`:

- **clippy** — the `SAFE_UTF16_PAYLOAD_BYTES` warning is gone. Four warnings remain and I am not calling that "clean": `cloudsc_ops.rs:1174`, `cloudsc_ops.rs:860`, `dropbox_transfer.rs:3`, `open_handlers.rs:74`. All pre-existing — the branch diff touches only `secure_store.rs` and `sync_pipeline.rs`, and none of the four is in either. `dropbox_transfer.rs` was last modified by DBSYNC-66 (f5aa98e); the two dead-code ones have platform-gated callers at `lib.rs:239` and `cloudsc_ops.rs:850`. The AC is "no new warnings", and that holds.
- **`cargo test`** — 177 passed, 0 failed.
- **Zero keychain dialogs** — `pgrep -x SecurityAgent` polled once a second across the whole suite: 0 samples up. ACL counts after: `access` 8, `access.0` 7, `access.1` 7, `refresh` 7, `expires` 7 — unchanged from the Slice 1 post-measurement. A new ACL entry requires a dialog, and there were none.

The PR body has been corrected rather than quietly overwritten: the clippy bullet now records that it was false when first written, and the "`security dump-keychain` hangs" caveat is struck through, since it responded normally this time.

### Unchanged disclosure

The suite still clobbers the live `overlay_state.json` (**DBSYNC-75**, out of scope here). Backed up before the run and restored byte-for-byte after.

mbsanchez:
## CTO Review (re-review of 609325c): APPROVED ✅

The blocking finding is fixed and I verified it rather than taking the commit message's word for it.

### The finding, closed

`cargo clippy --all-targets` re-run with `touch` on both changed files so nothing came from cache — `grep -c SAFE_UTF16_PAYLOAD_BYTES` over the output returns **0**. Four warnings remain, in `cloudsc_ops.rs` (×2), `dropbox_transfer.rs` and `open_handlers.rs` — none in a file this branch touches, so none is new. The AC reads "no new warnings", and it is met. `cargo test`: **177 passed, 0 failed**.

The cfg gate matches the constant's only consumer in that module, which is the narrowest fix available. It does not paper over the warning with `#[allow]`.

### What I checked beyond the delta

**The split did not weaken production.** `scan_local_changes_internal` keeps both real callers — `commands.rs:476` (the Tauri command) and `sync_pipeline.rs:1128` (the sync tick). Both still run the full local + remote path; only the test moved to the local half. `scan_local_changes_only` is private, so the narrower entry point cannot leak into production by accident.

**The overflow reasoning in the doc comment holds, and is stronger than it claims.** With `usize::MAX / 2`, `cur_bytes + add` at `secure_store.rs:169` cannot overflow — `cur_bytes` stays ≤ MAX/2 by the loop condition and `add` is at most 4. But the early return at `:161` means that off Windows the loop is never entered at all: the single-chunk path always wins. The invariant is belt *and* braces.

**The duplicated `2400` in `the_test_secret_really_is_over_the_windows_ceiling` is correct, not a DRY slip.** It cannot reference the constant, because off Windows the constant is `MAX / 2` and the assertion would pass vacuously — exactly the failure the guard exists to prevent.

### On the PR body

The stale clippy claim was corrected in place with an explicit note that it was false when written, rather than silently overwritten. That is the right disposition: the review record stays readable.

### Carried forward, not blocking

- **#101 (Slice 3) is still open** — the item-count drop is predicted from the code, not observed. Approving the code does not tick that box.
- **First run still costs three prompts, not one.** Documented as deliberate; collapsing the last three items is a real schema migration and correctly deferred.
- **Nothing here is validated on Windows** beyond the cfg-pivot execution recorded earlier. The Windows arm is unchanged by this commit.

Moving DBSYNC-81 to QA.

mbsanchez:
## QA Result: PASSED ✅ (with one AC explicitly not met)

Manual desktop QA on macOS 26.5.1 against 609325c, bundled and signed with the personal Developer ID (`XCAA3WMJM6`) and installed to `~/Applications`. Full evidence on issue #101.

**4 of 5 acceptance criteria met.** The fifth — "first run prompts exactly once" — is **not met and cannot be met by this change**: it describes Step B, which was deliberately deferred. I did not tick it and did not reword it to fit. It needs its own issue rather than sitting unticked on #101.

### The headline: the item-count drop is now observed, not predicted

The PR body says "the item-count drop is predicted from the code, not observed". That is no longer true — it can be updated.

```
5 items -> 3 items   (dropbox-session-access.0 and .1 are gone)
```

Causally attributed rather than assumed: the migration needs a session *write*, and the natural token refresh at `18:37:07.726299Z` supplied one — the surviving `dropbox-session-access` item carries `mdat = 20260820183708Z`, one second later. The firing time was predicted at 18:36:49Z **before** it happened, from `auth_session.rs:21`, and landed 18 s off.

### Zero prompts, twice

`pgrep -x SecurityAgent` polled once a second across two full quit/relaunch cycles, with a before/after pid guard so a dead app cannot be mistaken for a clean exit:

```
pre-migration   59252 -> (empty) -> 63491    0 samples
post-migration  63491 -> (empty) -> 75593    0 samples
```

The second matters most: it is the first read of the **migrated 3-item layout** from disk.

### Still signed in — proven, not inferred

Two authenticated Dropbox round trips, one per layout, each with a real `rev`: `6597e852610f956342f44` (old layout) and `6597edbac3cca56342f44` (migrated). Both probe files were then deleted and the deletions propagated, leaving nothing in the account.

`cargo test` (177 tests) added no ACL entry: full dump before/after, all five items identical, zero dialogs.

### One measurement I got wrong

My first post-migration dump said **0 items** — which would have meant total credential loss. It was a truncated `security dump-keychain` (3283 lines vs a 9958-line baseline), the same flakiness flagged earlier in this PR. Discarded, not reported. The replacement — `security find-generic-password -s dropbox-sync-desktop -a <account>`, attribute-only so it does not prompt — agreed with the watcher's independent grep at exactly 3 items.

Ready for PO approval.

mbsanchez:
## Correction to my QA verdict: AC 1 is withdrawn, not failed

The maintainer asked a question that broke my own framing: *does installing the app and running it the first time prompt five times to **save** the items?*

**No — saving never prompts.** Creating a keychain item adds the creator to its ACL automatically; the dialog only ever comes from a **read** by a binary that is not in that ACL. Verified with three throwaway items and a `SecurityAgent` poller: create → 0 dialogs, read back with `-w` → 0 dialogs, all three secrets returned. Probes deleted.

So the five dialogs never came from the item count on a first run. They came from `cargo test` binaries, whose ACL entries are **cdhash-keyed** and invalidated by every recompile — six of them are still sitting in the ACL, five with broken signatures. Same reason _Always Allow_ seemed not to stick: it stuck to a cdhash the next build destroyed.

### What this means for this PR

**Nothing in the code changes, and the approval stands.** But the framing on the ticket did, so the honest accounting is:

- **The end-user benefit of 5 → 3 items is zero prompts saved.** They already paid zero. The 5→3 drop is real, observed, and correct — it just does not buy what the ticket claimed.
- **The change that actually removes the maintainer's pain is the Slice 1 half of this same PR** — stopping the test suite from reaching the keychain. That is the fix; the chunking change is blast-radius reduction.
- **AC 1 of #101 is withdrawn as mis-specified**, not recorded as a failure. Its expected value for a signed release is zero, not one.
- **Step B should be reassessed rather than scheduled** — noted on #100.

DBSYNC-81's title and description have been corrected, including the retraction of the "five items, five prompts" claim I wrote into it.

Issue #105, opened 2026-08-21
DBSYNC-84 Slice 3: Live validation on the Mac, without navigating away


mbsanchez:
Vertical slice 3 of **DBSYNC-84**. Type: **manual / HITL**.

* Scope dir(s): none — validation only
* Blocked by: Slice 2 ([#104](https://github.com/mbsanchez/DropboxSync/issues/104))
* User stories: 1, 2, 3

## Acceptance criteria

- [ ] Hydrating a file updates its badge to the green check **without leaving the directory**.
- [ ] A `.cloudsc` placeholder shows its `cloud_only` badge as soon as it appears.
- [ ] A path removed from tracking loses its badge.
- [ ] **Evidence is a screen recording, or before/after screenshots captured without navigating away.**
- [ ] Idle CPU of the extension measured before and after; no busy-loop introduced.

## Why the evidence rule is an AC and not a suggestion

**Navigating away and back is the gesture that fixes the symptom.** A screenshot taken after navigating shows a correct badge whether or not the fix works. It would prove nothing and look like proof.

This is not a hypothetical failure mode on this project:

* DBSYNC-72 Slice 3 ticked *"the extension can be enabled and `synced` renders"* — true, and it could not have caught this bug, because the check navigated to a directory and looked.
* DBSYNC-72's sandbox ADR recorded an empirical check that exercised the configuration being chosen rather than the one being rejected. It failed exactly where it had not looked.
* On 2026-08-19 the Finder badges were reported working twice before anyone had actually looked.

The rule to apply: **the check must exercise the branch that can fail.** Here that means the badge changing while the directory stays open.

## How to measure idle CPU

The extension runs inside its own process managed by PlugInKit, not inside the app. Find it and sample it:

```
pgrep -f DropboxSyncFinderSync
top -pid <pid> -l 5 -stats pid,cpu,command
```

Take the reading with Finder open on the sync folder and nothing syncing. The diff now runs every 2 seconds where before it only re-read the file, so this is a real thing to check rather than a formality — though the `updated_at` short-circuit from Slice 1 should make it a non-event.

Jira: DBSYNC-84

mbsanchez:
## Implementation Plan

**Scope dir(s)**: none — validation only
**Implementing agent(s)**: the maintainer, on a Mac. No agent.
**Estimated complexity**: low to run, high to run *honestly* — see below.
**Dependencies**: blocked by Slice 2 (#104).

### Steps

[...]

Issue #104, opened 2026-08-21
DBSYNC-84 Slice 2: Clear the badge when a path leaves the state


mbsanchez:
Vertical slice 2 of **DBSYNC-84**. Type: **AFK**.

* Scope dir(s): `apps/desktop/native/macos/FinderSyncExtension`
* Blocked by: Slice 1 ([#103](https://github.com/mbsanchez/DropboxSync/issues/103)) — this slice consumes the diff's "removed" set
* User stories: 3

## The defect

```swift
guard let tier = state?.paths[relative] else { return }
```

`SyncExtension.swift:139`. When a path disappears from the state map this returns **without clearing the badge**, so Finder keeps showing the old one indefinitely.

Not yet observed in the wild — unlike the stale-update bug in Slice 1, which the maintainer reproduced. It is the same class of defect and the diff from Slice 1 already produces the set of removed paths, so fixing it here costs almost nothing.

## Acceptance criteria

- [ ] Paths present in the old snapshot and absent from the new one have their badge cleared.
- [ ] `requestBadgeIdentifier(for:)` clears the badge for an untracked path instead of returning silently.
- [ ] The pure diff function from Slice 1 reports removals, covered by the same `xcrun swift` script.
- [ ] **Clearing uses the documented mechanism.** Passing an empty-string identifier is the folklore answer; verify it against Apple's documentation before relying on it, and record what was verified. If it is not documented, find the mechanism that is.
- [ ] `npm --workspace apps/desktop run build:finder-sync` returns BUILD SUCCEEDED for both architectures.

## Note on sequencing

This is second, not first, deliberately. Slice 1 fixes what the maintainer actually hit; this fixes a defect found by reading. If the clearing mechanism turns out to be awkward, it must not hold up the change that matters.

Jira: DBSYNC-84

mbsanchez:
## Implementation Plan

**Scope dir(s)**: `apps/desktop/native/macos/FinderSyncExtension`
**Implementing agent(s)**: Swift / AppKit
**Estimated complexity**: low — *provided* the clearing mechanism turns out to be documented. If it is not, this becomes medium and should not hold up Slice 1.
**Dependencies**: blocked by Slice 1 (#103) — consumes the `removed` set the diff already returns.

### Steps

[...]

Issue #103, opened 2026-08-21
DBSYNC-84 Slice 1: Push badge updates when the state changes


mbsanchez:
Vertical slice 1 of **DBSYNC-84**. Type: **AFK**.

* Scope dir(s): `apps/desktop/native/macos/FinderSyncExtension`
* Blocked by: none
* User stories: 1, 2, 4

## The fix

`reloadState()` keeps the previous snapshot, diffs it against the new one, and calls `setBadgeIdentifier` for every path whose tier changed. Today `setBadgeIdentifier` is called from exactly one place — `requestBadgeIdentifier(for:)` (`SyncExtension.swift:141`) — which Finder invokes once per item and then caches. Nothing ever pushes an update.

## Acceptance criteria

- [ ] The tier diff is a **pure function** — old map + new map → changed paths — with no `FIFinderSyncController` dependency.
- [ ] That function is exercised by a script run with `xcrun swift`, following the precedent set by DBSYNC-72 Slice 1 (path equivalence proven at runtime rather than by grep). Cases: added, changed, unchanged, empty-to-populated.
- [ ] `reloadState()` pushes `setBadgeIdentifier` for every changed path.
- [ ] **The first load pushes nothing.** Previous state is `nil`, so every path reads as changed — and Finder is about to ask anyway. DBSYNC-80 measured the enumeration at 50,000 files, so a first-load push storm is a measured risk, not a hypothetical one.
- [ ] **Short-circuit on `updated_at`.** The Rust side stamps every snapshot (`overlay_state.rs`); if it has not changed, skip the diff rather than re-comparing the whole map every 2 seconds.
- [ ] Pushes are bounded to paths under the monitored sync folder.
- [ ] No Rust changes at all — `cargo test` and `cargo clippy` unaffected, provable by reading the diff.
- [ ] `npm --workspace apps/desktop run build:finder-sync` returns BUILD SUCCEEDED for both architectures.

## Explicitly NOT in this slice

**Do not replace the 2-second poll with a file watcher.** It is tempting and it is not the bug. Changing the trigger and the mechanism in one change makes it impossible to attribute the fix to either. The poll stays.

## Why this has no unit tests

[...]

PR #106, opened 2026-08-21
fix(#DBSYNC-84): push Finder badge updates instead of only answering when asked

mbsanchez:
## Summary

Finder badges were correct **once** and then frozen — a file that finished syncing kept its old badge until the user left the directory and came back. Reproducible, not flaky.

`setBadgeIdentifier` was called from exactly one place, `requestBadgeIdentifier(for:)`, which Finder invokes **once per item** and then caches. The 2-second poll refreshed the extension's own state but never told Finder anything.

`reloadState()` now keeps the previous snapshot, diffs it, and pushes what changed. Paths that leave the state get their badge cleared.

Closes slices [#103](https://github.com/mbsanchez/DropboxSync/issues/103) and [#104](https://github.com/mbsanchez/DropboxSync/issues/104). [#105](https://github.com/mbsanchez/DropboxSync/issues/105) — live validation — is deliberately **not** covered here; see below.

## Two things the plan got wrong

Both were caught by doing what the plan itself demanded: check before shipping.

### 1. Pushing every change violates Apple's documented rule

`FinderSync.h` in the SDK, alongside `setBadgeIdentifier:forURL:`:

> Avoid adding badges to items that the Finder hasn't displayed yet. When setting the initial badge, call this method from your Finder Sync extension's `requestBadgeIdentifierForURL:` method. **When updating badges, call this method only for items that have already received a badge.**

The plan said to push for every changed path. That would badge items Finder has never drawn — a file whose tier flips while its folder is closed, for instance.

Fixed with `BadgeDiff.pushable(_:alreadyBadged:)` filtering against a `badgedPaths` set that `requestBadgeIdentifier(for:)` populates. Kept as a **pure function** so the rule is enforced by tests rather than by a comment somebody can delete.

The same header settled #104's open question at the same time — *"Setting the identifier to an empty string (`@""`) removes the badge"* is **documented**, so the clearing mechanism is not folklore. Cited in the code.

### 2. The planned test command runs nothing while reporting success

```
xcrun swift BadgeDiff.swift Tests/BadgeDiffTests.swift   -> exit 0, no output, nothing executed
xcrun swift Tests/BadgeDiffTests.swift BadgeDiff.swift   -> error: cannot find 'BadgeDiff' in scope
xcrun swiftc -o out BadgeDiff.swift Tests/…              -> error: statements are not allowed at the top level
```

The first is the dangerous one: **exit 0 having tested nothing looks exactly like a pass.** Resolved with `@main` + `swiftc`, wrapped in `Tests/run-badge-diff-tests.sh`, with the caveat written into the script header.

## Stack

`desktop-tauri` — but this change is **Swift/AppKit only**. Flagged at the architecture step: neither `tauri-expert` nor `rust-pro` from the stack profile covers it.

## Jira Ticket

DBSYNC-84

## Test Plan

- [x] **`./Tests/run-badge-diff-tests.sh` — 13 checks, all passing.**
- [x] **The checks can actually fail** — verified by mutation, not assumed:
  ```
  break the first-load guard          -> 1 check FAILED
  break the "only differences" filter -> 4 checks FAILED
  restored                            -> all checks passed
  ```
- [x] **The new code is genuinely linked into the binary.** `BUILD SUCCEEDED` does not prove this: a file missing from the Sources build phase compiles to nothing and the build still succeeds. `project.pbxproj` here lists files **manually** (no `fileSystemSynchronizedGroups`), so registration is four hand edits and a real failure mode.
  ```
  nm -a DropboxSyncFinderSync | grep -ci BadgeDiff  ->  72
  lipo -archs                                       ->  x86_64 arm64
  ```
- [x] `npm run build:finder-sync` — `** BUILD SUCCEEDED **`, `Principal class OK: DropboxSyncFinderSync`.
- [x] **Rust untouched**, as the slice requires — `git status` shows 0 modified `.rs` files. `cargo test` 177 passed; `cargo clippy` reports the same 4 pre-existing warnings, none new.

### Not tested here

**Nothing in this PR proves a badge visibly changes in Finder.** That is slice #105, it is manual, and its acceptance criterion is deliberately awkward: evidence must be captured **without navigating away from the directory**, because navigating away is the gesture that makes a stale badge look correct. A screenshot taken after navigating would prove nothing and look like proof.

That is not hypothetical on this ticket — DBSYNC-72 Slice 3 ticked *"`synced` renders"*, which was true and structurally incapable of catching this bug.

## Design notes for review

- **The 2-second poll stays.** Replacing it with a file watcher was floated and rejected: it is not the bug, and changing trigger and mechanism together destroys attribution. Recorded in the ticket's Implementation Decisions.
- **First load pushes nothing.** With no previous snapshot every path reads as changed; DBSYNC-80 measured the enumeration at 50,000 files, so this is a measured risk. Encoded in `BadgeDiff.changes`' contract (`nil` in → nothing out) rather than at the call site, where a future caller could forget it.
- **`updated_at` short-circuit.** The Rust writer stamps every snapshot, so an unchanged stamp skips the diff entirely — one string compare per tick instead of a full map comparison.
- **`lastUpdatedAt` is cleared whenever `state` is.** Otherwise a file that becomes unreadable and returns with the same stamp would skip the diff forever and never push again.
- **Path containment is enforced, not assumed.** A relative path containing `..` would otherwise let the state file badge arbitrary locations.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
+++ b/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
@@ -1,0 +1,64 @@
+import Foundation
+
+/// Works out which badges Finder needs to be told about, given two consecutive
+/// `overlay_state.json` snapshots (DBSYNC-84).
+///
+/// Deliberately imports **Foundation only**. It must not reference `FinderSync` or
+/// `FIFinderSyncController`: a file that imports FinderSync cannot be compiled as a script
+/// by `xcrun swift`, and running it that way is the only automated coverage available here
+/// — there is no Swift test target, and adding one is out of scope per DBSYNC-72. Keeping
+/// this file pure is what makes `Tests/BadgeDiffTests.swift` possible, not a style choice.
+enum BadgeDiff {
+    /// What changed between two snapshots of the relative-path → tier map.
+    struct Changes: Equatable {
+        /// Paths whose badge must be set, with the tier to set it to.
+        let changed: [String: String]
+        /// Paths that were tracked and no longer are; their badge must be cleared.
+        let removed: [String]
+
+        var isEmpty: Bool { changed.isEmpty && removed.isEmpty }
+    }
+
+    /// Compares `old` against `new` and returns only what Finder has to be told.
+    ///
+    /// **A `nil` `old` yields no changes at all**, on purpose. On the first load every path
+    /// would otherwise count as changed, and Finder is about to ask for each visible item
+    /// anyway — so pushing would be redundant work at exactly the worst moment. DBSYNC-80
+    /// measured the enumeration at 50,000 files, so a first-load push storm is a measured
+    /// risk rather than a hypothetical one. Encoding it here rather than at the call site
+    /// means a future caller cannot forget it.
+    static func changes(from old: [String: String]?, to new: [String: String]) -> Changes {
+        guard let old else {
+            return Changes(changed: [:], removed: [])
+        }
+
+        var changed: [String: String] = [:]
+        for (path, tier) in new where old[path] != tier {
+            changed[path] = tier
+        }
+
+        let removed = old.keys.filter { new[$0] == nil }.sorted()
+
+        return Changes(changed: changed, removed: removed)
+    }
+
+    /// Narrows `changes` to the items Finder has **already badged**.
+    ///
+    /// Apple's own guidance, in `FinderSync.h` alongside `setBadgeIdentifier:forURL:`:
+    ///
+    /// > Avoid adding badges to items that the Finder hasn't displayed yet. When setting the
+    /// > initial badge, call this method from your Finder Sync extension's
+    /// > `requestBadgeIdentifierForURL:` method. When updating badges, call this method only
+    /// > for items that have already received a badge.
+    ///
+    /// So a push is only legitimate for a path Finder previously asked about. Anything else
+    /// is handled by `requestBadgeIdentifier(for:)` when Finder gets round to displaying it.
+    /// Without this filter a file that changes tier while its folder is closed would be
+    /// pushed at Finder for an item it has never drawn.
+    static func pushable(_ changes: Changes, alreadyBadged: Set<String>) -> Changes {
+        Changes(
+            changed: changes.changed.filter { alreadyBadged.contains($0.key) },
+            removed: changes.removed.filter { alreadyBadged.contains($0) }
+        )
+    }
+}
--- a/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
+++ b/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
@@ -42,7 +42,7 @@ enum BadgeDiff {
         return Changes(changed: changed, removed: removed)
     }
 
-    /// Narrows `changes` to the items Finder has **already badged**.
+    /// Narrows `changes` to the items Finder is **displaying**.
     ///
     /// Apple's own guidance, in `FinderSync.h` alongside `setBadgeIdentifier:forURL:`:
     ///
@@ -51,14 +51,19 @@ enum BadgeDiff {
     /// > `requestBadgeIdentifierForURL:` method. When updating badges, call this method only
     /// > for items that have already received a badge.
     ///
-    /// So a push is only legitimate for a path Finder previously asked about. Anything else
-    /// is handled by `requestBadgeIdentifier(for:)` when Finder gets round to displaying it.
-    /// Without this filter a file that changes tier while its folder is closed would be
-    /// pushed at Finder for an item it has never drawn.
-    static func pushable(_ changes: Changes, alreadyBadged: Set<String>) -> Changes {
+    /// The operative constraint is the first sentence: **displayed**. `displayed` therefore
+    /// holds every path Finder has asked about, not only the ones that came back with a
+    /// badge.
+    ///
+    /// Reading "already received a badge" literally is a trap, and this filter fell into it
+    /// once: a `.cloudsc` placeholder appears, Finder asks about it before the state file
+    /// has caught up, so nothing is badged — and when the tier arrives moments later the
+    /// path is filtered out and never badged at all. Finder does not ask twice. That is
+    /// symptom 1 of DBSYNC-84, reintroduced by the very filter meant to respect Apple's rule.
+    static func pushable(_ changes: Changes, displayed: Set<String>) -> Changes {
         Changes(
-            changed: changes.changed.filter { alreadyBadged.contains($0.key) },
-            removed: changes.removed.filter { alreadyBadged.contains($0) }
+            changed: changes.changed.filter { displayed.contains($0.key) },
+            removed: changes.removed.filter { displayed.contains($0) }
         )
     }
 
@@ -99,18 +104,51 @@ enum BadgeDiff {
     /// comes back. Which is the exact bug this ticket exists to fix, reached through a
     /// narrower door.
     ///
-    /// `alreadyBadged` is not lost in that reset, so the displayed items can be refreshed
-    /// directly. Every path here has already received a badge, so pushing to it is
-    /// legitimate under Apple's rule.
-    static func resync(to new: [String: String], alreadyBadged: Set<String>) -> Changes {
+    /// `displayed` is not lost in that reset, so those items can be refreshed directly.
+    /// Every path here is one Finder is showing, so pushing to it is legitimate.
+    static func resync(to new: [String: String], displayed: Set<String>) -> Changes {
         var changed: [String: String] = [:]
-        for path in alreadyBadged {
+        for path in displayed {
             if let tier = new[path] { changed[path] = tier }
         }
         return Changes(
             changed: changed,
-            removed: alreadyBadged.filter { new[$0] == nil }.sorted()
+            removed: displayed.filter { new[$0] == nil }.sorted()
         )
+    }
+
+    /// What to do when Finder asks about `path`.
+    ///
+    /// This exists because the placement of a single line was load-bearing and untestable.
+    /// Twice now a defect has lived in `SyncExtension.swift` where the check script cannot
+    /// reach it: the string-prefix containment bug, and then recording a path as displayed
+    /// only *after* a tier was found. The second one passed 27 green checks and a clean
+    /// build. Moving the decision here is what makes it a fact under test rather than a
+    /// property of where a statement happens to sit.
+    struct RequestOutcome: Equatable {
+        /// The relative path Finder is now displaying, or `nil` if it is outside the root.
+        ///
+        /// **Non-nil even when there is no tier to set.** Finder asking about an item means
+        /// it is on screen, and it will not ask twice — so a `.cloudsc` drawn before the
+        /// state file catches up must still count as displayed, or the tier arriving a
+        /// moment later is filtered out of the push and the file stays bare forever.
+        let displayedPath: String?
+        /// The identifier to hand to `setBadgeIdentifier`: a tier, or `""` to clear.
+        /// `nil` means do nothing at all.
+        let badgeIdentifier: String?
+    }
+
+    static func requestOutcome(
+        for path: String,
+        root: String,
+        paths: [String: String]?
+    ) -> RequestOutcome {
+        guard let relative = relativePath(of: path, under: root), !relative.isEmpty else {
+            return RequestOutcome(displayedPath: nil, badgeIdentifier: nil)
+        }
+        // `""` clears the badge — documented in `FinderSync.h`. An item we do not track
+        // must not keep a badge from a previous state.
+        return RequestOutcome(displayedPath: relative, badgeIdentifier: paths?[relative] ?? "")
     }
 }
 
--- a/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
+++ b/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
@@ -61,4 +61,56 @@ enum BadgeDiff {
             removed: changes.removed.filter { alreadyBadged.contains($0) }
         )
     }
+
+    /// Whether `path` is `root` itself or lives inside it.
+    ///
+    /// A plain `path.hasPrefix(root)` is **not** a containment test and was shipped as one
+    /// in the first version of this change: with a root of `/Users/x/DropboxSync`, the path
+    /// `/Users/x/DropboxSyncEvil/secret.txt` passes a string-prefix check while living in a
+    /// different directory entirely. Comparing against `root + "/"` is what makes the
+    /// sibling-with-a-longer-name case fail as it should.
+    ///
+    /// This lives in `BadgeDiff` rather than next to its callers so that it is reachable
+    /// from `Tests/BadgeDiffTests.swift`. That is the point: the string-prefix bug survived
+    /// review precisely because it sat in a file the test script cannot compile.
+    static func isContained(_ path: String, in root: String) -> Bool {
+        if path == root { return true }
+        let rootWithSeparator = root.hasSuffix("/") ? root : root + "/"
+        return path.hasPrefix(rootWithSeparator)
+    }
+
+    /// The portion of `path` below `root`, or `nil` if `path` is not inside `root`.
+    ///
+    /// Returns `""` for `root` itself. The old inline version sliced off `root.count`
+    /// characters after a bare prefix check, so `/Users/x/DropboxSyncEvil/f` yielded the
+    /// nonsensical relative path `Evil/f` and was then looked up in the state map.
+    static func relativePath(of path: String, under root: String) -> String? {
+        guard isContained(path, in: root) else { return nil }
+        if path == root { return "" }
+        let rootWithSeparator = root.hasSuffix("/") ? root : root + "/"
+        return String(path.dropFirst(rootWithSeparator.count))
+    }
+
+    /// What to push when the previous snapshot was lost but Finder is still showing badges.
+    ///
+    /// `reloadState()` drops its snapshot whenever the state file cannot be read. Without
+    /// this, the next successful read produces no diff — `changes(from: nil, …)` is empty by
+    /// design — and every visible badge stays frozen until the user leaves the directory and
+    /// comes back. Which is the exact bug this ticket exists to fix, reached through a
+    /// narrower door.
+    ///
+    /// `alreadyBadged` is not lost in that reset, so the displayed items can be refreshed
+    /// directly. Every path here has already received a badge, so pushing to it is
+    /// legitimate under Apple's rule.
+    static func resync(to new: [String: String], alreadyBadged: Set<String>) -> Changes {
+        var changed: [String: String] = [:]
+        for path in alreadyBadged {
+            if let tier = new[path] { changed[path] = tier }
+        }
+        return Changes(
+            changed: changed,
+            removed: alreadyBadged.filter { new[$0] == nil }.sorted()
+        )
+    }
 }
+
--- a/apps/desktop/native/macos/FinderSyncExtension/DropboxSyncFinderSync.xcodeproj/project.pbxproj
+++ b/apps/desktop/native/macos/FinderSyncExtension/DropboxSyncFinderSync.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		DBFC00000000000000000021 /* SyncExtension.swift in Sources */ = {isa = PBXBuildFile; fileRef = DBFC00000000000000000001 /* SyncExtension.swift */; };
+		DBFC00000000000000000030 /* BadgeDiff.swift in Sources */ = {isa = PBXBuildFile; fileRef = DBFC00000000000000000012 /* BadgeDiff.swift */; };
 		DBFC00000000000000000022 /* cloud-check.png in Resources */ = {isa = PBXBuildFile; fileRef = DBFC00000000000000000002 /* cloud-check.png */; };
 		DBFC00000000000000000023 /* cloud-alert.png in Resources */ = {isa = PBXBuildFile; fileRef = DBFC00000000000000000003 /* cloud-alert.png */; };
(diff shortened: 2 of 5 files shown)


mbsanchez:
## CTO Review: CHANGES REQUESTED ❌ — 1 blocking, 1 worth fixing

Reviewing my own PR, so GitHub will not take a formal verdict; this comment is the verdict.

The good parts hold up: the diff is genuinely pure and genuinely tested, the mutation check proves the tests can fail, `nm` proves the code is linked rather than trusting BUILD SUCCEEDED, and Rust is provably untouched. The two plan corrections were the right calls.

But there is a real defect in code this PR introduces.

---

### 🔴 BLOCKING — the path containment guard does not contain, and a comment claims it does

`SyncExtension.swift`, in the new `url(forRelative:under:)`:

```swift
/// The state file is written by the app, but a relative path containing `..` would
/// otherwise let it badge arbitrary locations on disk, so the containment is enforced
/// here rather than assumed.
private func url(forRelative relative: String, under root: URL) -> URL? {
    let url = root.appendingPathComponent(relative).standardizedFileURL
    guard url.path.hasPrefix(root.path) else { return nil }
    return url
}
```

`hasPrefix(root.path)` is a **string** prefix test, not a path containment test. A sibling directory whose name extends the root's name passes it. Measured, not reasoned:

```
rel=ok.txt                        -> /Users/x/DropboxSync/ok.txt          guard=true  inside=true   ok
rel=../DropboxSyncEvil/secret.txt -> /Users/x/DropboxSyncEvil/secret.txt  guard=true  inside=false  *** WRONG ***
rel=../../../etc/passwd           -> /etc/passwd                          guard=false inside=false  ok
rel=sub/../ok.txt                 -> /Users/x/DropboxSync/ok.txt          guard=true  inside=true   ok
```

**Failure scenario**: with a sync folder at `~/DropboxSync`, a state entry of `../DropboxSyncEvil/secret.txt` badges a file in `~/DropboxSyncEvil` — outside the monitored folder entirely. Deep traversal (`../../../etc/passwd`) is caught; the sibling-prefix case is not.

**Why this is blocking rather than a nit.** The practical risk is low — the state file is written by our own Rust side, not by an attacker. What makes it blocking is the comment: it states that containment *is enforced here rather than assumed*, and that is false. A guard that does not guard is a bug; a guard that does not guard **while documenting that it does** actively stops the next person from looking. This ticket exists because DBSYNC-72's ADR did the same thing in prose.

**Fix**: compare with the trailing separator, or better, compare path components:

```swift
guard url.path == root.path || url.path.hasPrefix(root.path + "/") else { return nil }
```

Add the sibling-prefix case to `Tests/BadgeDiffTests.swift` — or, since this lives in `SyncExtension.swift` and cannot be reached by the test script, move the containment check into `BadgeDiff` where it *can* be tested. The second option is better: this is the second time a piece of untestable logic in this file has been wrong.

**The same pattern pre-exists** in `requestBadgeIdentifier(for:)`:

```swift
guard item.path.hasPrefix(root.path) else { return }
var relative = item.path
relative.removeFirst(root.path.count)
```

For `/Users/x/DropboxSyncEvil/f` this yields the relative path `Evil/f`, which is nonsense. Not introduced by this PR, but this PR is where the function was touched and it should be fixed in the same pass rather than left as a matching landmine.

---

### 🟡 Worth fixing — a transient read failure reintroduces the exact bug being fixed

`reloadState()` sets `state = nil` and `lastUpdatedAt = nil` when the file cannot be read. On the next successful read, `previousPaths` is `nil`, so `BadgeDiff.changes` correctly returns nothing and **no badges are pushed**.

But `badgedPaths` is *not* cleared — the extension still knows exactly which items Finder has displayed. So it has the information needed to refresh them and deliberately declines to.

**Failure scenario**: the state file is momentarily unreadable (mid-write, permissions blip). Afterwards, every visible badge stays frozen until the user leaves the directory and comes back — which is verbatim the bug this PR exists to fix, just through a narrower door.

Pushing to those paths would be legitimate under Apple's rule, since they have already received a badge. Suggested: when `previousPaths` is `nil` but `badgedPaths` is not empty, push the current tier for every badged path.

---

### 🟢 Notes, not objections

- **`badgedPaths` grows without bound** — one entry per distinct path Finder has ever asked about in the sync folder. For a 50,000-file store browsed exhaustively that is a few MB of strings, and entries are only removed on clear. Acceptable; worth a comment so it is a known choice rather than an oversight.
- **The `updated_at` short-circuit returns before `state = decoded`**, leaving the previous object in place. Correct, since an identical stamp means identical content — but it is load-bearing on the Rust side always stamping a new value on every write. Worth one line saying so.
- **Scope is clean**: five files, all under `FinderSyncExtension/`. No secrets, no debug output, no Rust.

---

Moving DBSYNC-84 back to In Progress.

mbsanchez:
## Review fixes pushed — bcb7f0f

Both findings addressed, plus the pre-existing instance of the same bug that the review flagged in passing.

### 🔴 Blocking — path containment

```swift
guard BadgeDiff.isContained(url.path, in: root.path) else { return nil }
```

`isContained` compares against `root + "/"`, and `BadgeDiff.relativePath(of:under:)` replaces the inline character-slicing in `requestBadgeIdentifier(for:)`.

**The important half is that both moved into `BadgeDiff`.** The review's actual point was that the bug survived because it lived where `Tests/BadgeDiffTests.swift` cannot compile — fixing the comparison without fixing that would leave the next one just as invisible. `grep hasPrefix SyncExtension.swift` now returns only the comment explaining why.

Proven by reintroducing the exact bug:

```
=== bare hasPrefix restored ===
FAIL a SIBLING whose name extends the root is NOT contained — expected false, got true
FAIL a sibling yields nil, not a nonsense relative path — expected nil, got Optional("vil/secret.txt")
BadgeDiff: 2 check(s) FAILED
=== restored ===
BadgeDiff: all checks passed
```

That second line is the pre-existing defect, now visible: slicing `root.count` characters after a bare prefix check eats one character too few, so `/Users/x/DropboxSyncEvil/secret.txt` became the relative path `vil/secret.txt` and was looked up in the state map as if it were real. Same root cause, same test.

### 🟡 Transient read failure

`BadgeDiff.resync(to:alreadyBadged:)`, called when the snapshot was lost but `badgedPaths` is not empty. Every path it touches has already received a badge, so it stays inside Apple's rule.

### 🟢 The two notes

Both were about missing comments rather than behaviour, and both are now written down.

### Validation

**26 checks passing**, up from 13 — `isContained` (6), `relativePath` (4), `resync` (3) are new.

```
** BUILD SUCCEEDED **   Principal class OK: DropboxSyncFinderSync
lipo -archs              ->  x86_64 arm64
nm | grep -ci BadgeDiff  ->  82
```

**Rust untouched**: 0 modified `.rs` files; `cargo test` 177 passed; `cargo clippy` at the same 4 pre-existing warnings, none new.

### Still not tested here

Unchanged from the original PR text: **nothing here proves a badge visibly changes in Finder.** That is slice #105, manual, with evidence captured without navigating away.

mbsanchez:
## CTO Review (re-review of bcb7f0f): CHANGES REQUESTED ❌ — 1 blocking

The two previous findings are properly fixed, and the fix for the first one was made the right way — moving the logic into `BadgeDiff` rather than only correcting the comparison. Verified independently: reintroducing the bare `hasPrefix` fails 2 checks, one of them exposing the pre-existing `"vil/secret.txt"` off-by-one. 26 checks passing, build clean, both architectures, Rust untouched.

But tracing the four original symptoms against the new code turns up a hole, and it is in the symptom the ticket lists **first**.

---

### 🔴 BLOCKING — `.cloudsc` placeholders still never get a badge

`badgedPaths` is only populated on the success path of `requestBadgeIdentifier(for:)`:

```swift
guard let tier = state?.paths[relative] else {
    if badgedPaths.contains(relative) { … }
    return                          // <- not recorded
}
FIFinderSyncController.default().setBadgeIdentifier(tier, for: item)
badgedPaths.insert(relative)        // <- only here
```

**Failure scenario**, which is symptom 1 from the ticket verbatim:

1. A `.cloudsc` placeholder appears in an open Finder window.
2. Finder draws it and calls `requestBadgeIdentifier(for:)` — per `FinderSync.h`, this fires *"whenever a new item becomes visible in the Finder"*.
3. The Rust side has not refreshed `overlay_state.json` yet, so `state.paths` has no entry. We return **without recording the path**.
4. Rust refreshes; the placeholder is now in the map as `cloud_only`.
5. `reloadState()` diffs and reports it as changed — then `pushable` **filters it out**, because it is not in `badgedPaths`.
6. Finder never asks again. **No badge, indefinitely.**

So the PR fixes symptoms 3 and 4 (tier changes on already-badged files) and leaves symptom 1 exactly as reported. Ironically the `pushable` filter added in response to the previous review is what drops it.

**The cause is a wrong set.** Apple's constraint is about items Finder *has not displayed*:

> Avoid adding badges to items that the Finder hasn't displayed yet.

An item Finder asked about **has** been displayed — it simply was not given a badge, because we had nothing to say at the time. "Already received a badge" is Apple's shorthand for "Finder is showing it", and taking it literally excludes exactly the case that matters here.

**Fix**: record every contained path Finder asks about, whether or not a badge was set, and rename the set to match what it means:

```swift
private var displayedPaths: Set<String> = []

override func requestBadgeIdentifier(for url: URL) {
    …
    displayedPaths.insert(relative)   // before the tier lookup, not after
    guard let tier = state?.paths[relative] else { … }
    …
}
```

Then `pushable(_:alreadyBadged:)` filters against `displayedPaths`. The parameter name should change too — it is a lie about its contents right now, and a comment that misdescribes a guard is what made the previous round blocking.

**This needs a test**, and it is expressible in `BadgeDiff`: a path that is displayed-but-unbadged must be pushable. That case does not exist in the current 26 — which is why the hole survived a green run. Add it to the `pushable` group.

**And it needs to reach slice #105's acceptance criteria.** *"A `.cloudsc` placeholder shows `cloud_only` as soon as it appears"* is already written there, so the manual validation would have caught this — but only after a build, an install, the `pluginkit` dance and a screen recording. Catching it here costs one test.

---

### 🟢 Confirmed fixed

- **Path containment** — `isContained` compares against `root + "/"`. `grep hasPrefix SyncExtension.swift` returns only the explanatory comment. Both callers use the tested helpers.
- **Transient read failure** — `resync(to:alreadyBadged:)` refreshes displayed badges when the snapshot was lost. Note it is filtered twice: `resync` already narrows to badged paths and `pushBadgeChanges` narrows again via `pushable`. Redundant but harmless and defensively fine — no change requested.
- **Both notes** from the previous round are written down.

---

Moving DBSYNC-84 back to In Progress.

mbsanchez:
## Review fix pushed — 86fd4da

Blocking finding fixed. It took two attempts and the failed one is worth recording.

### The fix

`badgedPaths` → `displayedPaths`, populated when Finder **asks**, not when a badge is set. Apple's operative words are *"items that the Finder hasn't displayed yet"* — an item Finder asked about has been displayed; it simply had no tier to show yet.

### The first attempt was indistinguishable from working

Moving the `insert` above the tier lookup and adding a `pushable` test for the displayed-but-unbadged case looked done. Reintroducing the bug on purpose:

```
=== insert moved back below the guard ===
BadgeDiff: all checks passed
BUILD SUCCEEDED
```

**27 green checks, clean build, bug fully present.** The test stated the contract but could not detect the defect, because the defect was *where a line sat* in `SyncExtension.swift`.

That is the **third defect in this PR living in the one file the check script cannot compile** — string-prefix containment, then the resync gap, then this. At three, it stops being bad luck.

### So the decision moved into `BadgeDiff`

```swift
struct RequestOutcome: Equatable {
    let displayedPath: String?     // non-nil even when there is no tier to set
    let badgeIdentifier: String?   // a tier, "" to clear, nil to do nothing
}
static func requestOutcome(for path: String, root: String, paths: [String: String]?) -> RequestOutcome
```

`requestBadgeIdentifier(for:)` is now six lines that apply the outcome and decide nothing:

```swift
let outcome = BadgeDiff.requestOutcome(for: item.path, root: root.path, paths: state?.paths)
if let displayed = outcome.displayedPath { displayedPaths.insert(displayed) }
if let identifier = outcome.badgeIdentifier {
    FIFinderSyncController.default().setBadgeIdentifier(identifier, for: item)
}
```

**Verified the same way, and this time it holds:**

```
=== bug reintroduced (record only when a tier exists) ===
FAIL an UNTRACKED file inside the root still counts as displayed
FAIL with no state at all, the path is still displayed
BadgeDiff: 2 check(s) FAILED
=== restored ===
BadgeDiff: all checks passed
```

### A suggestion for the reviewer, beyond this PR

Three defects, all in `SyncExtension.swift`, none in `BadgeDiff.swift`. The difference is not care — it is reachability. Worth considering as a standing rule for this file: **it applies decisions, it does not make them.** Anything with a branch belongs in `BadgeDiff`, where the check script can compile it.

### Validation

**32 checks passing**, up from 26. New group `requestOutcome` (5), two of them regression guards for exactly this.

```
** BUILD SUCCEEDED **   Principal class OK: DropboxSyncFinderSync
lipo -archs              ->  x86_64 arm64
nm | grep -ci BadgeDiff  ->  130
```

**Rust untouched**: 0 modified `.rs` files; `cargo test` 177 passed.

### Still not tested here

Unchanged: **nothing in this PR proves a badge visibly changes in Finder.** That remains slice #105 — manual, on the Mac, with evidence captured without navigating away.

mbsanchez:
## CTO Review (3rd pass, 86fd4da): CHANGES REQUESTED ❌ — 1 blocking, one line

All four reported symptoms now trace clean through the code, and the previous blocking finding is properly fixed — moving the decision into `BadgeDiff.requestOutcome` was the right call, and the mutation check confirms it holds where the first attempt did not.

There is one more instance of the same defect, and I should have caught it in the previous pass when the removal loop was written.

---

### 🔴 BLOCKING — `displayedPaths.remove(relative)` conflates displayed with badged, again

`SyncExtension.swift:188`, in the removal loop:

```swift
for relative in pushable.removed {
    guard let url = url(forRelative: relative, under: root) else { continue }
    controller.setBadgeIdentifier("", for: url)
    displayedPaths.remove(relative)          // <- this line
}
```

**Failure scenario** — a file that leaves the state map and comes back:

1. `a.txt` is tracked and displayed. `displayedPaths = {a.txt}`, badge `synced`.
2. `a.txt` leaves the state map. The badge is cleared correctly — **and the path is dropped from `displayedPaths`**.
3. `a.txt` returns to the state map as `synced`.
4. `changes` reports it. `pushable` filters it out: not in `displayedPaths`.
5. Finder never asks again — the item has been on screen the whole time.
6. **No badge, indefinitely.**

This is the same bug that was just rejected, through a third door. Clearing a badge does not mean Finder stopped displaying the item; the file is still on screen, it just has no status to show. `displayedPaths` shrinking on a badge clear is the "displayed == badged" conflation the last fix was supposed to remove.

**The set has no legitimate reason to shrink here.** The only real signal that Finder stopped displaying something is `endObservingDirectoryAtURL:`, and this extension does not implement it (`grep endObserving` → no matches). Absent that signal, an entry in `displayedPaths` should be treated as "Finder has seen this path", which stays true.

**Fix**: delete the line. That is the whole change.

**Test**: the `pushable` group needs the round trip — a path that was removed and returns must still be pushable. Expressible in `BadgeDiff` and currently absent, which is why 32 green checks did not catch it.

**Consequence**: `displayedPaths` then never shrinks at all. That was already flagged as acceptable in the first review pass — one entry per distinct path Finder has shown in the sync folder, a few MB at 50,000 files. The comment on the property should say so explicitly, since the growth is now unconditional rather than incidental.

---

### 🟢 Confirmed fixed, and traced end to end

All four symptoms from the ticket:

| Symptom | Path through the code | |
|---|---|---|
| 1. `.cloudsc` appears with no badge | request → `displayedPath` non-nil despite no tier → later diff pushes | ✅ |
| 2. Hydrating one makes others appear | each change now pushes on its own | ✅ |
| 3. Green check missing after hydration | tier change → in `displayedPaths` → pushed | ✅ |
| 4. Fixed by leaving and returning | no longer the only route | ✅ |

Also confirmed: containment via the tested helper, `resync` after a read failure, `updated_at` short-circuit, `lastUpdatedAt` cleared alongside `state`, no Rust, both architectures, 130 symbols linked.

---

### On the pattern, not the code

Four defects in this PR, **all four** in `SyncExtension.swift`, none in `BadgeDiff.swift`. This one is a single line in a file that now holds only three decisions — and it was still wrong.

The suggestion from the last pass stands and should probably become the rule: **that file applies decisions, it does not make them.** The removal loop deciding to mutate `displayedPaths` is a decision, and it belongs where it can be checked.

Moving DBSYNC-84 back to In Progress.

mbsanchez:
## Review fix pushed — 2a3634f

`displayedPaths.remove(relative)` deleted from the removal loop. That is the whole code change.

The rationale went onto the property, because the guarantee is now the **absence** of a line:

> **This set only ever grows, and that is deliberate.** Clearing a badge does not mean Finder stopped showing the item — the file is still on screen with no status to display. […] The only genuine "no longer displayed" signal is `endObservingDirectoryAtURL:`, which this extension does not implement.
>
> The cost of never shrinking is one string per distinct path Finder has shown inside the sync folder — a few MB at the 50,000 files DBSYNC-80 measured. Accepted.

### The requested test, and what it honestly does not do

The round trip is added — removed then tracked again, pushable both ways, 34 checks now. **But it does not guard this fix**, and the comment above it says so instead of implying otherwise:

> it pins the composition of `changes` and `pushable`. It cannot catch someone dropping the path from the displayed set inside `SyncExtension.swift` — that is a property of a line that no longer exists.

The alternative worth naming and rejecting: a `displayedAfter(_:applying:)` helper whose entire body would be `return displayed`. A pure function that does nothing, existing only so a check could assert on it. That would satisfy the letter of "make it testable" while trading a true comment for a fake test, and this ticket's Implementation Decisions already reject exactly that kind of abstraction.

### Validation

**34 checks passing**, up from 32.

```
** BUILD SUCCEEDED **   Principal class OK: DropboxSyncFinderSync
lipo -archs   ->  x86_64 arm64
```

**Rust untouched**: 0 modified `.rs` files.

### The pattern, for the record

Four defects in this PR. All four in `SyncExtension.swift`. None in `BadgeDiff.swift`.

Each fix has moved decisions out of that file, and this one removed the last state mutation from the push path — the removal loop now only calls `setBadgeIdentifier`. The reviewer's standing suggestion, that this file **applies decisions and does not make them**, is now true rather than aspirational.

That belongs on DBSYNC-72's Testing Decisions next to the existing "no Swift test target" constraint, rather than living in a PR thread. Flagged, not done — editing another ticket is outside this slice.

### Still not tested here

Unchanged through four rounds: **nothing in this PR proves a badge visibly changes in Finder.** That is slice #105 — manual, on the Mac, evidence captured without navigating away.

mbsanchez:
## CTO Review (4th pass, 2a3634f): APPROVED ✅

The line is gone, `displayedPaths` no longer shrinks anywhere, and the reasoning sits on the property. I asked for a test that guards it; the response was that a deleted line cannot be guarded and that the alternative would be a pure function whose body is `return displayed`. **That is the correct answer and I accept it.** A comment that is true beats a check that only looks like one — which is the lesson this ticket has been teaching in one form or another since the first round.

### Verified independently, not taken on trust

Re-ran the checks from scratch and re-broke the three load-bearing invariants:

```
first-load guard broken   -> 2 checks FAIL
containment broken        -> 4 checks FAIL
displayed-without-tier    -> 3 checks FAIL
restored                  -> all 34 checks pass
```

Every invariant this PR depends on fails loudly when removed. That is the property that was missing in round 2, when 27 green checks coexisted with a live bug.

```
** BUILD SUCCEEDED **   Principal class OK: DropboxSyncFinderSync
lipo -archs  ->  x86_64 arm64
```

**Scope**: 5 files, all under `FinderSyncExtension/`. No Rust, no IPC, no capabilities, no secrets, no stray debug output — the only `print(` calls are inside the check script, where they are the output.

**All four reported symptoms** trace clean through the code, individually walked in the third pass.

### One note, deliberately not a finding

`reloadState()` still branches between `resync` and `changes` on `previousPaths == nil && !displayedPaths.isEmpty`. That is a routing decision living in `SyncExtension.swift`, and the standing rule from this review says decisions belong in `BadgeDiff`.

It could fold into a single `changes(from:to:displayed:)`. **I am not requesting it.** The branch is correct, both of its outcomes are covered by checks, and asking for a refactor of correct code in a fourth round would be manufacturing a finding to justify another cycle. Worth doing next time this file is opened, not now.

### What this approval does NOT cover — read before merging

**Nothing in this PR demonstrates that a badge visibly changes in Finder.** Four review rounds have improved the code and none of them could touch that. The 34 checks cover the diff, containment, the request decision and the resync — none of them talks to Finder.

The three things that still need a human at a Mac, all in slice #105:

- Hydrating a file updates its badge **without leaving the directory**.
- A `.cloudsc` placeholder shows `cloud_only` as soon as it appears — symptom 1, which two of the four defects in this PR were specifically about.
- Idle CPU unchanged.

And the evidence rule stands: **captured without navigating away**, because navigating away is what makes a stale badge look correct.

Moving DBSYNC-84 to QA.

mbsanchez:
## QA Result: PASSED ✅

*(this comment was re-posted — the first version had its code spans mangled by shell substitution)*

The badge changes on screen. Proven with before/after captures taken by **window ID**, so the Finder window was never raised or touched: a navigation was physically impossible between the state change and the observation.

```
14:32:57  file created in the open sync folder
t+5s      state: out_of_sync   ->  RED badge
t+10s     state: synced        ->  GREEN badge
```

Same window, same scroll, same expanded folders; the only difference between the two images is the badge on that file. Before this PR the red badge would have persisted until the user left the directory and came back.

Idle CPU: 0.0–0.1% before and after, 5 samples each.

**Two ACs left unticked rather than assumed**: a `.cloudsc` badged at the instant it first appears (none arrived during the session), and a badge clearing on removal (deleting the file removes the row, which forces a redraw and destroys the measurement). Both have unit coverage; neither has a live observation.

**Bonus**: the `out_of_sync` badge is on screen for the first time in this project's history — DBSYNC-72 Slice 3 recorded it as never observed and it stayed that way through every round since. `syncing` is still unseen; that transition is too fast for a 5-second sampling interval.

Full evidence, plus the false start that made this look like a regression at first, on issue #105.

Ready for PO approval.

Issue #109, opened 2026-08-21
DBSYNC-86 Slice 3: Live validation of all three banner states


mbsanchez:
Vertical slice 3 of **DBSYNC-86**. Type: **manual / HITL**.

* Scope dir(s): none — validation only
* Blocked by: Slice 2 ([#108](https://github.com/mbsanchez/DropboxSync/issues/108))
* User stories: 1, 2, 3

## Acceptance criteria

- [ ] With the extension **disabled**, the banner appears in the flyout.
- [ ] The button opens the Finder extensions pane in System Settings.
- [ ] After enabling, the banner switches to the **restart hint**.
- [ ] After restarting Finder, the banner disappears and badges render.
- [ ] A screenshot of **each** of those states — three images, not one.
- [ ] The app does not crash while any of this runs. Exercised, not assumed: this is the project's first Objective-C interop.

## Procedure

The state is toggled from **System Settings → Privacy & Security → Extensions → Finder**. Note that the app must be running and focused for the state to refresh — the check runs on window focus, so switch away to System Settings and back.

## The trap that already produced one false failure

**After enabling the extension, badges do not appear until Finder restarts.**

On 2026-08-21 this cost about half an hour: the checkbox was ticked, nothing happened, and it looked like a code regression. It was not — the extension was being hosted by `openAndSavePanelService` rather than by Finder, and only `killall Finder` moved it.

A validation run that stops at *"I enabled it and saw nothing"* will report a false failure. That is precisely the state this slice exists to make visible, so seeing the restart hint at that moment is a **pass**, not a problem.

## Verifying the host, if a state looks wrong

```
pluginkit -mv -p com.apple.FinderSync | grep mobsanchez
```

The `+`/`-` prefix is the enablement flag that matters — the plain `pluginkit -m` listing shows a **different** flag and reading it produced a wrong conclusion once already.

```
log show --last 2m --style compact \
  --predicate 'process == "DropboxSyncFinderSync" AND eventMessage CONTAINS "Identity resolved"'
```

Look for `osservice<com.apple.Finder>` as the host. Anything else means Finder has not loaded it.

Jira: DBSYNC-86

mbsanchez:
## Implementation Plan

**Scope dir(s)**: none — validation only
**Implementing agent(s)**: the maintainer, on the Mac. No agent.
**Estimated complexity**: low to run. The risk is not difficulty, it is stopping too early.
[...]

Issue #108, opened 2026-08-21
DBSYNC-86 Slice 2: One click to the setting, and the Finder-restart hint


mbsanchez:
Vertical slice 2 of **DBSYNC-86**. Type: **AFK**.

* Scope dir(s): `apps/desktop/src-tauri`, `apps/desktop/src`
* Blocked by: Slice 1 ([#107](https://github.com/mbsanchez/DropboxSync/issues/107))
* User stories: 2, 3

## Acceptance criteria

- [ ] A command calls `FIFinderSyncController.showExtensionManagementInterface`, `cfg`-gated to macOS.
- [ ] The banner from Slice 1 gets a button that invokes it.
- [ ] After the state flips **disabled → enabled**, the banner switches to a hint that Finder may need restarting, with a button that does it.
- [ ] **Restarting Finder is never automatic.** It closes the user's open windows; the maintainer chose explicitly that the app asks rather than acts.
- [ ] No `pluginkit` shelled out from the app, for anything.
- [ ] `cargo test`, `cargo clippy --all-targets`, `npm run build` clean.

## Why the restart hint is a heuristic, and why that is the right call

On 2026-08-21 the maintainer ticked the checkbox and **badges still did not appear**. The extension was hosted by `openAndSavePanelService`, not by Finder. Only `killall Finder` fixed it. Without a hint, a user does exactly what we did: concludes it did not work.

But **there is no supported way to ask whether Finder has loaded the extension.** The only signal found was a `runningboard` log line naming the host process — undocumented, and not something an app should parse.

So the hint fires on the **disabled → enabled transition**, which the app already observes on window focus. It is a heuristic: it will occasionally suggest a restart that was not needed. That is a far cheaper failure than staying silent while the user believes the feature is broken.

**The rejected alternative**, recorded so it is not re-proposed: have the extension write a heartbeat file that the host polls. That invents an entire IPC channel, adds a file to keep clean, and creates a new "stale heartbeat" failure mode — all to avoid one hint that costs nothing when wrong. Revisit only if the heuristic actually misleads someone.

Jira: DBSYNC-86

mbsanchez:
## Implementation Plan

**Scope dir(s)**: `apps/desktop/src-tauri`, `apps/desktop/src`
**Implementing agent(s)**: rust-pro (command) then frontend (buttons)
**Estimated complexity**: low — the FFI groundwork lands in Slice 1; this adds one more selector and two buttons.
[...]

Issue #107, opened 2026-08-21
DBSYNC-86 Slice 1: The app knows the extension is off, and says so


mbsanchez:
Vertical slice 1 of **DBSYNC-86**. Type: **AFK**.

* Scope dir(s): `apps/desktop/src-tauri`, `apps/desktop/src`
* Blocked by: none
* User stories: 1, 4

## Why this is worth doing on its own

It delivers the whole point of the ticket by itself: the user finds out that badges are off rather than concluding the app is broken. The button in Slice 2 is convenience on top.

## Acceptance criteria

- [ ] Rust reads `FIFinderSyncController.isExtensionEnabled` via `objc2`, `cfg`-gated to macOS.
- [ ] **Non-macOS returns "not applicable", never "disabled".** Windows has no Finder; a Windows user must never see this banner. Three states, not a boolean.
- [ ] The state rides in the existing startup-requirements shape, not a new parallel channel.
- [ ] The flyout shows a banner when disabled, reusing the existing `flyout-banner` class and its `role="alert"`.
- [ ] Refresh happens on window focus — `useStartupRequirements` already does this. Do not add a second mechanism.
- [ ] Unit test on the three-state mapping, following the `tray_icon_for_label` precedent from DBSYNC-78.
- [ ] `cargo test`, `cargo clippy --all-targets`, and `npm run build` all clean.

## The actual risk is the FFI, not the UI

**This is the project's first Objective-C interop.** `objc2` 0.6.4 is already in `Cargo.lock` transitively via Tauri, so no new dependency — but nothing in `src/` currently calls into Objective-C, which means this introduces a new category of failure into the app process.

Requirements that follow from that:

- [ ] The call must not panic or crash the app if it fails. A missing class or a nil return is a "cannot determine" answer, not a crash.
- [ ] The `unsafe` block carries a SAFETY comment saying what is being assumed.
- [ ] `FIFinderSyncController` lives in the **FinderSync** framework, which must be linked. Confirm what happens on a machine where that framework is unavailable rather than assuming it always is.

## Existing patterns to reuse rather than reinvent

- `apps/desktop/src/hooks/useStartupRequirements.ts` — answers "is this configured?", refreshes on mount **and on window focus**. Apple's own header comment prescribes exactly that timing: *"When the application becomes active again, it can check whether the user has enabled the extension."*
- `flyout-banner` in `FlyoutApp.tsx` — the mass-delete banner already establishes the look and the `role="alert"` semantics.
[...]

PR #110, opened 2026-08-21
feat(#DBSYNC-86): tell the user when the Finder extension is switched off

mbsanchez:
## Summary

Badges vanish silently when macOS disables the Finder Sync extension, and the app says **nothing**. Every visible signal reads as "the feature is broken" rather than "a switch is off". It cost about half an hour during DBSYNC-84's QA, where it looked like a code regression until the checkbox was found.

The app now reads `FIFinderSyncController.isExtensionEnabled`, warns in the flyout when it is off, and offers a button that opens the right settings pane.

Closes slices [#107](https://github.com/mbsanchez/DropboxSync/issues/107) and [#108](https://github.com/mbsanchez/DropboxSync/issues/108). [#109](https://github.com/mbsanchez/DropboxSync/issues/109) — live validation — is deliberately **not** covered here.

## The ticket's original ask was impossible, and the scope changed

It asked the app to **enable** its own extension. `FinderSync.h` ships `isExtensionEnabled` and `showExtensionManagementInterface` and deliberately **no setter**; the header's own comment prescribes the flow (check when the app becomes active, show the management UI). `pluginkit -e use` was measured on 2026-08-21 and does not work either.

Detect-and-ask is what Apple supports, and — the maintainer's own point — what Dropbox actually does. It does not enable the extension; it notices and asks.

## Design notes for review

- **Three states, never a `bool`.** `NotApplicable` is not a synonym for `Disabled`: a Windows user has no Finder extension to be warned about, and only `Disabled` renders the banner. Collapsing them is the mistake the enum exists to prevent.
- **First Objective-C interop in this project.** `AnyClass::get` returns `Option`, so a missing class is a *value* rather than a crash — that is how the "must not panic" requirement is met, without `catch_unwind`. The single `unsafe` block carries a SAFETY comment.
- **`showExtensionManagementInterface` runs inside `run_on_main_thread`.** It presents AppKit UI and Tauri commands carry no main-thread guarantee; getting this wrong produces an intermittent hang rather than a clean failure.
- **`objc2` had to be declared** even though it is already in `Cargo.lock` transitively — a transitive crate cannot be imported. Added under the existing `[target.'cfg(...)'.dependencies]` pattern.
- **The restart hint is a heuristic, and says so in the code.** It fires on the disabled → enabled transition, because there is no supported way to ask whether Finder has reloaded the plug-in. It will occasionally suggest a restart that was not needed — one wasted click, against a user who ticks the box, sees nothing, and concludes it is broken.
- **`killall Finder` is not the thing the ACs forbid.** They forbid `pluginkit`, which depends on undocumented behaviour and was measured not to work. `killall Finder` is documented, reversible, and something users do by hand. Never automatic — it closes their open windows.
- **No new hook, no new component, no new dependency.** `useStartupRequirements` already refreshes on window focus, which is exactly when Apple says to re-check; `flyout-banner` already exists.

## Stack

`desktop-tauri` — Rust FFI + React.

## Jira Ticket

DBSYNC-86

## Test Plan

- [x] `cargo test` — **179 passed**, up from 177.
- [x] `cargo clippy --all-targets` — the same 4 pre-existing warnings, none new. No complaint about the `unsafe` block.
- [x] `npm run build` — clean, typecheck included.
- [x] **The FFI genuinely reaches Apple's class.** A temporary probe in the test harness printed `Disabled`, not `NotApplicable` — the value a missing class returns. So the framework linked and the message send returned a real `BOOL`. Probe removed; `grep PROBE` returns 0.
  This mattered because **the tests pass either way**: had the framework not linked, everything would still be green and the feature would do nothing.
- [x] **Verified with the extension ENABLED: no banner.** Screenshot taken of the flyout.
  This is the check that matters. An always-`false` reader is **indistinguishable from a correct one while the extension is off**, and would nag every user who has it on. Testing only the disabled case would have proven nothing — the same shape as DBSYNC-72's sandbox ADR, which verified the branch that could not fail.

### Not tested here

**No button has been clicked.** Opening the settings pane and restarting Finder belong to #109, along with the disabled banner and the transition. In particular, **an incorrect main-thread dispatch would not show up in a build** — it surfaces as an intermittent hang at click time.

## A correction to the ticket

The Jira Problem Statement claims, as a hard rule, that *every* reinstall registers a new plug-in instance that starts **disabled**. Installing this build produced a new instance (`597A06E8…`) that came up **enabled**.

The accurate version: a reinstall registers a new instance and the enablement is **sometimes** lost — observed twice on 2026-08-21, not on the third. The mechanism is not understood.

This does not weaken the ticket. An intermittent, silent loss of badges is arguably worse than a reliable one, because it is harder to attribute. But the description overstates it, and a correction is recorded on the ticket rather than left standing.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/Cargo.lock
+++ b/apps/desktop/src-tauri/Cargo.lock
@@ -947,6 +947,7 @@ dependencies = [
  "chrono",
  "keyring",
  "notify-debouncer-mini",
+ "objc2",
  "rand 0.8.5",
  "reqwest 0.12.28",
  "rfd",
--- a/apps/desktop/src-tauri/Cargo.toml
+++ b/apps/desktop/src-tauri/Cargo.toml
@@ -41,6 +41,11 @@ zeroize = "1.9.0"
 [dev-dependencies]
 tempfile = "3"
 
+[target.'cfg(target_os = "macos")'.dependencies]
+# Already in Cargo.lock transitively via Tauri, but a transitive crate cannot be imported —
+# it has to be declared. Used only by `finder_extension.rs` (DBSYNC-86).
+objc2 = "0.6"
+
 [target.'cfg(windows)'.dependencies]
 winreg = "0.55"
 windows-sys = { version = "0.59", features = [
--- a/apps/desktop/src-tauri/build.rs
+++ b/apps/desktop/src-tauri/build.rs
@@ -1,3 +1,9 @@
 fn main() {
+    // `finder_extension.rs` sends a message to `FIFinderSyncController`, which lives in
+    // FinderSync.framework. Without this the class is simply not registered at runtime and the
+    // lookup returns None — a silent "not applicable" rather than a link error (DBSYNC-86).
+    #[cfg(target_os = "macos")]
+    println!("cargo:rustc-link-lib=framework=FinderSync");
+
     tauri_build::build()
 }
--- a/apps/desktop/src-tauri/src/commands.rs
+++ b/apps/desktop/src-tauri/src/commands.rs
@@ -190,6 +190,7 @@ pub(crate) fn compute_startup_requirements(
                 auth_ok: true,
                 sync_folder_ok,
                 sync_folder,
+                finder_extension: crate::finder_extension::finder_extension_state(),
             });
         }
     }
@@ -208,6 +209,7 @@ pub(crate) fn compute_startup_requirements(
         auth_ok,
         sync_folder_ok,
         sync_folder,
+        finder_extension: crate::finder_extension::finder_extension_state(),
     })
 }
 
--- a/apps/desktop/src-tauri/src/finder_extension.rs
+++ b/apps/desktop/src-tauri/src/finder_extension.rs
@@ -1,0 +1,170 @@
+//! Whether the user has switched our Finder Sync extension on (DBSYNC-86).
+//!
+//! This is the only Objective-C interop in the project, and it exists because there is no
+//! Rust binding for `FinderSync.framework`. It is kept in its own module so the single
+//! `unsafe` block has a name and a home rather than sitting inside `commands.rs`.
+//!
+//! **Why the app needs to ask at all.** macOS registers a *new* plug-in instance on every
+//! reinstall and the new one starts **disabled** — measured on 2026-08-21, two installs two
+//! minutes apart produced two UUIDs, the second one off. So an ordinary app update silently
+//! switches badges off, and without this the app has no way to say so.
+//!
+//! **Why it cannot enable it.** `FinderSync.h` ships `isExtensionEnabled` and
+//! `showExtensionManagementInterface` and deliberately no setter. The header's own comment
+//! prescribes the flow: check when the app becomes active, and show the management UI.
+
+use serde::Serialize;
+
+/// Three states, never a `bool`.
+///
+/// `NotApplicable` is not a synonym for `Disabled`: Windows has no Finder, and a Windows user
+/// must never be shown a banner about a Finder extension. Collapsing these two is the mistake
+/// this type exists to make impossible.
+#[derive(Serialize, Clone, Copy, Debug, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) enum FinderExtensionState {
+    /// The user has it switched on. Nothing to say.
+    Enabled,
+    /// Registered but switched off — badges will not appear. This is the one that warns.
+    Disabled,
+    /// Not macOS, or the state could not be determined. Say nothing.
+    NotApplicable,
+}
+
+/// Reads `FIFinderSyncController.isExtensionEnabled`.
+///
+/// **The answer is scoped to the calling bundle.** Measured during planning: a standalone
+/// process gets `false` while the extension is genuinely enabled, because the property reports
+/// on the extension inside the *caller's* app bundle. So this returns a meaningful value only
+/// from inside `DropboxSyncDesktop.app` — and, correspondingly, it cannot be verified by any
+/// test or spike outside it. The check that matters is running the app with the extension ON
+/// and confirming this says `Enabled`; an implementation stuck at `Disabled` looks perfectly
+/// correct as long as you only ever test with it switched off.
+pub(crate) fn finder_extension_state() -> FinderExtensionState {
+    #[cfg(target_os = "macos")]
+    {
+        use objc2::runtime::AnyClass;
+        use objc2::msg_send;
+
+        // A missing class is an answer, not a crash: if `FinderSync.framework` is not loaded
+        // there is nothing to report and `NotApplicable` is the honest value. This is why the
+        // "must not panic" requirement needs no `catch_unwind` — `get` already returns Option.
+        let Some(cls) = AnyClass::get(c"FIFinderSyncController") else {
+            return FinderExtensionState::NotApplicable;
+        };
+
+        // SAFETY: `FIFinderSyncController` is an Apple class from FinderSync.framework, and
+        // `isExtensionEnabled` is a documented class property (macOS 10.14+) taking no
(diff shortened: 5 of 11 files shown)


mbsanchez:
## CTO Review: APPROVED ✅ — with one gap closed during the review rather than filed

The scope change is correct and well-evidenced: the SDK header settles that no enable API exists, and `pluginkit -e use` was measured not to work. Detect-and-ask is what Apple prescribes.

### The check that mattered, verified independently

The reader was confirmed with the extension **enabled** and no banner shown. That is the right way round: an always-`false` implementation is **indistinguishable from a correct one while the extension is off** and would nag every user who has it on. Testing only the disabled case would have proven nothing — the shape of DBSYNC-72's sandbox ADR failure.

The probe evidence is also the right kind. `Disabled` rather than `NotApplicable` from the test harness proves the framework linked and the message send returned a real `BOOL` — and it matters precisely because **the tests pass either way**: an unlinked framework would leave everything green and the feature inert.

### The gap I found, and closed

`non_macos_is_not_applicable_never_disabled` is `cfg`-gated to non-macOS, so it **never runs on this machine**. Two of the three tests execute here. The claim "a Windows user never sees a Finder banner" was therefore asserted by a test that cannot fail.

The precedent the plan told this slice to follow — `tray_icon_for_label` from DBSYNC-78 — has exactly the same shape: `other_platforms_keep_their_colour_icons` is gated the same way and does not run either. So this is a property of the established pattern, not a lapse in this PR, and rejecting for it would be moving the goalposts after the fact.

**So I ran it instead of filing it**, using the `cfg`-pivot technique DBSYNC-81 established for exactly this:

```
=== cfg predicates swapped ===
test finder_extension::tests::non_macos_is_not_applicable_never_disabled ... ok
test result: ok. 2 passed; 0 failed
=== reverted, grep __PIVOT__ -> 0 ===
```

The non-macOS arm compiles and returns `NotApplicable`. The Windows guarantee is now **verified**, not merely written down. No code change needed.

### Verified beyond the diff

- **No capability change is needed and none was made.** `capabilities/default.json` lists only plugin permissions; commands registered through `generate_handler!` are app commands and are not gated there. The ticket's claim holds.
- **`run_on_main_thread` is the right call** for `showExtensionManagementInterface`, and the reason is in the doc comment rather than left implicit.
- **No `unwrap`, no `expect`, no leftover debug** in the new module. The temporary probe is gone — `grep PROBE` returns 0.
- **Scope clean**: 11 files, all under the two declared directories.
- `cargo test` 179, `cargo clippy --all-targets` at the same 4 pre-existing warnings, `npm run build` clean.

### Notes, not objections

- **Both button handlers swallow errors** — `void invoke(...)`. For "open a settings pane" and "restart Finder" that is defensible: neither failure is worth an error dialog, and both are retryable by the user. Worth knowing it is a choice rather than an oversight.
- **The banner lives only in the flyout**, not in the setup window. A brand-new user during onboarding would not see it. Out of this slice's scope and arguably correct — but if a fresh install ever reports "no badges", that is the gap.
- **The restart hint is a heuristic** and says so in the code, so it will not be "fixed" later by someone reading it as a bug.

### What this approval does NOT cover

**No button has been clicked.** In particular an incorrect main-thread dispatch surfaces as an **intermittent hang at click time** and would never appear in a build or a test run. That, the disabled banner, and the transition are all #109 — and it needs the checkbox toggled by hand.

Moving DBSYNC-86 to QA.

mbsanchez:
## QA Result: PASSED ✅ — with one wording defect found

All four UI states observed live on the dev Mac. Full evidence and screenshots on issue #109.

```
A  extension ENABLED   -> no banner            (checked first, deliberately)
B  extension DISABLED  -> warning banner
C  "Open settings"     -> System Settings opens, app still alive (pid 74742)
D  re-enabled          -> restart hint; Restart Finder: 16323 -> 2473;
                          banner gone, badges render again
```

**State A first is the point.** An always-`false` reader is indistinguishable from a correct one while the extension is off, and would nag every user who has it switched on. Checking only the disabled case would have proven nothing.

**State C clears the PR's real risk.** `showExtensionManagementInterface` presents AppKit UI, and a wrong main-thread dispatch surfaces as an **intermittent hang at click time** — invisible to every build, test and clippy run in the PR. Four clicks, no hang, app alive throughout.

### 🟡 Defect found — not blocking, one string

The banner tells the user:

> Enable it under System Settings → Privacy & Security → **Extensions → Finder**.

On macOS 26 the button actually lands on **General → Login Items & Extensions**, in a sheet titled **"File Providers"** — our extension sits there next to Dropbox, OneDrive and odrive. The path in the text does not exist, and the sheet is not called "Finder".

The button works, so the feature is sound; the text is the fallback for anyone who does not click it, and it would send them hunting. Worth correcting, and cheap.

### 🔵 Finding worth keeping

**`pluginkit -e ignore` disables the extension**, even though `pluginkit -e use` cannot enable it. macOS permits switching a Finder Sync extension off from the CLI but not on.

That asymmetry is what let this QA run almost unattended — the disabled state is reachable without a human. Only the re-enable needed the checkbox.

Ready for PO approval.

mbsanchez:
## Copy defect fixed — c0db591

QA found this and I fixed it before merging rather than filing it. Shipping instructions that send the user to a pane that does not exist would contradict the ticket's own purpose.

**Before:**
> Enable it under System Settings → Privacy & Security → **Extensions → Finder**.

**After:**
> Use the button below, or find DropboxSyncDesktop under System Settings → **General → Login Items & Extensions**.

Two changes of substance, not just wording:

- **The path is now correct for macOS 26.** The old one was wrong twice: the pane is under General, not Privacy & Security, and the sheet is titled "File Providers", not "Finder".
- **The button leads, the text follows.** Apple has already relocated this pane once — over-specifying a path is what produced the defect in the first place. Naming only the outer pane keeps it true for longer, and the button is the route that cannot go stale.

### Verified on screen

Rebuilt, installed, extension disabled, flyout opened: the corrected copy renders as intended. Screenshot taken.

`npm run build` clean. No Rust touched by this commit.

### Note on process

QA had already passed when this landed, so **this commit has not been through a review round**. It is one string plus a comment, and the change is visible in the diff — but it is a change after approval, and worth a second pair of eyes on the copy before merge rather than assuming a text edit is risk-free.

mbsanchez:
## CTO Review (2nd pass, c0db591): APPROVED ✅

Short pass, because the change is short. Reviewed anyway rather than waving through a copy edit made after QA had already approved — the rule that nothing merges unreviewed is the one that found four real defects on this project today.

### The change

One file, 9 insertions, 2 deletions. **Zero Rust files.** Pure JSX text plus the comment explaining it. No behaviour, no state, no new call.

### It is correct, and correct for the right reason

The new copy names **DropboxSyncDesktop** — which is exactly the label the System Settings sheet shows for that row, verified in the screenshot from QA. A user reading the banner and then scanning the pane sees the same string. The old text named neither the right pane nor the right sheet.

More importantly the change fixes the *class* of defect, not just the instance: it puts the button first and names only the outer pane. Apple has already relocated this once between releases, and the previous text was wrong precisely because it over-specified. A shorter path stays true longer.

The comment records why it is deliberately vague, so the next person does not "improve" it by adding the specificity back.

### Verified

- HTML entities properly escaped (`&rarr;`, `&amp;`).
- `npm run build` clean, typecheck included.
- `cargo clippy --all-targets` — still the same 4 pre-existing warnings; nothing here could have changed that, and it confirms the commit is as narrow as it claims.
- Rendered on screen with the extension disabled — the copy appears as written.

### On the process point raised in the previous comment

Flagging "this went in after approval" was the right call, and the fix for it is this review, not a note. Approved on its merits.

QA does not need re-running: the four UI states were validated against the same mechanism, and no mechanism changed. The copy itself was verified on screen after the edit.

Moving DBSYNC-86 to QA — where it already was — and it is ready for PO approval.

Issue #113, opened 2026-08-21
DBSYNC-85 Slice 3: Fix the inversion, using the measurement


mbsanchez:
Vertical slice 3 of **DBSYNC-85**. Type: **AFK**.

* Scope dir(s): `apps/desktop/src-tauri`
* Blocked by: Slice 2 ([#112](https://github.com/mbsanchez/DropboxSync/issues/112)) — **hard block**
* User stories: 1

## Acceptance criteria

- [ ] Convert the tray position into the space `monitor_from_point` expects, using what Slice 2 measured — not what anyone expected it to measure.
- [ ] A test reproduces the maintainer's reported symptom against their actual arrangement, and shows it fixed. The layout is test data now, so this is expressible.
- [ ] The symmetric case both ways: click on screen 1 → flyout on screen 1; click on screen 2 → flyout on screen 2.
- [ ] Verified live on two displays, both directions, with screenshots.
- [ ] Windows untouched — provable by reading the diff.

## Why this is blocked rather than merely sequenced

The fix is likely a single arithmetic change. That is precisely what makes it dangerous: it is small enough to feel obvious, and a wrong sign will look correct on one of the two arrangements.

The block is not ceremony. It is the difference between a fix and a coin flip that happened to land.

Jira: DBSYNC-85

mbsanchez:
## Nothing left to fix — and the reason is not what the ticket assumed

Measurement is in #112. Summary: the tray reports in the **same top-left-origin, physical-pixel
global space** the monitor lookup expects. There is no Cocoa/winit mismatch, so **the conversion
this slice was created to write does not exist**.

The maintainer confirmed on 2026-08-22, with both displays connected, that the flyout now opens on
the clicked screen in both directions. The log backs that up: four clicks, on both displays.

## So what was the original symmetric inversion?

Worth pinning down, because "it works now" without a cause is how a bug comes back.

**It was not the Y anchor.** That was the other half of the diagnosis, and the arithmetic rules it
out. Taking the real geometry from #112 and the pre-fix code (`y = rect_pos.y - win_h - gap`,
clamped to the chosen monitor):

| Click | Chosen monitor | Raw `y` | Clamp range | Result |
|---|---|---|---|---|
| Built-in, `y≈0` | `(0,0) 3360x2100` | `0 - 1200 - 16 = -1216` | `0 .. 900` | `0` — **on the built-in** |
[...]

Issue #112, opened 2026-08-21
DBSYNC-85 Slice 2: Measure what coordinate space the tray reports in


mbsanchez:
Vertical slice 2 of **DBSYNC-85**. Type: **manual / HITL — requires the second display**.

* Scope dir(s): none — measurement only
* Blocked by: Slice 1 ([#111](https://github.com/mbsanchez/DropboxSync/issues/111)), which ships the instrumentation
* User stories: 1

## The one thing no test can answer

Everything else about flyout placement became testable in Slice 1 by making the monitor list an input. This did not, and cannot: **what coordinate space `tray-icon` reports a click in** is a property of the running system, not of our code.

Hence its own slice, its own AC, and a hard block on the fix that depends on it.

## Acceptance criteria

- [ ] Both displays connected. Record the arrangement — **stacked or side-by-side, and which is primary**. The interpretation is meaningless without it.
- [ ] Click the tray icon **on each screen** and capture the logged line from Slice 1's instrumentation.
- [ ] **Name the coordinate space in the ticket, with the numbers.** Is the reported Y measured from the top or the bottom? Of the primary screen, or of the screen the icon is on?
- [ ] State plainly whether the Cocoa-coordinates hypothesis is confirmed or refuted. Either is a result.

## How to read the log

Compare the reported `position` against the monitor origins printed alongside it. If the Y lands inside a monitor the icon is *not* on, that names the flip. If X is the one that misplaces it, the hypothesis is wrong and the fix is different.

The instrumentation prints all of it together on purpose — so the answer is visible in one line rather than reconstructed from three.

## Why this is worth a whole slice

The ticket already records four premises on this project that were falsified by running a command, and one ADR that failed because it reasoned instead of measuring. This is the same fork: guess, or spend one click.

Jira: DBSYNC-85

mbsanchez:
## Measurement taken — 2026-08-22, two displays connected

Captured from the Slice 1 instrumentation (`target: dbsync85`) in
`~/Library/Application Support/DropboxSyncDesktop/dropbox-sync.2026-08-22.log`.
Four clicks, on **both** displays.

### Display arrangement

| Display | Origin (physical px) | Size (physical px) |
|---|---|---|
| Built-in (primary) | `(0, 0)` | `3360 x 2100` |
| External | `(-6016, -1076)` | `6016 x 3384` |

The external display sits to the **left of** and slightly **above** the primary — hence its negative origin on both axes.

### The clicks

```
[...]

Issue #111, opened 2026-08-21
DBSYNC-85 Slice 1: Make the flyout geometry pure, testable, and right-way-up


mbsanchez:
Vertical slice 1 of **DBSYNC-85**. Type: **AFK**.

* Scope dir(s): `apps/desktop/src-tauri`
* Blocked by: none
* User stories: 2, 3

## Why this slice exists in this shape

**Only one display is connected to the development machine**, so neither the measurement nor the live fix can happen here. Rather than block on hardware, this slice does everything that is defensible without it — and changes the design so that most of it *becomes* defensible.

The key move: the placement math takes the **monitor list as an input** rather than looking it up. A monitor layout is four numbers; once they are a parameter, the two-display arrangements nobody here owns become ordinary test data.

## Acceptance criteria

- [ ] Placement math extracted into a pure function — tray rect + monitor list + window size → chosen monitor and final position. No `AppHandle`, no I/O, no globals.
- [ ] **Y anchor corrected for macOS**: the flyout goes *below* the tray rect. Windows keeps its bottom-taskbar behaviour, `cfg`-split so neither platform can drift.
- [ ] Tests over layouts this machine does not have: side-by-side, vertically stacked, differing scale factors, a click originating on each screen.
- [ ] Instrumentation logs the raw tray `position`, the tray `rect`, and every monitor's origin/size — once per click, not per frame.
- [ ] Single-display behaviour does not regress.
- [ ] `cargo test` and `cargo clippy --all-targets` clean; no new warnings.

## Explicitly NOT in this slice

**Do not fix the inversion.** Do not add a `-`, do not flip a sign, do not "correct" the Y based on the Cocoa-coordinates hypothesis.

That hypothesis is unmeasured. A coordinate flip has roughly a coin-flip chance of *appearing* correct on one arrangement while being wrong on another — which is the worst outcome available, because it would look like a fix and ship as one.

The measurement is #112 and the fix is #113. Slice 1 makes that measurement a single click instead of a debugging session; that is its contribution to the inversion, and it is enough.

## The Y anchor, which is a separate and certain defect

```rust
let mut y = rect_pos.y - win_h - gap;   // ABOVE the tray rect
```

This is a bottom-taskbar assumption inherited from Windows. On macOS the menu bar is at the top. On a single display the subsequent clamp forces `y` up to `min_y`, so the result currently looks right **by accident** — which is exactly why it has survived.

No measurement is needed to confirm it. It is arithmetic.

[...]

PR #114, opened 2026-08-21
fix(#DBSYNC-85): make the flyout geometry pure and anchor it below the menu bar

mbsanchez:
## Summary

Two defects in `position_flyout`, and this PR fixes one of them on purpose.

**Fixed — the Y anchor.** `y = rect_pos.y - win_h - gap` put the flyout *above* the tray rect: a bottom-taskbar assumption inherited from Windows. On macOS the menu bar is at the top, and the clamp then dragged the result back to `min_y`, which is why it has looked correct **by accident** on a single display. Windows keeps its original anchor, `cfg`-split.

**Not fixed — the multi-monitor inversion.** Deliberately. See below.

Closes slice [#111](https://github.com/mbsanchez/DropboxSync/issues/111). [#112](https://github.com/mbsanchez/DropboxSync/issues/112) (measure) and [#113](https://github.com/mbsanchez/DropboxSync/issues/113) (fix) remain open.

## The constraint that shaped the design

**This machine has one display.** Checked, not assumed:

```
screen 0  origin=(0, 0)  size=1440x900  scale=2.0  Built-in Retina Display
```

So the reported bug cannot be reproduced here and a fix could not be validated. Rather than block, the placement math now takes the **monitor list as an input** instead of looking it up through `AppHandle`.

That is the whole point: a layout is four numbers. Once they are a parameter, the two-display arrangements nobody here owns — stacked, side-by-side, mixed scale factors — become ordinary test data. **Two-monitor behaviour is now testable on a one-monitor machine.**

## Why the inversion is untouched

The theory is that `tray-icon` reports clicks in Cocoa coordinates (origin bottom-left) while `monitor_from_point` expects winit's top-left space. It is **unmeasured**.

A coordinate flip has roughly a coin-flip chance of *appearing* correct on one arrangement while being wrong on another — the worst outcome available, because it would look like a fix and ship as one. This project already carries four premises falsified by running a command, and an ADR that failed because it reasoned instead of measuring.

So instead this PR ships the instrumentation: one log line per click carrying the raw tray position, the tray rect and every display's origin/size **together**, so the space can be named from a single line rather than reconstructed from three. That turns #112 into one click instead of a debugging session.

## Stack

`desktop-tauri` — Rust only. No frontend, no IPC, no capability change, no new dependency.

## Jira Ticket

DBSYNC-85

## Test Plan

- [x] **10 new tests**, 189 total (was 179). They cover layouts this machine does not have: side-by-side, stacked with the secondary above the primary, a non-Retina second display, clamping onto the correct screen, and a window larger than its display.
- [x] **The tests can fail** — verified by mutation, not assumed:
  ```
  restore the Windows anchor on macOS      -> macos_hangs_the_flyout_below_the_menu_bar FAILED
  make monitor lookup always pick [0]      -> 5 tests FAILED
  restored                                 -> 10 passed
  ```
- [x] **The Windows arm was compiled and executed**, via the `cfg`-pivot technique DBSYNC-81 established: `other_platforms_keep_the_bottom_taskbar_anchor ... ok`. Pivot reverted. Without this the Windows guarantee would rest on a test that never runs.
- [x] `cargo clippy --all-targets` — same 4 pre-existing warnings, none new.
- [x] Release bundle builds and installs.

### Not verified — read before merging

**The instrumentation has never been seen to emit, and single-display placement has not been re-checked visually.** The machine's display went to sleep partway through validation, so the click never reached the tray and `screencapture` returned *"could not create image from rect"*.

What that leaves unverified:
- That the `dbsync85` log line actually appears on a real click. It compiles and the global filter is `info`, so it *should* — but "should" is what this project keeps getting wrong.
- That the corrected Y anchor looks right on screen. The arithmetic is tested; the pixels are not.

Neither blocks the merge in my view — both are covered by tests or trivially observable — but they are stated rather than glossed, and the first one matters because **#112 depends on that log line existing**.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/flyout_geometry.rs
+++ b/apps/desktop/src-tauri/src/flyout_geometry.rs
@@ -1,0 +1,210 @@
+//! Where the tray flyout goes (DBSYNC-85).
+//!
+//! Split out of `position_flyout` in `lib.rs` so the placement math takes the **monitor list
+//! as an input** instead of looking it up. That is not tidiness: the development machine has
+//! one display, and the bug being chased only appears with two. Once a layout is four numbers
+//! in a parameter, the arrangements nobody here owns become ordinary test data.
+//!
+//! **What this module deliberately does NOT do** is correct the reported multi-monitor
+//! inversion. The suspicion is that `tray-icon` reports clicks in Cocoa coordinates (origin
+//! bottom-left) while `monitor_from_point` expects winit's top-left space — but that is
+//! unmeasured, and a coordinate flip has roughly a coin-flip chance of looking correct on one
+//! arrangement while being wrong on another. Measurement is GitHub #112, the fix is #113.
+
+/// A display, as far as placement is concerned. Physical pixels, top-left origin — the space
+/// Tauri's `Monitor` reports in.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub(crate) struct MonitorBox {
+    pub x: f64,
+    pub y: f64,
+    pub width: f64,
+    pub height: f64,
+}
+
+impl MonitorBox {
+    fn contains(&self, px: f64, py: f64) -> bool {
+        px >= self.x && px < self.x + self.width && py >= self.y && py < self.y + self.height
+    }
+}
+
+/// The tray icon's rectangle, in physical pixels.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub(crate) struct TrayRect {
+    pub x: f64,
+    pub y: f64,
+    pub width: f64,
+    pub height: f64,
+}
+
+/// Picks the display a click at `(px, py)` landed on.
+///
+/// Falls back to the first entry — the caller passes the primary first — when the point is
+/// outside every display. That happens more often than it sounds: it is exactly what a
+/// coordinate-space mismatch produces, and returning `None` here would leave the window
+/// unplaced rather than merely misplaced.
+pub(crate) fn monitor_for_point(monitors: &[MonitorBox], px: f64, py: f64) -> Option<MonitorBox> {
+    monitors
+        .iter()
+        .find(|m| m.contains(px, py))
+        .or_else(|| monitors.first())
+        .copied()
+}
+
+/// The flyout's top-left corner, clamped to stay wholly inside `monitor`.
+///
+/// Right-aligned to the tray icon, and **below** it on macOS: the menu bar is at the top of the
+/// screen, so a window anchored above it would be off-screen. The previous code subtracted the
+/// window height — a bottom-taskbar assumption inherited from Windows — and the clamp then
+/// pushed the result back to the top of the display, which made it look correct by accident on
+/// a single monitor. Windows keeps the original behaviour: its taskbar really is at the bottom.
+pub(crate) fn flyout_origin(
+    tray: TrayRect,
+    monitor: MonitorBox,
+    win_w: f64,
+    win_h: f64,
+    gap: f64,
+) -> (f64, f64) {
+    let x = tray.x + tray.width - win_w;
+
+    #[cfg(target_os = "macos")]
+    let y = tray.y + tray.height + gap;
+    #[cfg(not(target_os = "macos"))]
+    let y = tray.y - win_h - gap;
+
+    clamp_into(x, y, monitor, win_w, win_h)
+}
+
+/// Keeps the window inside the display.
+///
+/// When the window is larger than the display, `max < min` and `clamp` would panic, so that
+/// case pins to the origin instead. A window wider than the screen is a bad look; a panic in
+/// the click handler is a crash.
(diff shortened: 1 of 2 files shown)


mbsanchez:
## Instrumentation now verified — and why it could not be, before

The PR shipped with the instrumentation **unverified**, which mattered because #112 depends on that log line existing.

**Why it could not be checked:** the tray icon is not visible in the menu bar on this machine — the bar is full and macOS overflows it, exactly as the maintainer reported earlier today. No icon to click, so no way to make the click-time line fire.

**The fix, in the same spirit as the rest of the ticket:** also log the layout at `setup()`, so the plumbing does not depend on a gesture that may be impossible.

```
INFO dbsync85: display layout at startup
  monitors=[MonitorBox { x: 0.0, y: 0.0, width: 2880.0, height: 1800.0 }]
```

One display, 2880x1800 physical — correct for a 1440x900 @2x panel. That proves the `dbsync85` target passes the global filter, that the monitor collection works, and that the output is readable. The click-time line uses the same macro and target, so **#112 is no longer a leap of faith** — and it gets half its answer for free, since the layout is logged before anyone clicks.

`collect_monitor_boxes` was extracted so startup and click share one definition rather than drifting — same reasoning as `tray_icon_for_label` in DBSYNC-78.

189 tests still pass.

### Still not verified

**Single-display placement has not been re-checked visually**, for the same reason: no clickable tray icon. The Y-anchor arithmetic is tested and mutation-checked; the pixels are not.

mbsanchez:
## The unverified claim is now verified — 47b54dc

The previous comment flagged that **the instrumentation had never been seen to emit**, and that #112 depends on it. That is resolved, and it needed a code change to resolve rather than another attempt.

### Why it could not be verified before

The click-time line only fires when someone clicks the tray icon — and **the icon was not in the menu bar at all**. The bar had overflowed, which the maintainer had already reported this morning as a macOS behaviour rather than a bug. So there was no click to be had, on a machine that also has only one display.

### The fix, which stands on its own merits

The display layout is now logged at **launch** as well. The plumbing no longer depends on a click, and the measurement in #112 gets half its context for free.

```
INFO dbsync85: display layout at startup
     monitors=[MonitorBox { x: 0.0, y: 0.0, width: 2880.0, height: 1800.0 }]
```

Emitted on two consecutive launches. **2880x1800 physical is the 1440x900@2x built-in display** — so the numbers are correct, not merely present. A logging call that fired with garbage would have looked identical in a grep for the target.

Monitor collection moved into a helper shared by both call sites, so the startup line and the click line cannot drift into describing different things.

### Test plan delta

- `cargo test` — **189 passed**, unchanged.
- The instrumentation claim moves from *"should work"* to *"observed working, with the right numbers"*.

### Still not verified

**The click-time line specifically**, and the corrected Y anchor on screen. Both need the tray icon to be reachable, which needs the menu bar to have room. The startup line proves the target, the filter and the formatting all work, so the remaining risk on the click line is small — but it is not zero, and it is not being claimed.

mbsanchez:
## CTO Review: APPROVED ✅ — and this one has CI behind it

Run [32562594498](https://github.com/mbsanchez/DropboxSync/actions/runs/32562594498): **success**. The branch was merged up to `main` first so it would run against the pipeline restored in #115 — which means this is the **first PR in this repository ever checked by CI**, rather than by someone remembering to run things locally.

### What the review verified beyond the diff

- **No positioning arithmetic left outside the pure module.** `grep -nE "rect_pos\.y -|clamp\(" lib.rs` on the branch returns nothing. The old inline math is gone, not duplicated — which was the risk with an extract-and-rewrite of this shape.
- **Scope**: 2 files. No frontend, no IPC, no capability change, no new dependency.
- **The tests can fail** — mutation-checked, not assumed: restoring the Windows anchor on macOS fails `macos_hangs_the_flyout_below_the_menu_bar`; forcing the monitor lookup to always pick `[0]` fails 5.
- **The Windows arm was compiled and executed** via the `cfg`-pivot technique from DBSYNC-81. Without it, the Windows guarantee would rest on a test that never runs on any machine anyone has.
- **The instrumentation is observed working**, with the right numbers — `2880x1800` matching the 1440x900@2x display. That was the one claim the first commit could not make.

### The decision I want on record

**Not fixing the inversion in this PR is the correct call**, and it would have been easy to get wrong. The Cocoa-coordinates theory is plausible and unmeasured; a sign flip has roughly even odds of looking right on one arrangement while being wrong on another. That is worse than shipping nothing, because it would close the ticket.

What ships instead is the thing that makes the measurement cheap: one log line carrying the tray position, the tray rect and every display together. #112 is now a click, not an investigation.

### Still not covered, and stated rather than glossed

- **The click-time log line specifically.** The startup line proves the target, filter and formatting all work, so the residual risk is small — but the tray icon was unreachable (menu bar overflow), so it has not been seen.
- **The corrected Y anchor on screen.** The arithmetic is tested; the pixels are not.

Neither is worth blocking on. Both are named so that if #112 comes back empty, the first thing to check is already written down.

Merging.

Issue #85, opened 2026-08-19
DBSYNC-72 Slice 5: CI release workflow reproducing the proven manual path (AFK)


mbsanchez:
Vertical slice 5 of DBSYNC-72 (https://manuelbsan.atlassian.net/browse/DBSYNC-72).

**Type:** AFK · **Blocked by:** Slice 4 (hard gate — every Slice 4 AC must pass first) · **Scope:** `.github/workflows`, `apps/desktop/src-tauri`

Automates only what was already proven by hand in Slice 4. Introduces no new signing approach.

## Acceptance criteria
- [ ] A new tag-triggered release workflow is added, separate from `.github/workflows/ci.yml`.
- [ ] `ci.yml` is left building unsigned on PRs with `DROPBOX_APP_KEY: dummy` and still passes.
- [ ] Secrets used match the environment-variable table on the Jira ticket: `APPLE_CERTIFICATE`, `APPLE_CERTIFICATE_PASSWORD`, `APPLE_SIGNING_IDENTITY`, `APPLE_ID`, `APPLE_PASSWORD`, `APPLE_TEAM_ID`.
- [ ] A temporary keychain is created for the job and deleted in an `always()` step.
- [ ] The release job reproduces the exact commands proven manually in Slice 4.
- [ ] No Apple secret is ever echoed to the job log.
- [ ] Secrets documented in the repo (names only, never values) so the pipeline is reproducible.

mbsanchez:
## Plan: CI release workflow reproducing the proven manual path (GitHub #85)

**Scope dir(s)**: `.github/workflows/` (new file), `apps/desktop/src-tauri` (documentation only — a secrets-names table, no code)
**Implementing agent(s)**: `cowork-workflow:devops-engineer` for the workflow YAML, `cowork-workflow:tauri-expert` to confirm the Tauri build invocation matches Slice 4's proven commands exactly
**Estimated complexity**: medium — the YAML itself is not complex, but every step must exactly mirror commands already proven by hand in Slice 4, and secret handling (temp keychain lifecycle, non-echoing) needs care.
**Dependencies**: Blocked by Slice 4 (#84) — hard gate, every AC on #84 must pass before this slice starts. Also logically depends on Slice 2 (#82) for the `tauri.conf.json`/`project.pbxproj` state this workflow builds against, and implicitly on Slice 1 (#81) since the shipped build must contain the corrected overlay path.

### Steps

[...]

PR #115, opened 2026-08-22
ci(#DBSYNC-74): restore the pipeline, and make it worth running

mbsanchez:
## Summary

CI was switched off on **2026-07-12, four minutes and 39 seconds after its 28th consecutive red run**. This re-enables it — and changes what it does, because re-enabling alone would have delivered almost nothing.

Closes DBSYNC-74.

## Why it was red for three months

Verified, not inferred: **28 runs, 28 failures, zero successes, ever.**

```
$ gh run list --limit 100 --status success --json databaseId --jq 'length'
0
$ gh run list --limit 100 --json databaseId --jq 'length'
28
```

Every one died on the same thing:

```
error[E0425]: cannot find type `PathBuf` in this scope
  --> src/run_events.rs:17:32
```

A one-line missing import, in a `#[cfg(any(target_os = "macos", target_os = "ios"))]` arm. It compiled fine on Windows — where the heavy CfAPI work was happening — and failed only on the macos-latest runner. That is why nobody hit it locally and why it survived from the very first run.

**It is already fixed.** `1334c12` (2026-08-19) added the cfg-gated import. Nothing stood between this workflow and a green run except the disable flag.

## Why simply re-enabling was not enough

The old workflow compiled and bundled. It ran **zero** of the 214 Rust tests, no clippy, and no Swift checks.

Of the defects that reached code review on 2026-08-21 — a string-prefix path guard, a badge push filtered by the wrong set, a settings path that does not exist on macOS 26 — **a compile-only pipeline would have caught none of them.** Restoring that would have restored a green badge and no safety.

So the workflow now runs:

| Step | What it covers |
|---|---|
| `cargo test` | 214 tests that have never once run in CI |
| `cargo clippy --all-targets` | lints, deliberately not fatal — see below |
| `Tests/run-badge-diff-tests.sh` | the Swift checks added for DBSYNC-84 |
| `npm run build` | frontend build and typecheck |
| `npm run bundle:dev` | the app links |

## Two decisions worth arguing with

**clippy runs without `-D warnings`.** `main` carries 4 pre-existing warnings. Failing on them would put this workflow straight back to permanently red — which is precisely the failure mode that got it switched off. Clearing those warnings is its own change; making them visible is this one. Turning the screw before the workflow has ever been green would repeat history.

**`bundle:dev` instead of `tauri build`.** `tauri.conf.json` sets `targets: "all"`, which on macOS means app **plus DMG**. Because every run died at the cargo step, **the bundling stage has never executed in CI even once** — so it is entirely unproven ground. A PR check needs to prove the app links; producing an installer belongs to the release workflow (GitHub #85). Deferring it keeps the first green run reachable.

## Also: a false claim corrected

`build-appex.sh` carried a comment asserting its principal-class guard *"ran green on macos-latest in CI via `tauri build`"*. It never did — `git log -S` places that comment in `e3f045c` on 2026-08-19, **five weeks after CI was disabled**, and CI has never been green at any point. The comment now says what is actually true.

This is the second confident-wrong claim found in this repo in two days. It is worth noticing that both were about things nobody could easily check.

## Stack

`desktop-tauri` — CI configuration and one shell comment. No application code.

## Test Plan

- [x] **Every new step run locally first, exit code checked**:
  ```
  Finder Sync extension checks  -> exit 0   (34 checks pass)
  cargo clippy --all-targets    -> exit 0   (4 pre-existing warnings, non-fatal)
  cargo test                    -> exit 0   (189 passed)
  npm run build                 -> exit 0
  ```
- [x] YAML parses.
- [x] The `PathBuf` fix confirmed present on `main` at `run_events.rs:2-3`.
- [x] Workflow re-enabled via `gh workflow enable ci.yml`.

### What this PR does NOT establish

**No CI run has been green yet.** This PR should trigger the first one in the repo's history, and that run is the real test — not anything asserted here.

Two things could still make it red, and both are named rather than hoped away:
- The Finder Sync extension has changed substantially since July (sandbox entitlements, principal-class guard). It was built locally under **Xcode 15.4**; the runner has a newer Xcode. That combination has never been exercised.
- `npm run bundle:dev` invokes `xcodebuild` through `tauri:before-build`. The July logs show that path working on the runner, but for the July version of the extension.

If the run comes back red, that is information the project has not had since April, and it is the point of the exercise.

🤖 Generated with Claude Code

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,8 @@ jobs:
   desktop-build:
     runs-on: macos-latest
     env:
+      # Literals, not secrets. The app reads these through `option_env!` with fallbacks
+      # (auth/oauth.rs), so CI only needs them to be present, never to be real.
       DROPBOX_APP_KEY: dummy
       DROPBOX_REDIRECT_URI: http://localhost:53682/callback
     steps:
@@ -19,9 +21,36 @@ jobs:
           cache: 'npm'
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
+
       - name: Install dependencies
         run: npm install
-      - name: Test frontend build
+
+      # The four steps below are the reason this workflow is worth running at all.
+      # Until 2026-08-22 it only compiled and bundled: 214 Rust tests and the Swift
+      # checks had never run in CI once. Of the defects that reached code review on
+      # 2026-08-21, a compile-only pipeline would have caught none of them.
+
+      - name: Rust tests
+        run: cargo test --manifest-path apps/desktop/src-tauri/Cargo.toml
+
+      - name: Rust lints
+        # Not `-D warnings`: main currently carries 4 pre-existing warnings, and failing
+        # the build on them would put this workflow straight back to permanently red —
+        # which is how it got switched off in the first place. Clearing them is its own
+        # change; making them visible is this one.
+        run: cargo clippy --manifest-path apps/desktop/src-tauri/Cargo.toml --all-targets
+
+      - name: Finder Sync extension checks
+        run: apps/desktop/native/macos/FinderSyncExtension/Tests/run-badge-diff-tests.sh
+
+      - name: Frontend build and typecheck
         run: npm --workspace apps/desktop run build
-      - name: Build macOS alpha bundle
-        run: npm --workspace apps/desktop run tauri build
+
+      - name: Build macOS app bundle
+        # `bundle:dev` (--bundles app), not `tauri build`. The full build also produces a
+        # DMG, a stage that has never executed in CI even once because every run died
+        # earlier. Producing an installer belongs to the release workflow (GitHub #85);
+        # a PR check only needs to prove the app links.
+        run: npm --workspace apps/desktop run bundle:dev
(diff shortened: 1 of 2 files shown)


mbsanchez:
## CI is green — run [32559863181](https://github.com/mbsanchez/DropboxSync/actions/runs/32559863181)

**The first successful run in this repository's history.** 28 runs from 2026-04-07 to 2026-07-12, all failures; this is the 29th.

```
conclusion: success          08:14 elapsed

  success  Rust tests                      <- 189 tests, never run in CI before
  success  Rust lints
  success  Finder Sync extension checks    <- 34 Swift checks, never run in CI before
  success  Frontend build and typecheck
  success  Build macOS app bundle
```

### The two risks named in the PR body both cleared

Neither was hoped away, and both were real unknowns:

- **The Finder Sync extension built under a newer Xcode.** It has changed substantially since July — sandbox entitlements, `CODE_SIGN_ENTITLEMENTS`, the principal-class guard — and had only ever been built locally under Xcode 15.4. It builds on the runner.
- **The bundling stage.** It had never executed in CI once, because every previous run died at the cargo step. It works.

### What changed for the project, concretely

Before this run, **every quality claim on this project rested on commands run by hand on one Mac.** That is no longer true for compilation, 189 Rust tests, clippy, the Swift checks, and the app bundle.

It does not cover everything — nothing here exercises Finder badges, the keychain, or anything needing a UI — but the gap between "someone remembered to run it" and "it is checked" just closed for the parts that can be automated.

Ready to merge.

PR #117, opened 2026-08-23
fix(#DBSYNC-87): badge files that are born while their folder is open

mbsanchez:
Fixes the badge that never appears after hydrating a `.cloudsc` — [DBSYNC-87](https://manuelbsan.atlassian.net/browse/DBSYNC-87). Found on a second Mac during the manual validation of DBSYNC-72 Slice 4 (#84).

## The bug

Hydrate a placeholder with its folder open: the `cloud_only` badge is there, the `syncing` animation plays, and then the hydrated file sits with **no badge** until you leave the directory and come back.

This is symptom 3 of DBSYNC-84, verbatim — *"Hydrating a file does not show its green check"* — surviving the fix that closed that ticket.

## Why it survived

Two measurements taken on the affected machine split the problem cleanly:

- **The backend is fine.** Right after hydrating and before navigating anywhere, `overlay_state.json` already lists the hydrated path as `synced`.
- **The push works.** A file whose tier changes *without changing name* updates live, red to green, without leaving the directory.

So the fault is specific to paths that are **new** — and hydration produces exactly that. It downloads `foo.txt` and deletes `foo.txt.cloudsc`; the tier arrives against a path Finder has never named. Finder only asks about an item while *enumerating* a directory, so a file created in an already-open folder never enters `displayedPaths`, and `pushable` dropped every push for it.

Two other explanations were checked and ruled out, so nobody needs to re-walk them: `BadgeDiff.changes` does report a path that is new in the snapshot, and `displayedPaths` is never cleared.

## The fix

`pushable` now admits a path by either of two routes — Finder asked about it, **or** its parent directory is on screen. The second is tracked by implementing `beginObservingDirectory(at:)` / `endObservingDirectory(at:)`, which this extension never implemented; the comment on `displayedPaths` had already named that gap as the only genuine "no longer displayed" signal.

This is not a relaxation of Apple's rule. `FinderSync.h` says *"Avoid adding badges to items that the Finder hasn't displayed yet"*, and an item in a directory Finder is observing **is** displayed. The previous test approximated "displayed" as "individually asked about", which is narrower than what the header states.

Three things kept deliberately tight:

- **Only the last separator is cut**, so `sub/deep/a.txt` does not match an observed `sub`. Badging items that are not on screen is the opposite mistake.
- **Observed directories are stored absolute**, not relative. The callbacks can fire before the first state read lands, when there is no sync root to be relative to. Conversion happens at push time.
- **The new parameter is required**, not defaulted, so a future call site cannot silently get the old behaviour back.

The decision lives in `BadgeDiff.swift`, where `run-badge-diff-tests.sh` reaches it. DBSYNC-84 produced four defects and all four sat in the file the test script cannot compile.

## Evidence

Full suite green. The check that matters replays the reported sequence as one composition:

```
end to end: hydrating a .cloudsc with the folder open (DBSYNC-87)
  ok   the hydrated file is badged and the placeholder is cleared, in place
  ok   with the folder closed, only the placeholder is touched
```

The second one is the control: with the folder closed, the same snapshot pair reproduces the old broken behaviour exactly.

**Mutation check.** Removing the new clause from `pushable` turns five checks red:

```
FAIL a path never asked about IS pushable when its directory is on screen
FAIL the same path in a subfolder on screen is pushable too
FAIL removals in an observed directory are pushable
FAIL displayed and observed both admit, in one call
FAIL the hydrated file is badged and the placeholder is cleared, in place
BadgeDiff: 5 check(s) FAILED
```

A check that cannot fail is not evidence, so this is the part that counts.

**Build.** `npm run build:finder-sync` → BUILD SUCCEEDED, with 22 real Swift compile tasks (the previous run compiled nothing and still said SUCCEEDED). The four new symbols are linked, against a negative control that returns 0:

```
beginObservingDirectory  4     noSuchSymbolXYZ  0
endObservingDirectory    4
observedDirectories     12
parentDirectory          2
```

`override func` also does real work here: it makes the compiler verify both callbacks exist on `FIFinderSync`, so a mistyped selector could not have compiled.

**No Rust changed** — `git diff --name-only -- '*.rs'` returns nothing. `cargo test` was deliberately not run: it has no bearing on a Swift-only diff, and on this machine it overwrites the real `overlay_state.json` and points Finder Sync at a deleted tempdir (DBSYNC-75, still open). Say the word and I will run it anyway.

## What this does NOT prove

The checks prove the decision, not that Finder calls `beginObservingDirectory` for the sync folder. If it does not, the observed set stays empty and the bug is unchanged — a silent failure of exactly the shape this project has paid for before.

Manual validation is still required, and its AC is written to defeat the way this bug hides: **evidence captured without navigating away from the directory**.

#DBSYNC-87

### DIFF
--- a/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
+++ b/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
@@ -60,11 +60,45 @@ enum BadgeDiff {
     /// has caught up, so nothing is badged — and when the tier arrives moments later the
     /// path is filtered out and never badged at all. Finder does not ask twice. That is
     /// symptom 1 of DBSYNC-84, reintroduced by the very filter meant to respect Apple's rule.
-    static func pushable(_ changes: Changes, displayed: Set<String>) -> Changes {
-        Changes(
-            changed: changes.changed.filter { displayed.contains($0.key) },
-            removed: changes.removed.filter { displayed.contains($0) }
+    ///
+    /// `observedDirectories` closes a second hole in the same idea (DBSYNC-87). Finder only
+    /// asks about an item while **enumerating** a directory, so a file that is *created*
+    /// while its folder is already open is never asked about and never enters `displayed`.
+    /// Hydration does exactly that: it downloads `foo.txt` and deletes `foo.txt.cloudsc`, so
+    /// the tier arrives against a path Finder has never named. Every push for it was dropped
+    /// here, and only re-entering the directory — which forces a fresh enumeration — showed
+    /// the badge.
+    ///
+    /// An item in a directory Finder is currently observing **is** displayed, which is the
+    /// condition Apple's first sentence actually states. Widening to "asked about, or sitting
+    /// in a folder on screen" is therefore closer to the documented rule than the narrower
+    /// test it replaces, not a relaxation of it.
+    static func pushable(
+        _ changes: Changes,
+        displayed: Set<String>,
+        observedDirectories: Set<String>
+    ) -> Changes {
+        func isPushable(_ path: String) -> Bool {
+            displayed.contains(path) || observedDirectories.contains(parentDirectory(of: path))
+        }
+        return Changes(
+            changed: changes.changed.filter { isPushable($0.key) },
+            removed: changes.removed.filter(isPushable)
         )
+    }
+
+    /// The directory part of a relative path: `"sub/a.txt"` → `"sub"`, `"a.txt"` → `""`.
+    ///
+    /// `""` is the sync folder itself, which is a real answer rather than a missing one —
+    /// files sitting directly in the sync root are the common case, and their parent has to
+    /// compare equal to the relative path of the root that [`relativePath`] produces.
+    ///
+    /// Only the LAST separator is cut, so a file in a subfolder of an observed directory does
+    /// not match that directory. `"sub/deep/a.txt"` yields `"sub/deep"`, not `"sub"` — and it
+    /// should, because Finder showing `sub` does not put the contents of `sub/deep` on screen.
+    static func parentDirectory(of relative: String) -> String {
+        guard let separator = relative.lastIndex(of: "/") else { return "" }
+        return String(relative[relative.startIndex..<separator])
     }
 
     /// Whether `path` is `root` itself or lives inside it.
--- a/apps/desktop/native/macos/FinderSyncExtension/SyncExtension.swift
+++ b/apps/desktop/native/macos/FinderSyncExtension/SyncExtension.swift
@@ -24,13 +24,27 @@ final class DropboxSyncFinderSync: FIFinderSync {
     /// **This set only ever grows, and that is deliberate.** Clearing a badge does not mean
     /// Finder stopped showing the item — the file is still on screen with no status to
     /// display. Dropping a path here when its badge is cleared means a file that leaves the
-    /// state map and comes back is filtered out of the push and never badged again. The
-    /// only genuine "no longer displayed" signal is `endObservingDirectoryAtURL:`, which
-    /// this extension does not implement.
+    /// state map and comes back is filtered out of the push and never badged again.
     ///
     /// The cost of never shrinking is one string per distinct path Finder has shown inside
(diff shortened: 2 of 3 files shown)


mbsanchez:
## Review — no blocking findings, with one caveat about the review itself

**This is a self-review.** The same session wrote the change, so it is weaker than an independent pass by construction. It is recorded that way rather than presented as an approval of equal weight. What follows is the result of going at the places this change could break silently, not a read-through.

### Checked by running, not by reasoning

**Trailing separators — the most likely silent failure, and it does not happen.** `pushable` matches `parentDirectory(of: "sub/a.txt")` → `"sub"` against the relative form of an observed directory. If Finder handed the directory with a trailing slash, the observed entry would be `"sub/"`, nothing would ever match, and the fix would be inert **with every check still green**. Measured:

```
file:///Users/x/DropboxSync/sub/    -> standardized.path=[/Users/x/DropboxSync/sub]
file:///Users/x/DropboxSync/        -> standardized.path=[/Users/x/DropboxSync]
file:///Users/x/DropboxSync/./sub/  -> standardized.path=[/Users/x/DropboxSync/sub]
```

`standardizedFileURL.path` never keeps one, and it collapses `./` too. Both sides of the comparison are built through the same normalisation, so they agree.

**`override` is load-bearing.** Both new callbacks are declared `override`, so the compiler verifies they exist on `FIFinderSync`. A mistyped selector — the classic way to add a no-op hook that looks right — could not have compiled.

### Considered and accepted

**Threading.** `observedDirectories` is mutated from Finder's callbacks and read from the 2-second timer. That is the *same* arrangement `displayedPaths` has had since DBSYNC-84 — mutated in `requestBadgeIdentifier`, read in the timer — so this adds no new class of risk. It does not make the existing assumption safer either; if that assumption is ever wrong, both sets are wrong together. Not this PR's to fix.

**Push volume.** Admitting a whole observed directory could in principle push more than before. In practice Finder enumerates a directory when it opens it, so almost everything in an open folder is already in `displayedPaths`; the delta is the files born afterwards, which is the case being fixed. The first-load guard is untouched and explicitly covered.

**Leak if `endObservingDirectory` never fires** (Finder crash, or begin/end handing back non-identical path forms). The set would retain a closed directory and keep pushing into it. Bounded by the number of directories opened in one session, and pushing to a directory Finder is not showing is ignored rather than harmful. Noted, not guarded.

### On the shape of the fix

The decision sits in `BadgeDiff.swift` where the check script reaches it, and `SyncExtension.swift` only applies it — the rule DBSYNC-84 produced after four defects hid in the file the tests cannot compile. The new parameter is required rather than defaulted, so the old behaviour cannot come back by omission at a future call site.

### What review cannot settle

Whether Finder actually calls `beginObservingDirectory` for this extension's monitored folder. If it does not, the observed set stays empty, every check stays green, and the bug is unchanged. That is the failure mode this project has paid for repeatedly, and only the screen answers it.

**Verdict: approved to QA**, with the live check as the gate — hydrate a `.cloudsc` and watch the badge without navigating away.

mbsanchez:
## QA PASSED ✅ — verified live on the second Mac

Maintainer's report, on the notarized build of `eb22d1b` (submission `d3cab45a-a7c6-4090-af02-8ee16beb1377`, Accepted), installed on a Mac where the app was never built:

> **the badge now appears after hydrating**

That is the AC, and it is the one the previous two tickets could not reach: the badge arrives **without leaving the directory**. Re-entering was the gesture that masked this bug, so a result obtained without it is the whole point.

### What this settles beyond the AC

The open question at review time was whether Finder actually calls `beginObservingDirectory(at:)` for this extension's monitored folder. It does. Every check in this PR was green regardless of the answer, so that question could only ever be closed on the machine — and it is now closed affirmatively. The mechanism works end to end, not just the decision it rests on.

### Not covered by this run

- Nested subfolders. The pushable widening deliberately matches only the immediate parent, so a file in `sub/deep` is not admitted by an observed `sub`. Covered by checks, not observed on screen.
- Removal in an open folder (a file leaving the state map while visible). Covered by checks only.

Neither blocks: both are the conservative direction — they can only withhold a badge, never plant a wrong one.

**Verdict: PASS.** Jira ticket stays in QA per this project's flow; the verdict lives here.

#DBSYNC-87

PR #118, opened 2026-08-24
fix(#DBSYNC-88): tell the user the app is translocated instead of blaming the extension

mbsanchez:
Fixes the banner that insists the Finder extension is off while it is on — [DBSYNC-88](https://manuelbsan.atlassian.net/browse/DBSYNC-88).

## The bug

On the second Mac, with the extension enabled and all three badge tiers rendering on screen, the app kept showing **"Finder badges are switched off"**. It survived a full relaunch and a complete uninstall/reinstall into `/Applications`.

Four explanations were ruled out on the machine before this one was reached: stale UI, multiple registered plug-in instances (`pluginkit` reports exactly one), two copies of the app, and a stale process running the old binary.

## Root cause

`FIFinderSyncController.isExtensionEnabled` is an AppKit class property, and it was read on whatever thread called it. `get_startup_requirements` is a **synchronous** Tauri command, which Tauri runs on a pool thread rather than the main one — so on macOS 26 the read returned `false` while the extension was enabled.

What makes this the mechanism rather than a guess is sitting in the same file: `open_finder_extension_settings` already dispatches through `run_on_main_thread`, with a comment saying AppKit requires it. DBSYNC-86 applied that reasoning to the call that *shows* the settings pane and not to the one that *reads* the state. Nothing connected the two until the banner was caught lying on a real machine.

## The fix

The read goes through `run_on_main_thread` using the existing `APP_HANDLE` static, so no signatures change and no call site moves.

Two deliberate details:

**It reads directly when already on the main thread.** Dispatching would queue work for a thread that is blocked waiting for that work — a deadlock. This function is reachable from both sides, so the check is not defensive noise.

**Every failure path yields `NotApplicable`, never `Disabled`.** No handle yet (unit tests, early startup), dispatch refused, or a 500 ms timeout on a busy main thread. That is exactly what the third state was introduced for in DBSYNC-86: an unknown must not render a warning. A false alarm shown to someone who already did the right thing is worse than silence, so the failure direction is silence.

Also corrects the copy promising that restarting Finder closes your open windows. It does not — macOS restores them. That wording cost real time in this very session: Finder was restarted, the windows stayed open, and the honest conclusion was that nothing had happened.

## Checks

```
cargo test    192 passed; 0 failed
cargo clippy  clean — its warnings are pre-existing and in other files (0 mentions of finder_extension.rs)
npm run build tsc + vite, clean
```

`cargo test` was run with `~/Library/Application Support/DropboxSyncDesktop/overlay_state.json` backed up first and restored afterwards. It **did** overwrite it — hashes before and after differ — which is live confirmation that DBSYNC-75 is still real, in passing.

## What this does NOT prove

`isExtensionEnabled` is bundle-scoped: outside the shipped app it reports on nothing, so no test in this repo can tell a fixed read from a broken one. That limitation is documented on the function itself and is why DBSYNC-86 shipped this defect in the first place.

**The only check that counts is the banner on a Mac with the extension ON.** And the mutation matters as much as the fix: with the extension switched **off**, the warning must still appear. "Never warns" is a way to pass this PR while removing the feature.

#DBSYNC-88

### DIFF
--- a/apps/desktop/src-tauri/src/finder_extension.rs
+++ b/apps/desktop/src-tauri/src/finder_extension.rs
@@ -338,6 +338,22 @@ mod tests {
         assert!(!warns(FinderExtensionState::Translocated));
     }
 
+    /// Pins the wire format against `FinderExtensionState` in `apps/desktop/src/types.ts`.
+    ///
+    /// Nothing else checks this seam. The frontend compares these values as string literals,
+    /// so renaming a variant here — or changing the serde casing — would make every comparison
+    /// silently false: no banner, no error, no failing build, in either language. That is the
+    /// same shape of defect as the one this ticket exists for, so it gets a test rather than a
+    /// convention.
+    #[test]
+    fn serialized_names_match_the_typescript_union() {
+        let wire = |s: FinderExtensionState| serde_json::to_string(&s).unwrap();
+        assert_eq!(wire(FinderExtensionState::Enabled), "\"enabled\"");
+        assert_eq!(wire(FinderExtensionState::Disabled), "\"disabled\"");
+        assert_eq!(wire(FinderExtensionState::Translocated), "\"translocated\"");
+        assert_eq!(wire(FinderExtensionState::NotApplicable), "\"notApplicable\"");
+    }
+
     /// The exact path the instrumented build logged on the machine that reproduced DBSYNC-88.
     /// Pinned verbatim: this is the observation the whole fix rests on, and a predicate that
     /// stopped matching it would silently restore the false "extension is off" banner.
--- a/apps/desktop/src-tauri/src/finder_extension.rs
+++ b/apps/desktop/src-tauri/src/finder_extension.rs
@@ -40,34 +40,91 @@ pub(crate) enum FinderExtensionState {
 /// test or spike outside it. The check that matters is running the app with the extension ON
 /// and confirming this says `Enabled`; an implementation stuck at `Disabled` looks perfectly
 /// correct as long as you only ever test with it switched off.
+/// **Must be read on the main thread** (DBSYNC-88). This is not a precaution, it is the
+/// bug: DBSYNC-86 read it on whatever thread called it, and `get_startup_requirements` is a
+/// synchronous Tauri command, which Tauri runs on a pool thread rather than the main one. On
+/// macOS 26 that returned `false` with the extension enabled and rendering badges, so the app
+/// warned the user that a working feature was switched off, permanently and across relaunches.
+///
+/// The sibling API in this same file already knew: `open_finder_extension_settings` dispatches
+/// through `run_on_main_thread` and says AppKit requires it. The read did not, and nothing
+/// connected the two until the banner was seen lying on a real machine.
 pub(crate) fn finder_extension_state() -> FinderExtensionState {
     #[cfg(target_os = "macos")]
     {
-        use objc2::runtime::AnyClass;
-        use objc2::msg_send;
+        // Already on the main thread: read directly. Dispatching instead would deadlock —
+        // `run_on_main_thread` queues the closure for a thread that is currently blocked
+        // waiting for it.
+        if is_main_thread() {
+            return read_extension_enabled();
+        }
 
-        // A missing class is an answer, not a crash: if `FinderSync.framework` is not loaded
-        // there is nothing to report and `NotApplicable` is the honest value. This is why the
-        // "must not panic" requirement needs no `catch_unwind` — `get` already returns Option.
-        let Some(cls) = AnyClass::get(c"FIFinderSyncController") else {
+        // No handle yet (unit tests never call `setup()`, and startup can ask early). "Could
+        // not determine" is `NotApplicable`, never `Disabled` — the whole point of having a
+        // third state is that an unknown must not render a warning.
+        let Some(app) = crate::state::APP_HANDLE.get() else {
             return FinderExtensionState::NotApplicable;
         };
 
-        // SAFETY: `FIFinderSyncController` is an Apple class from FinderSync.framework, and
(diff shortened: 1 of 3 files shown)


mbsanchez:
## Correction: the main-thread change was NOT the fix

This PR opened claiming that reading `isExtensionEnabled` off the main thread was the cause. **It was not.** That build (`9b8549d`) was installed on the affected Mac and the banner did not move. A reasoned guess, falsified by the machine — the fourth time this ticket family has had a premise killed by running something.

The title has been updated to describe what this PR actually does now.

### What the instrument found

Rather than propose a fourth hypothesis, `969302b` added one log line at the moment of the read. It answered the question immediately:

```
finder_extension: enabled=false
  exe=/private/var/folders/…/T/AppTranslocation/71877875-…/d/DropboxSyncDesktop.app/Contents/MacOS/dropbox_sync_desktop
  bundle_id=com.mobsanchez.dropboxsyncdesktop  activation_policy=Some(1)  app_is_active=Some(false)
```

**App Translocation.** macOS was running the app from a randomised read-only mount. `isExtensionEnabled` reports on the appex inside the *calling* bundle, so it was answering about the translocated copy — not the installed one the user enabled. `false` was true about the wrong bundle.

It explains every observation together: survives relaunch and reinstall (translocation recurs each launch), `pluginkit` shows one registered instance (the real one), badges keep rendering (drawn by that real one), and the dev Mac never reproduced it (a locally built app carries no quarantine flag, so **that machine cannot reproduce this class of bug at all**).

Causal chain, with the middle step verified directly rather than assumed — setting `com.apple.quarantine` on a zip and extracting with `ditto -x -k` propagates it to the `.app` and its executable:

```
zip picks up com.apple.quarantine in transfer → ditto -x -k preserves it
                                              → macOS launches translocated
                                              → isExtensionEnabled answers for the wrong bundle
```

Confirmed by removing the cause: `xattr -dr com.apple.quarantine …` plus a relaunch, and the banner disappears.

### What this PR now does

Answers `Translocated` **before** the read. Reading first and overriding afterwards would leave `Disabled` reachable from a condition that can never justify it, and `Disabled` is the value that raises the false alarm.

It is its own state, not `Disabled` (a lie) and not `NotApplicable` (silence, leaving the user with no badges and no explanation). Its banner carries **no "Open settings" button** — the extension is very probably fine, and sending someone to a checkbox that is already ticked is precisely what made this expensive.

Detection is a pure predicate on the executable path, so it is testable. The test pins the exact path the instrumented build logged: that observation is what the whole fix rests on, and a predicate that stopped matching it would silently restore the false banner. Ordinary install locations are covered too, including a directory merely *named* after the mechanism — without that, "never warns" could pass as a fix.

### Kept, and re-labelled

The main-thread dispatch stays as correctness hygiene: an AppKit class property should be read on the main thread, and its failure paths all yield `NotApplicable` rather than `Disabled`. It is no longer described as the fix.

The instrumentation stays too. It ended a three-hypothesis investigation in one reading, and the property it reports on cannot be reached by any test in this repo.

### Checks

```
cargo test    194 passed (2 new); mutating the detector to `false` turns the pinned test red
cargo clippy  clean, nothing in the changed file
npm run build tsc + vite clean
```

`cargo test` ran with `overlay_state.json` backed up and restored — it still overwrites it (DBSYNC-75).

### Still to verify on the machine

Both halves, and the second matters as much: translocated → the move-to-Applications banner; extension genuinely off and **not** translocated → the original warning still appears.

#DBSYNC-88

mbsanchez:
## Review — approved to QA, with one finding fixed during the pass

**Self-review**, same caveat as PR #117: the session that wrote this reviewed it, so it is weaker than an independent pass. Recorded rather than dressed up.

### Finding, now fixed (3b9f0cb)

**The Rust↔TypeScript seam was unguarded.** `FinderExtensionState` serialises to string literals that `FlyoutApp.tsx` compares directly. Renaming a variant, or changing the serde casing, would make every comparison silently false — no banner, no error, no failing build, in either language. Exactly the shape of defect this ticket exists for: a wrong answer indistinguishable from a correct one.

The seam turned out to be correct (`"translocated"` matches `types.ts`), but it was correct by luck. It is now pinned by a test over all four variants.

### Checked, and fine

**`current_exe()` really does return the translocated path.** Not assumed — it is how the bug was found. The instrumented build printed it from the affected machine, and the test pins that exact string.

**The check runs before the read, not after.** This matters more than it looks: reading first and overriding afterwards would leave `Disabled` reachable from a condition that can never justify it, and `Disabled` is the value that raises the false alarm.

**No banner flash on launch.** `finderExtension` initialises to `"notApplicable"`, so nothing renders before the first read lands.

**The mutation check bites.** Forcing the predicate to `false` turns the pinned test red; the tree was restored from the commit afterwards, not by hand-editing back.

### Accepted, with eyes open

**A directory named exactly `AppTranslocation` would false-positive.** A near-miss (`AppTranslocationNotes`) is covered by a test, but an exact match is not distinguishable by a path predicate. The consequence is the app saying "move me" instead of reading the state — annoying, never a false "extension is off". `SecTranslocateIsTranslocatedURL` is the documented alternative and is noted in the code as such; it was not taken, because it is a C API needing CFURL bridging for one boolean while the marker is stable and already observed in logs. Recorded as a decision, not skipped.

**`current_exe()` on every read.** One syscall per startup check and per window focus. Not worth caching, and caching it would reintroduce staleness in a component whose entire history is staleness bugs.

### What review cannot settle

The banner is bundle-scoped behaviour on a machine this one cannot imitate: a locally built app never carries quarantine, so **the dev Mac cannot reproduce this class of bug at all**. Both halves have to be seen on the affected machine — translocated shows the move banner, and extension-off-but-not-translocated still shows the original warning. "Never warns" would otherwise pass as a fix.

**Verdict: approved to QA.**

PR #119, opened 2026-08-24
fix(#DBSYNC-75): write overlay_state.json beside the database, not into the user's real data dir

mbsanchez:
Stops `cargo test` from overwriting the running user's real `overlay_state.json` — [DBSYNC-75](https://manuelbsan.atlassian.net/browse/DBSYNC-75), open since 2026-08-19.

## The problem

`refresh_overlay_state_inner` resolved its destination with the global `db::app_data_dir()`. The test harness redirects only the **database** to a `tempdir()`, so every test that reached the writer wrote to `~/Library/Application Support/DropboxSyncDesktop/overlay_state.json` — with a `sync_folder` pointing at a temporary directory the test then deleted.

The Finder Sync extension reads that file every two seconds and feeds `sync_folder` into `directoryURLs`. So after any `cargo test` run, the user's badges silently stopped rendering.

**The mess is not the point; the trap is.** Anyone validating badges by hand after running the tests sees no badges and starts debugging the extension — the one place the fault is not. It cost real time this week, and during one session the file had to be backed up and restored by hand six times just to keep working.

## The fix

`Db` records the directory it was opened in, and the overlay writer follows it instead of reaching for a global.

In production nothing moves: the database and the overlay file already live in the same directory. Under test the overlay lands in the same tempdir as the database, which is what the harness intended all along.

`Db::new_at`'s own doc already promised that *"running `cargo test` never touches a user's real DB"*. This extends the same promise to everything written beside it, and puts the guarantee somewhere a future sibling writer will trip over rather than in a convention nobody reads.

Chosen over threading a `data_dir` through `AppState`: that would have meant editing all 13 construction sites, 12 of them test helpers, for the same result. Following the database also says something true — these files belong together — rather than merely passing a value around.

## Evidence

The ticket's acceptance criterion is "a full `cargo test` run must not create or modify anything under the real app data dir". Measured directly, on the real file:

```
hash BEFORE:  11762ac30f7f1a9c3d31b14aff1e3f1a9ea4f3955d11d422da81b2a95e3453b2
cargo test →  196 passed; 0 failed
hash AFTER:   11762ac30f7f1a9c3d31b14aff1e3f1a9ea4f3955d11d422da81b2a95e3453b2
```

Byte-identical.

**Mutation check.** Reverting the writer to `db::app_data_dir()` and running the new guard:

```
test overlay_state::tests::the_writer_follows_the_database_and_never_the_global_data_dir ... FAILED
```

and the real file's hash changed to `14110c16…` — the damage is observable, not theoretical. Restored afterwards; the tree is clean.

The guard deliberately does **not** call `db::app_data_dir()` to compare against: that function creates the directory, and a test for "never touch the real data dir" must not touch the real data dir.

## Notes

`clippy` clean, and it says nothing about either changed file. Two stale doc comments corrected in passing — one described the writer as global, and one justified a function split as a workaround for this very bug. The split is kept on its own merits, with the reason rewritten.

#DBSYNC-75

### DIFF
--- a/apps/desktop/src-tauri/src/overlay_state.rs
+++ b/apps/desktop/src-tauri/src/overlay_state.rs
@@ -31,7 +31,7 @@ pub enum OverlayTier {
 /// Versioned shell **status contract** (DBSYNC-51): the single source of per-path
 /// sync state read by every native surface — macOS Finder Sync badges, Linux
 /// file-manager emblems, and the Windows status column. Written atomically to
-/// `overlay_state.json` under [`db::app_data_dir`].
+/// `overlay_state.json` beside the database that produced it (DBSYNC-75).
 ///
 /// Schema (`version = 1`):
 /// - `version`: contract version; bump on a breaking change.
@@ -72,9 +72,11 @@ pub(crate) fn refresh_overlay_state_internal(state: &AppState) {
 /// Per-path tiers for the current state — the whole decision, with no I/O of its own
 /// beyond reading the database and the sync folder.
 ///
-/// Split out of [`refresh_overlay_state_inner`] so it can be tested directly: that
-/// function writes to the real [`db::app_data_dir`], and a test calling it would clobber
-/// the running user's `overlay_state.json` (DBSYNC-75).
+/// Split out of [`refresh_overlay_state_inner`] so it can be tested directly. That was
+/// originally a workaround: the writer resolved its destination globally, so any test
+/// calling it clobbered the running user's `overlay_state.json`. Fixed in DBSYNC-75 — the
+/// writer now follows the database — so the split is no longer load-bearing, but it is
+/// kept: a pure decision function is worth having on its own terms.
 fn compute_overlay_paths(state: &AppState) -> AppResult<HashMap<String, OverlayTier>> {
     let job_paths = active_job_paths(&state.db)?;
     let conflict_paths = unresolved_conflict_paths(&state.db)?;
@@ -140,8 +142,12 @@ fn refresh_overlay_state_inner(state: &AppState) -> AppResult<()> {
         paths,
     };
 
-    let dir = db::app_data_dir()?;
-    let dest = overlay_state_path(&dir);
+    // Beside the database, NOT via the global `app_data_dir()` (DBSYNC-75). In production
+    // these are the same directory. Under `cargo test` they are not: the harness builds its
+    // `AppState` on a `tempdir()`, and resolving globally here overwrote the running user's
+    // real `overlay_state.json` with a sync folder the test then deleted — leaving the
+    // Finder Sync extension pointed at nothing and every badge gone until it was repaired.
+    let dest = overlay_state_path(state.db.data_dir());
     let json = serde_json::to_string_pretty(&payload)?;
 
     let tmp = dest.with_extension("json.tmp");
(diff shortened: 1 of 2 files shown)


mbsanchez:
## Review — approved, with one finding fixed during the pass

**Self-review**, same caveat as the last two: the session that wrote it reviewed it.

### Finding, fixed (0eab004)

**The fallback was dead, and the comment describing it was wrong.** The code read:

```rust
data_dir: path.parent().unwrap_or_else(|| Path::new(".")).to_path_buf(),
```

with a comment claiming `.` covered the bare-filename case. Measured rather than assumed:

```
     app.db  -> Some("")
    /app.db  -> Some("/")
          /  -> None
```

`parent()` returns `Some("")` for a bare filename, so `unwrap_or` never fires for it and `data_dir` would have been the **empty path** — placing `overlay_state.json` in the process's working directory. That is a quieter version of the bug this PR exists to fix.

Unreachable from either call site, both of which pass absolute paths. Fixed anyway: the code promised something it did not do, and a third call site would have inherited the promise rather than the behaviour. Now guarded by a test that asserts the recorded directory is never empty.

Worth noting how this was found — by checking a claim in a comment I had written myself an hour earlier, rather than by reading the logic. The logic was fine.

### Checked, and fine

**Production behaviour is unchanged.** `Db::new()` opens at `db_path()`, which is inside `app_data_dir()`, so the overlay lands exactly where it always did. Nothing to migrate, no existing user file moves.

**The Windows branch is untouched.** `sync_placeholder_states` and `sync_folder_states` still read from `payload`, not from the destination path.

**`overlay_state_path` was already parameterised** on a directory, so the change is at the single call site that fed it a global.

### Accepted

**`Db` grows a field it does not need for its own work.** Justified: it is the only object in `AppState` that already knows where instance data lives, and the alternative was threading a `data_dir` through 13 construction sites, 12 of them test helpers. The field carries the reasoning in its own doc comment rather than in a commit message nobody will find.

### Evidence

The ticket's AC is measured directly on the real file, not inferred:

```
hash BEFORE:  5e43847e14d4fcb8ede8d846de19814a2e30cbcaa28f219827fc5e389d5c0828
cargo test →  197 passed; 0 failed
hash AFTER:   5e43847e14d4fcb8ede8d846de19814a2e30cbcaa28f219827fc5e389d5c0828
```

And the mutation bites: reverting the writer to `db::app_data_dir()` turns the guard red **and** changes that hash.

### On QA for this ticket

Unlike DBSYNC-87 and -88, this one needs no second machine and no screen. Its acceptance criterion is a hash comparison across a test run, which is exactly what was measured above. Recording that here so nobody waits on manual QA that has nothing to look at.

**Verdict: approved.**

mbsanchez:
**Note for the next Windows session.** This change is cross-platform — no `cfg` gates — but it was only exercised on macOS. Production behaviour should be identical on Windows, since `app.db` and `overlay_state.json` already share `%LOCALAPPDATA%\DropboxSyncDesktop\`, and the `cfg(windows)` CfAPI block in `refresh_overlay_state_inner` reads from `payload` rather than from the destination path, so it is textually untouched.

That is reasoning, not a measurement. The maintainer's decision (2026-08-24) is to accept it and verify on the next Windows session rather than spend a cfg-pivot on it now. Recorded here so the assumption is one someone made on purpose, not one nobody remembers making.

CI does not cover Windows yet — DBSYNC-38.

Issue #121, opened 2026-08-24
DBSYNC-38 Slice 2: compile and test the Windows-only code, for the first time ever (AFK)


mbsanchez:
Slice 2 of 3 for [DBSYNC-38](https://manuelbsan.atlassian.net/browse/DBSYNC-38) — **AFK**. Blocked by #120.

**This slice is the whole point of the ticket.** Everything else is scaffolding around it.

## What is currently unchecked

`apps/desktop/src-tauri/src/lib.rs` gates five modules behind `#[cfg(windows)]`:

```
cloud_filter          2024 lines / 92 KB
windows_file_assoc
windows_identity
windows_shell_menu
windows_startup
```

plus a Windows-only dependency block in `Cargo.toml` (`winreg`, `windows-sys`, `windows` 0.61 with `Win32_Storage_CloudFilters`, `Storage_Provider`, `Win32_System_IO`).

**None of it has ever been compiled by CI — not once in the pipeline's entire history.** Of the 222 `#[test]` functions in `src-tauri/src`, roughly 22 sit inside Windows-gated code (19 in `cloud_filter.rs` alone) and have never executed on any machine except the maintainer's. `cargo clippy --all-targets` has never linted a line of the CfAPI layer.

Concretely, today someone can delete a function that `cloud_filter.rs` calls and CI goes green.

## Acceptance criteria

- [ ] The `rust` job runs as a matrix over `macos-latest` and `windows-latest`: `cargo test` and `cargo clippy --all-targets`.
- [ ] **`fail-fast: false`.** With the default, a macOS failure cancels the Windows job — cancelling precisely the signal this ticket exists to buy.
- [ ] The Windows run demonstrably compiles the `cfg(windows)` modules and executes the Windows-only tests. Quote the test count from the Windows job on the PR; it must exceed the macOS count.
- [ ] **Mutation check — the AC that makes this real.** On a throwaway branch, break something in `cloud_filter.rs` (delete a call, change a signature) and confirm the Windows job goes red while macOS stays green. A job that has never been observed failing is not yet evidence that it can.
- [ ] Clippy **without** `-D warnings` on both legs. These modules have never been linted, so the first run will very likely surface warnings; clearing them is DBSYNC-90's neighbour, not this ticket.
[...]

Issue #120, opened 2026-08-24
DBSYNC-38 Slice 1: split the CI job before adding any platform (AFK)


mbsanchez:
Slice 1 of 3 for [DBSYNC-38](https://manuelbsan.atlassian.net/browse/DBSYNC-38) — **AFK**.

Restructure `ci.yml` into separate jobs **without adding any platform**. No new signal is bought here; the point is a provably-green floor that Slices 2 and 3 can stand on. If a Windows job later goes red, this ordering is what lets the failure be attributed to Windows rather than to the restructuring.

## Why it is not a matrix keyword

Three of the six current steps cannot run on `windows-latest`:

- `Finder Sync extension checks` → `Tests/run-badge-diff-tests.sh` uses `xcrun swiftc`
- `Build macOS app bundle` → `bundle:dev` is `--bundles app`; `.app` is a macOS artifact (Windows uses `bundle:win`, `--no-bundle`)
- `Frontend build and typecheck` → portable, but byte-identical on every runner; matrixing it multiplies cost for zero signal

Naively adding `strategy.matrix.os` breaks all three. The job has to be split.

## Target shape

```
rust        (macos-latest)              cargo test + cargo clippy --all-targets
macos-app   (macos-latest)              badge-diff checks + bundle:dev
frontend    (one runner)                npm run build
```

`bundle:dev` stays on the macOS job deliberately: `tauri.conf.json`'s `beforeBuildCommand` runs `scripts/tauri-before-build.mjs`, which invokes `build:finder-sync` when `process.platform === "darwin"`. So that step already compiles the Swift appex via `xcodebuild`. It must never be matrixed.

## Acceptance criteria

- [ ] `ci.yml` splits into the three jobs above. Every check that runs today still runs, and none runs twice.
- [ ] The `env:` block (`DROPBOX_APP_KEY`, `DROPBOX_REDIRECT_URI`) reaches every job that needs it. These are literals read through `option_env!`, not secrets — the existing comment explains why.
- [ ] Rust caching added (`Swatinem/rust-cache` or equivalent). **This is what makes Slice 2 affordable**: the current single job takes 8m05s–8m40s uncached, and a cold Windows build of the `windows` 0.61 family plus `rusqlite` (which compiles SQLite from C) lands well beyond that.
- [ ] npm caching kept — `actions/setup-node` already does it via `cache: 'npm'`.
[...]

Issue #122, opened 2026-08-24
DBSYNC-38 Slice 3: build the Windows app and the shell-menu crate (AFK)


mbsanchez:
Slice 3 of 3 for [DBSYNC-38](https://manuelbsan.atlassian.net/browse/DBSYNC-38) — **AFK**. Blocked by #121.

Extends Windows coverage from "the library compiles and its tests pass" to "the app links and the shell extension builds".

## The gap this closes

`apps/desktop/native/windows/shell-menu/` is a **second, independent cargo crate** — its own `Cargo.toml` and `Cargo.lock`, ten source files, the Explorer context-menu COM DLL. It sits outside the npm workspace and outside every current CI step, yet `windows_shell_menu.rs` registers it at runtime. Nothing has ever built it automatically.

## Acceptance criteria

- [ ] A `windows-app` job on `windows-latest` runs `npm --workspace apps/desktop run bundle:win`. That script is `tauri build --no-bundle`: it proves the app links without producing an installer, mirroring the macOS choice already documented at `ci.yml:52-55`.
- [ ] The same job runs `cargo check --manifest-path apps/desktop/native/windows/shell-menu/Cargo.toml`. `check` rather than `build`: the DLL is not consumed by CI, so type-checking is the signal and linking is cost.
- [ ] The crate's separate `Cargo.lock` is respected — it is not part of the workspace and must not be folded into one.
- [ ] Caching covers this crate's target dir too, or its cost is measured and accepted on the PR.
- [ ] **No installer is produced and nothing is signed.** That is DBSYNC-72 Slice 5, a tag-triggered `release.yml`, and it must not leak into a PR check.
- [ ] Wall-clock delta over Slice 2 recorded on the PR.

## Out of scope

- Registering or loading the DLL. That needs a real Explorer session and a sparse package identity — manual, tracked in DBSYNC-89.
- Any change to the crate's own code.

#DBSYNC-38

mbsanchez:
## Implementation Plan

**Scope dir(s)**: `.github/workflows/` · **Agent(s)**: tauri-expert · **Complexity**: low · **Dependencies**: blocked by #121

### Steps

1. Add a `windows-app` job on `windows-latest`, carrying the same `env:` block.
2. Run `npm --workspace apps/desktop run bundle:win` — that is `tauri build --no-bundle`: it proves the app links and produces no installer.
3. Run `cargo check --manifest-path apps/desktop/native/windows/shell-menu/Cargo.toml`.
[...]

PR #123, opened 2026-08-25
ci(#DBSYNC-38): split the pipeline into one job per kind of check

mbsanchez:
Slice 1 of 3 for [DBSYNC-38](https://manuelbsan.atlassian.net/browse/DBSYNC-38) — GitHub #120.

## Summary

`desktop-build` becomes three jobs. **No platform is added here, on purpose.**

| Job | Runner | Checks |
|---|---|---|
| `rust` | macos-latest | `cargo test`, `cargo clippy --all-targets` |
| `macos-app` | macos-latest | badge-diff checks, `bundle:dev` |
| `frontend` | macos-latest | `npm run build` (tsc + vite) |

## Why a split and not a matrix

Three of the old job's steps cannot run on Windows: the Finder Sync checks shell out to `xcrun swiftc`, and `bundle:dev` produces a macOS `.app` — it also compiles the Swift appex, because `beforeBuildCommand` calls `build:finder-sync` on darwin. Adding `strategy.matrix.os` to the monolith breaks all three.

Keeping Windows out of *this* change is deliberate: when the Windows leg lands in #121, a red build is then attributable to Windows rather than to the restructuring.

## Caching arrives now, not with Windows

The two Rust-touching jobs get `Swatinem/rust-cache@v2` with **distinct `prefix-key`s** — a shared key would make them evict each other and buy nothing.

It belongs in this slice because it is what makes #121 affordable: baseline is 8m05s–8m40s uncached, and a cold Windows build of the `windows` 0.61 family plus `rusqlite` (which compiles SQLite from C) lands well past that. A 25-minute matrix is a pipeline people stop reading — and this repo has already been there: 28 consecutive red runs, switched off from 2026-07-12 to 2026-08-22 (DBSYNC-74).

## Two judgement calls to review

**`npm install` dropped from the `rust` job.** Nothing there needs it: `cargo test` compiles the crate and runs unit tests, and `tauri-build` does not require `dist/`. The old job demonstrated exactly that by running these same tests *before* the frontend build ever produced it. If that reading is wrong, this PR's own run says so — which is what this slice is for.

**Everything stays on `macos-latest`.** The `frontend` job is platform-independent and would be roughly 10x cheaper on `ubuntu-latest`, but npm's platform-specific optional packages give that its own failure mode. It does not belong in a change whose whole point is to be provably green. Worth doing later, alone.

## Stack

`desktop-tauri`

## Test Plan

- [ ] `cargo test --manifest-path apps/desktop/src-tauri/Cargo.toml` — green on the `rust` job
- [ ] `cargo clippy ... --all-targets` — green (still **not** `-D warnings`; see the comment kept at the step)
- [ ] `run-badge-diff-tests.sh` — green on `macos-app`
- [ ] `npm --workspace apps/desktop run build` — green on `frontend`
- [ ] `bundle:dev` — green on `macos-app`, appex included
- [ ] **Same set of checks as before this PR, each running exactly once.** Verified locally against the job/step inventory; the run itself is the proof.
- [ ] **Wall-clock recorded below**, against the 8m05s–8m40s baseline.

Local validation is limited by construction — GitHub Actions does not run here. What was checked locally: the YAML parses, and the inventory shows all five original checks present exactly once, none duplicated. Everything else is what this PR's own run is for.

## Evidence

Wall-clock and per-job timings to be added here once the run completes.

#DBSYNC-38

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,9 +44,18 @@ jobs:
           prefix-key: rust-checks
           workspaces: apps/desktop/src-tauri
 
-      # No `npm install` here: `cargo test` compiles the crate and runs its unit tests,
-      # and `tauri-build` does not need `dist/` to exist for that — the old single job
-      # proved it by running these tests before the frontend build step.
+      # No `npm install` here, and the reason is narrower than "the Rust tests don't need
+      # the frontend" — it is conditional, so read this before changing the command.
+      #
+      # `generate_context!` DOES embed `frontendDist`, but only outside dev mode:
+      # `tauri-build` sets `dev = !has_feature("custom-protocol")` (build.rs), and
+      # `tauri-codegen`'s `else if dev && config.build.dev_url.is_some()` branch then skips
+      # the embed entirely. So this job compiles without `dist/` because two config facts
+      # happen to hold: `custom-protocol` is off, and `devUrl` is set in tauri.conf.json.
+      #
+      # Change either — add `--all-features`, or drop `devUrl` — and this step fails with an
+      # error about a missing frontend directory that never mentions npm. Restore
+      # `npm install` and the frontend build here if that day comes.
       - name: Rust tests
         run: cargo test --manifest-path apps/desktop/src-tauri/Cargo.toml
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,14 +5,62 @@ on:
     branches: [main, master]
   pull_request:
 
+# Literals, not secrets. The app reads these through `option_env!` with fallbacks
+# (auth/oauth.rs), so CI only needs them to be present, never to be real. Hoisted to the
+# workflow level in DBSYNC-38 so every job inherits them and none can be forgotten.
+env:
+  DROPBOX_APP_KEY: dummy
+  DROPBOX_REDIRECT_URI: http://localhost:53682/callback
+
+# Split into one job per kind of check (DBSYNC-38, GitHub #120). It was a single
+# `desktop-build` job until 2026-08-25.
+#
+# The split is not cosmetic: it is what makes a Windows leg possible at all. Three of the
+# old job's steps cannot run on Windows — the Finder Sync checks shell out to `xcrun
+# swiftc`, and `bundle:dev` produces a macOS `.app`. Adding `strategy.matrix.os` to the
+# monolith would have broken all three. Slices #121 and #122 add Windows on top of this
+# shape; this change deliberately adds no platform of its own, so that a later red build
+# on Windows is attributable to Windows rather than to the restructuring.
+#
+# Everything runs on macOS for now. The frontend job is platform-independent and would be
+# roughly ten times cheaper on `ubuntu-latest`, but moving it is an optimisation with its
+# own risk (npm's platform-specific optional packages) and does not belong in a change
+# whose whole point is to be provably green first.
 jobs:
-  desktop-build:
+  # Portable Rust. Becomes a macOS + Windows matrix in #121 — the reason this ticket exists.
+  rust:
     runs-on: macos-latest
-    env:
-      # Literals, not secrets. The app reads these through `option_env!` with fallbacks
-      # (auth/oauth.rs), so CI only needs them to be present, never to be real.
-      DROPBOX_APP_KEY: dummy
-      DROPBOX_REDIRECT_URI: http://localhost:53682/callback
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
+      - uses: Swatinem/rust-cache@v2
+        with:
+          # Distinct per job, or `rust` and `macos-app` evict each other and the cache
+          # buys nothing. The runner OS is part of the key by default, which is what keeps
+          # the Windows leg from thrashing against this one later.
+          prefix-key: rust-checks
+          workspaces: apps/desktop/src-tauri
+
+      # No `npm install` here: `cargo test` compiles the crate and runs its unit tests,
+      # and `tauri-build` does not need `dist/` to exist for that — the old single job
+      # proved it by running these tests before the frontend build step.
+      - name: Rust tests
+        run: cargo test --manifest-path apps/desktop/src-tauri/Cargo.toml
+
+      - name: Rust lints
+        # Not `-D warnings`: main currently carries pre-existing warnings, and failing the
+        # build on them would put this workflow straight back to permanently red — which is
+        # how it got switched off in the first place. Clearing them is its own change;
+        # making them visible is this one. The same reasoning will apply, harder, to the
+        # Windows leg: those modules have never been linted by anything.
+        run: cargo clippy --manifest-path apps/desktop/src-tauri/Cargo.toml --all-targets
+
+  # macOS-only artefacts: the Swift appex and the `.app` bundle.
+  macos-app:
+    runs-on: macos-latest
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
@@ -21,36 +69,48 @@ jobs:
           cache: 'npm'
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
         with:
-          components: clippy
+          prefix-key: macos-app
+          workspaces: apps/desktop/src-tauri
 
       - name: Install dependencies
         run: npm install
 
-      # The four steps below are the reason this workflow is worth running at all.
-      # Until 2026-08-22 it only compiled and bundled: 214 Rust tests and the Swift
-      # checks had never run in CI once. Of the defects that reached code review on
-      # 2026-08-21, a compile-only pipeline would have caught none of them.
-
-      - name: Rust tests
-        run: cargo test --manifest-path apps/desktop/src-tauri/Cargo.toml
-
-      - name: Rust lints
-        # Not `-D warnings`: main currently carries 4 pre-existing warnings, and failing
-        # the build on them would put this workflow straight back to permanently red —
-        # which is how it got switched off in the first place. Clearing them is its own
-        # change; making them visible is this one.
-        run: cargo clippy --manifest-path apps/desktop/src-tauri/Cargo.toml --all-targets
-
       - name: Finder Sync extension checks
+        # Compiles the real BadgeDiff.swift with `swiftc` and runs its checks. macOS-only
+        # by construction, and the only automated coverage the extension has.
         run: apps/desktop/native/macos/FinderSyncExtension/Tests/run-badge-diff-tests.sh
-
-      - name: Frontend build and typecheck
-        run: npm --workspace apps/desktop run build
 
       - name: Build macOS app bundle
         # `bundle:dev` (--bundles app), not `tauri build`. The full build also produces a
         # DMG, a stage that has never executed in CI even once because every run died
         # earlier. Producing an installer belongs to the release workflow (GitHub #85);
         # a PR check only needs to prove the app links.
+        #
+        # This step also builds the Swift appex: `beforeBuildCommand` runs
+        # `scripts/tauri-before-build.mjs`, which calls `build:finder-sync` on darwin.
+        # That is why it must never be matrixed onto another OS.
         run: npm --workspace apps/desktop run bundle:dev
+
+  # TypeScript typecheck + production bundle. Portable, so it runs exactly once.
+  frontend:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Frontend build and typecheck
+        # `npm run build` is `tsc && vite build`, so this is the typecheck too.
+        #
+        # It is knowingly duplicated: `tauri-before-build.mjs` runs the same build
+        # unconditionally, so `bundle:dev` above repeats it (and `bundle:win` will, in
+        # #122). At roughly half a second it is not worth engineering away — recorded here
+        # so it is not later mistaken for a defect.
+        run: npm --workspace apps/desktop run build


mbsanchez:
## CI evidence — the AC this slice could not meet locally

Run [32827513001](https://github.com/mbsanchez/DropboxSync/actions/runs/32827513001), all green.

| Job | Result | Duration |
|---|---|---|
| `rust` | pass | 5m07s |
| `macos-app` | pass | 5m09s |
| `frontend` | pass | **24s** |
| **Total wall-clock** | **success** | **5m12s** |

Baseline before this PR: **8m05s–8m40s** for the single job.

So roughly a third faster — and that is with a **cold cache**. This run had nothing to restore; it only populated. The comparison that will matter is the next run on this branch, which should be the first to actually benefit.

The saving is not really the cache anyway, it is the parallelism: three jobs run at once, so wall-clock is now the slowest job rather than the sum of every step. `frontend` finishing in 24s is the clearest illustration — that signal used to be buried at position five in an 8-minute serial queue.

### The two judgement calls, both settled by this run

**`npm install` removed from the `rust` job** — correct. `cargo test` and clippy passed with no `node_modules` present at all, confirming `tauri-build` does not need `dist/` for a plain compile. It was cargo-culted from the old job.

**All jobs on `macos-latest`** — stands, and `frontend` at 24s makes the ubuntu question less interesting than it looked. Moving it would save runner minutes, not developer time. Still worth doing on its own someday; still not worth the npm-optional-packages risk inside this slice.

### AC status

- [x] Same set of checks as before, each running exactly once
- [x] Rust caching added (both Rust-touching jobs, distinct prefixes)
- [x] Wall-clock recorded: 8m05s–8m40s → 5m12s
- [x] `bundle:dev` stayed on macOS and built the appex
- [x] No new operating system in this slice

Ready for review. #121 is unblocked by this being observed green — which was the entire point of doing the restructuring first.

mbsanchez (commented):
## CTO review — Slice 1 of DBSYNC-38

**Verdict: APPROVE.** No critical issues. Every acceptance criterion is met and I verified each one rather than taking the PR body's word for it. The findings below are non-blocking, but three of them land on the cache — which is the part of this slice that #121 is depending on, and the part this run cannot prove.

Stack profile: `desktop-tauri` (`CLAUDE.md` override, corroborated by `apps/desktop/src-tauri/tauri.conf.json`), `qa_strategy: manual`. CI is the only automated gate this project has, so the review is weighted toward *which failures this workflow would actually catch*.

---

### AC verification

I parsed both YAML files and diffed the `run:` inventory rather than eyeballing the diff:

| AC | Result |
|---|---|
| Split into `rust` / `macos-app` / `frontend` | Confirmed |
| Every check still runs, none twice | **Confirmed.** 5 checks in, 5 out. Zero missing, zero added. The only duplicated `run:` is `npm install` (`macos-app`, `frontend`) — setup, not a check |
| Rust caching added | Confirmed, both Rust-touching jobs |
| Wall-clock recorded | Confirmed, and **understated** — see below |
| `bundle:dev` pinned to macOS, appex included | Confirmed: `tauri-before-build.mjs` gates `build:finder-sync` on `process.platform === "darwin"` |
| No new OS | Confirmed, all three `runs-on: macos-latest` |

**The wall-clock claim is conservative, not favourable.** The quoted 8m05s–8m40s baseline is lower than what main has actually been doing. Measuring the same way (`run_started_at` → `updated_at`) the last three main runs are 8m19s, 9m20s and 10m24s, against this PR's 5m12s. The honest headline is closer to *half*, not *a third*.

**Workflow-level `env:` reaches all three jobs.** No job declares its own `env`, so nothing shadows the hoisted block. `DROPBOX_APP_KEY` is consumed at compile time by `option_env!` (`auth/oauth.rs:20,27`), and the `rust` job compiles, so it needs it and gets it.

---

### Judgement call 1 — `npm install` dropped from `rust`

**The conclusion is right. The stated reason is not the operative one, and that matters.**

Traced through the vendored sources:

- `tauri-2.10.3/build.rs:253` — `let dev = !has_feature("custom-protocol");`
- `tauri-codegen-2.5.5/src/context.rs:180` — `} else if dev && config.build.dev_url.is_some() {` → **no assets embedded, `../dist` is never read.**

`cargo test` and `cargo clippy` don't enable `custom-protocol`, and `tauri.conf.json` sets `devUrl: http://localhost:1420`. Both conditions hold, so `generate_context!` (`lib.rs:490`) skips the embed. `bundle:dev` is `tauri build`, which *does* enable `custom-protocol` — hence release embeds `../dist`, hence `macos-app` genuinely needs `npm install`.

So it did not pass for a reason you failed to identify, but it passes for a **conditional** reason, and the comment states an unconditional one. `tauri-build` indeed doesn't need `dist/`; `generate_context!` very much does, and only two config facts are keeping it quiet. Two named triggers would break the `rust` job with an error that mentions embedded assets and says nothing about npm:

1. Anyone adding `--all-features` or `--features custom-protocol` to the rust job. Worth guarding explicitly: the generic `rust` stack profile's canonical command is `cargo test --all-features`, so this is a plausible future edit, not a hypothetical.
2. Anyone removing `devUrl` from `tauri.conf.json`.

Suggest the comment name the condition:

```yaml
      # No `npm install` here. `cargo test`/`clippy` do not enable tauri's
      # `custom-protocol` feature, so `dev` is true, and with `devUrl` set in
      # tauri.conf.json `generate_context!` skips embedding `frontendDist`
      # (tauri-codegen context.rs). `tauri build` (macos-app) enables it, embeds
      # `../dist`, and does need the install.
      # Adding --all-features or --features custom-protocol here breaks this.
```

Second-order consequence worth recording somewhere: the `rust` job therefore **never compiles the crate in the configuration that ships**. On macOS `macos-app` covers that gap. When #121 adds the Windows leg to `rust` only, Windows is compiled in dev cfg alone until #122's `bundle:win` lands. There is currently no `#[cfg(dev)]` or `#[cfg(custom_protocol)]` code under `src/`, so today's exposure is genuinely zero — the point is to know that it is zero by accident rather than by design.

### Judgement call 2 — everything on `macos-latest`

**Agree, and the usual counter-argument doesn't even apply here.** The repo is public, so Actions minutes are free on all runners. Moving `frontend` to `ubuntu-latest` saves no money and 24 seconds of a runner that nothing is waiting on. Against that: npm's platform-specific optional packages are a real failure mode with a canonical instance (`@rollup/rollup-linux-x64-gnu` going missing on a lockfile install). Taking that risk inside a slice whose entire premise is *provably green* would be a bad trade at any price, and at zero price it isn't a trade at all.

One correction to the framing rather than the decision: the PR presents ubuntu as deferring the npm-optional-deps risk, but #122 runs `npm install` on `windows-latest` and meets the same class of failure regardless. What you are deferring is one platform, not the failure mode. That is still the right call — meet it once, on a platform you actually ship.

---

### Warnings

These are all about whether the cache delivers what #121 is counting on. As the PR itself says, this run had a cold cache, so the cache is currently an **unmeasured** claim. The 5m12s is parallelism.

**W1 — `save-if` is unbounded; the cache budget will be spent on caches nobody can read.** Both `rust-cache` steps default to `save-if: true`, so every PR branch writes. GitHub gives the repo 10 GB total with LRU eviction, and a cache written on a PR ref is readable *only by that PR* — not by main, not by other PRs. Those writes therefore buy one branch a warm second run while evicting main's caches, which are the ones every PR actually restores from. Two macOS caches today (one debug/clippy target dir, one release); #121 adds Windows, where `rusqlite` compiles SQLite from C and the `windows` 0.61 family is large. This plausibly exceeds 10 GB across active branches, at which point the cache quietly stops hitting and nobody gets a signal that it stopped. Fix, both steps:

```yaml
          save-if: ${{ github.ref == 'refs/heads/main' }}
```

**W2 — `cache-on-failure` defaults to false.** Every retry after a red run is a cold ~5 minute build. This repo is about to add a Windows leg that will be red several times before it is green; that is precisely when you want the compile output kept. Add `cache-on-failure: true` to both steps.

**W3 — the cache key does not track `DROPBOX_APP_KEY`, and `option_env!` does not force a rebuild.** rust-cache hashes only env vars prefixed `CARGO`, `CC`, `CFLAGS`, `CXX`, `CMAKE`, `RUST` into its key (`config.ts`, `envPrefixes`). `DROPBOX_APP_KEY` matches none of them. Separately, `option_env!` does not emit `cargo:rerun-if-env-changed`, so cargo will not rebuild on a change either. Today the value is the literal `dummy` and nothing moves, so the impact is nil — but this hazard **did not exist before this PR**, because there was no cache to serve a stale artifact from. The day a real key arrives from secrets for the release workflow (#85), a warm cache can hand back a binary with `dummy` compiled into it, and the build will be green. Cheap insurance now, on both steps:

```yaml
          env-vars: DROPBOX_APP_KEY DROPBOX_REDIRECT_URI
```

---

### Suggestions

**S1 — `npm install` → `npm ci`.** `package-lock.json` (lockfileVersion 3) is committed and `cache: 'npm'` is already configured. `npm install` silently reconciles a drifted lockfile and *cannot fail on it*; `npm ci` fails. By your own standard — a check that cannot fail is not evidence — that is the difference between a check and a formality, and it is also faster. Caveat, and it is the reason this is a suggestion and not a warning: if the lockfile is currently drifted, this turns the pipeline red, which is the one thing this slice is built to avoid. Verify locally, then land it as its own change.

**S2 — the distinct `prefix-key`s are correct, but the comment justifying them is wrong twice.** Keep the keys; fix the comment, because a wrong mechanism in a comment is exactly what gets the right line deleted later.

- rust-cache's `add-job-id-key` defaults to `"true"` (`action.yml`), so `GITHUB_JOB` — `rust` vs `macos-app` — was already in the key. These would have been distinct with no `prefix-key` at all.
- Same-key writes do not "evict each other." Actions cache keys are immutable: the second save no-ops with *Cache already exists*. The real hazard of a shared key is milder and different — `rust` caches a debug/clippy target dir and `macos-app` a release one, so they would churn over one key with each restoring artifacts it can't use.

**S3 — the one load-bearing claim in that comment checks out, and survives #121.** `config.ts` appends `-${os.type()}-${os.arch()}` to the key, so a `windows-latest` matrix leg of `rust` gets a distinct key from the macOS leg despite sharing `GITHUB_JOB`. Nothing to change; recording it because it is the specific property the next slice rests on.

**S4 — no `permissions:` block.** Public repo, so the default `GITHUB_TOKEN` grants are broad, and every third-party action inherits them. This PR newly introduces `Swatinem/rust-cache@v2`, and `dtolnay/rust-toolchain@stable` resolves a mutable *branch*, not a tag or SHA. Nothing here writes to the repo. Add `permissions: contents: read` at the workflow level, right beside the `env:` block you just hoisted — same reasoning, same place.

**S5 — no `concurrency:` group.** `on.push` is main-only so PR branches don't double-run; the real case is a force-push mid-run, which now orphans three macOS jobs instead of one. Public repos cap at 5 concurrent macOS jobs and #121/#122 push toward that ceiling.

```yaml
concurrency:
  group: ci-${{ github.ref }}
  cancel-in-progress: true
```

---

### On the `frontend` job's right to exist

Worth stating, since your standard invites the challenge: `frontend` cannot fail independently. `bundle:dev` runs `npm run build` unconditionally through `tauri-before-build.mjs`, so any failure `frontend` catches, `macos-app` also catches. Its value is entirely latency and attribution — 24 seconds instead of five minutes, and a failure that is unambiguously frontend because that job has no Rust toolchain at all. That is a legitimate reason to keep a job, and it gets stronger in #122, when the same breakage would otherwise surface as a Windows failure. The PR already flags the duplicated build so it isn't later mistaken for a defect; the *job's* justification deserves the same one-line treatment, because "runs the same command as another job" is what someone deletes during a future cleanup.

---

**Recommendation: merge.** Fold W1–W3 into #121 before the Windows leg lands, since that slice's cost argument is the thing they protect. S1, S4 and S5 are independent and can go whenever.

mbsanchez:
## Review findings — one applied here, the rest carried to #121

Thanks for tracing the `generate_context!` path rather than accepting the passing build as proof.

### Applied now (9fa8fc0, comment only)

**The `npm install` comment was misleading and is fixed.** It claimed `tauri-build` does not need `dist/`. It does — the embed is skipped only because `custom-protocol` is off and `devUrl` is set. A conditional reason stated as an unconditional one is exactly the kind of comment that sends the next person looking in the wrong place: the failure would name a missing frontend directory and never mention npm. The comment now names both conditions and says what to restore if either changes.

That is the finding I most wanted from this review — the conclusion was right and my stated reason was not, which no green build would ever have revealed.

### Corrected framing, accepted

**The repo is public, so Actions minutes are free.** My cost argument for keeping `frontend` on macOS was therefore worthless. The decision stands on the npm optional-deps risk alone — and your point lands: #122 runs `npm install` on `windows-latest` anyway, so this defers one platform rather than avoiding the failure mode.

### Carried to #121, deliberately not fixed here

The three cache warnings all concern behaviour this cold run left **unproven**, and #121 is where the cache's affordability argument actually gets tested:

- unbounded `save-if` spending the 10 GB budget on PR-scoped caches while evicting main's
- `cache-on-failure` false, right before a Windows leg expected to be red repeatedly
- the cache key not tracking `DROPBOX_APP_KEY` while `option_env!` will not force a rebuild — nil impact at `dummy`, a green-but-wrong binary the day a real key is wired

Folding them in here would mean changing the cache in the same PR that introduces it, with no way to tell which behaviour came from which change. They land in #121 with a Windows leg to exercise them.

`npm ci` over `npm install` is a genuinely good call — a check that can fail versus one that cannot — and goes to #121 with them.

Ready for QA.

mbsanchez:
## QA Result: PASSED ✅

QA run against the **latest commit** (`9fa8fc0`), not the one the review saw — the comment fix landed after it, and it has its own run.

`qa_strategy` for this stack is `manual`, and it was deliberately not followed here: this slice changes no source file, so there is nothing to validate in an app window. Every AC is observable in the pipeline run itself. Recorded as a deviation with its reason, rather than staged as theatre.

### AC verification

| AC | Evidence |
|---|---|
| Splits into `rust` / `macos-app` / `frontend` | The run shows exactly those three jobs, all green |
| Every check still runs, **none twice** | Step inventory below — from the run, not the YAML |
| Rust caching added | Proven by the warm-run measurement below |
| `bundle:dev` pinned to macOS | Runs only in `macos-app`; the appex builds through `beforeBuildCommand` |
| Wall-clock recorded | Below |
| No new OS | Three jobs, all `macos-latest` |

### Check inventory, taken from the executed run

```
1  Rust tests                     [rust]
1  Rust lints                     [rust]
1  Finder Sync extension checks   [macos-app]
1  Build macOS app bundle         [macos-app]
1  Frontend build and typecheck   [frontend]
```

Five checks, one occurrence each, all `success`. The old single job on `main` carries the same five, verified by parsing `git show main:.github/workflows/ci.yml`. Nothing lost, nothing duplicated.

### Wall-clock — and the cache is no longer unproven

| Run | Total | `rust` | `macos-app` | `frontend` |
|---|---|---|---|---|
| `main` baseline (single job) | 8m19s – 10m24s | — | — | — |
| `a4919ce` — **cold** cache | **5m12s** | 5m07s | 5m09s | 24s |
| `9fa8fc0` — **warm** cache | **2m23s** | 1m14s | 2m19s | 26s |

The review noted the cache was unproven because the only run available was cold. The second run settles it: `rust` drops from 5m07s to **1m14s**, roughly 4x.

That matters beyond this PR. The whole affordability argument for #121 — that a cold Windows build of the `windows` 0.61 family plus `rusqlite` compiling SQLite from C would otherwise be unaffordable — rested on the cache working. It is now measured rather than assumed.

Against the baseline, the pipeline went from ~9 minutes to **2m23s warm**.

### Not covered here

The cache warnings from the review (unbounded `save-if`, `cache-on-failure`, key not tracking `DROPBOX_APP_KEY`) are behaviours this slice cannot exercise. They are tracked on #121, where a Windows leg will.

Ready for PO approval.

PR #124, opened 2026-08-25
ci(#DBSYNC-38): compile and test the Windows code, for the first time ever

mbsanchez:
Slice 2 of 3 for [DBSYNC-38](https://manuelbsan.atlassian.net/browse/DBSYNC-38) — GitHub #121. Blocked by #120, which is merged.

**This is the slice the ticket exists for.**

## What has never been checked until now

Five modules sit behind `#[cfg(windows)]`:

```
cloud_filter          2024 lines / 92 KB
windows_file_assoc
windows_identity
windows_shell_menu
windows_startup
```

plus a Windows-only dependency block (`winreg`, `windows-sys`, `windows` 0.61 with `Win32_Storage_CloudFilters`, `Storage_Provider`, `Win32_System_IO`).

Roughly 4000 lines that **no pipeline has ever compiled**, containing ~22 unit tests that have only ever run on one machine. `cargo clippy --all-targets` has never linted a line of the CfAPI layer.

Concretely: until this merges, a PR can delete a function `cloud_filter.rs` calls and CI reports green — on the largest single module in the codebase.

## The change

`rust` becomes a matrix over `macos-latest` and `windows-latest`, with **`fail-fast: false`**.

That flag is load-bearing, not taste. On the default, a macOS failure cancels the Windows job — hiding the result being bought at exactly the moment something is already known to be wrong.

## Review findings from #120, folded in here

#120 deferred these deliberately: changing a cache in the same PR that introduces it makes later behaviour unattributable. This is where they get exercised.

- **`save-if` restricted to `main`.** A 10 GB budget shared by every open PR evicts main's cache — the one every new PR starts from — to store caches no other branch can read.
- **`cache-on-failure: true`.** This leg is *expected* to go red while the Windows code compiles for the first time. Discarding the cache on each failure would make every retry a cold build of the `windows` 0.61 family plus `rusqlite` compiling SQLite from C. This is the warning most likely to bite this very slice.
- **`DROPBOX_APP_KEY` in the cache key.** `option_env!` does not force a rebuild when it changes, so a cached artifact can embed a stale value with nothing reporting it. Nil impact at `dummy`; a green-but-wrong binary the day the release workflow wires a real key.
- **`npm ci` replaces `npm install`** — it fails when the lockfile disagrees with `package.json` instead of quietly resolving something new. Verified against the current lockfile with `npm ci --dry-run` before committing, so this does not ship a red pipeline for a self-inflicted reason.

Plus `permissions: contents: read` and a `concurrency` group: nothing here writes to the repo, and a superseded push should not keep burning a Windows runner nobody will read.

## Stack

`desktop-tauri`

## Test Plan

- [ ] Both matrix legs green — or, if Windows is red, **that is a finding, not a setback** (see below)
- [ ] **Windows test count quoted below and higher than the macOS count.** That difference is the proof the `cfg(windows)` tests actually ran, rather than the job merely passing
- [ ] Clippy runs on both legs, still without `-D warnings`
- [ ] **Mutation check**: on a throwaway branch, break something in `cloud_filter.rs`, confirm the Windows leg goes red while macOS stays green, quote it here, delete the branch. A job nobody has watched fail is not yet known to work
- [ ] Total wall-clock recorded, warm and cold

## If the Windows leg is red

**Do not fix it in this PR.** File the failure as its own bug and let this land the job. Folding "fix what the check found" into "add the check" makes both unreviewable, and the plan on the Jira ticket calls this out in advance. Binding drift, a stale call, or tests assuming `/` path separators are all plausible after ~4000 lines went uncompiled this long.

## Evidence

Test counts, timings and the mutation check to be added here once the run completes.

#DBSYNC-38

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,17 @@ on:
     branches: [main, master]
   pull_request:
 
+# Read-only by default: nothing in this workflow writes to the repo, and a token that
+# cannot write is one that cannot be abused by a compromised action.
+permissions:
+  contents: read
+
+# One run per ref. Pushing twice to a PR should not leave the first run burning a Windows
+# runner nobody will read.
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
 # Literals, not secrets. The app reads these through `option_env!` with fallbacks
 # (auth/oauth.rs), so CI only needs them to be present, never to be real. Hoisted to the
 # workflow level in DBSYNC-38 so every job inherits them and none can be forgotten.
@@ -19,17 +30,28 @@ env:
 # old job's steps cannot run on Windows — the Finder Sync checks shell out to `xcrun
 # swiftc`, and `bundle:dev` produces a macOS `.app`. Adding `strategy.matrix.os` to the
 # monolith would have broken all three. Slices #121 and #122 add Windows on top of this
-# shape; this change deliberately adds no platform of its own, so that a later red build
-# on Windows is attributable to Windows rather than to the restructuring.
+# shape. Splitting first is what made a later red build on Windows attributable to
+# Windows rather than to the restructuring.
 #
-# Everything runs on macOS for now. The frontend job is platform-independent and would be
-# roughly ten times cheaper on `ubuntu-latest`, but moving it is an optimisation with its
-# own risk (npm's platform-specific optional packages) and does not belong in a change
-# whose whole point is to be provably green first.
+# `rust` now runs on both platforms (#121). `macos-app` and `frontend` stay on macOS:
+# the first is macOS-only by construction, the second is platform-independent and would
+# be cheaper on `ubuntu-latest` — an optimisation with its own risk (npm's
+# platform-specific optional packages), worth doing on its own and not here.
 jobs:
-  # Portable Rust. Becomes a macOS + Windows matrix in #121 — the reason this ticket exists.
+  # Portable Rust, on every platform that ships.
   rust:
-    runs-on: macos-latest
+    # THE POINT OF DBSYNC-38 (GitHub #121). Until this leg existed, everything behind
+    # `#[cfg(windows)]` — five modules, ~4000 lines, `cloud_filter.rs` alone at 2024 —
+    # had never been compiled by any pipeline, and the ~22 unit tests inside them had
+    # only ever run on the maintainer's own machine. A PR could delete a function
+    # `cloud_filter.rs` calls and CI would report green.
+    strategy:
+      # NOT the default. With fail-fast on, a macOS failure cancels the Windows job —
+      # cancelling precisely the signal this leg exists to buy. Load-bearing, not taste.
+      fail-fast: false
+      matrix:
+        os: [macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
       - name: Install Rust
@@ -39,10 +61,24 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           # Distinct per job, or `rust` and `macos-app` evict each other and the cache
-          # buys nothing. The runner OS is part of the key by default, which is what keeps
-          # the Windows leg from thrashing against this one later.
+          # buys nothing. rust-cache appends the runner OS and arch to the key, so the
+          # two matrix legs get separate caches without any help from us.
           prefix-key: rust-checks
           workspaces: apps/desktop/src-tauri
+          # Only the default branch writes. A 10 GB budget shared by every open PR would
+          # evict main's cache — the one every new PR starts from — to store caches that
+          # no other branch can even read.
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+          # Keep the cache from a red run. This leg is expected to go red while the
+          # Windows code is compiled for the first time, and discarding the cache on
+          # each failure would make every retry a cold build of the `windows` 0.61
+          # family plus `rusqlite` compiling SQLite from C.
+          cache-on-failure: true
+          # `option_env!` does not force a rebuild when this changes, so without it in the
+          # key a cached artifact can embed a stale value and nothing reports it. Harmless
+          # at `dummy`; a green-but-wrong binary the day the release workflow wires a real
+          # key.
+          env-vars: CARGO CC CFLAGS CXX CMAKE RUST DROPBOX_APP_KEY
 
       # No `npm install` here, and the reason is narrower than "the Rust tests don't need
       # the frontend" — it is conditional, so read this before changing the command.
@@ -82,9 +118,15 @@ jobs:
         with:
           prefix-key: macos-app
           workspaces: apps/desktop/src-tauri
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+          cache-on-failure: true
+          env-vars: CARGO CC CFLAGS CXX CMAKE RUST DROPBOX_APP_KEY
 
       - name: Install dependencies
-        run: npm install
+        # `ci`, not `install`: it fails when package-lock.json disagrees with package.json
+        # instead of quietly resolving something new. A check that cannot fail is not a
+        # check, and on a runner the lockfile is the whole point.
+        run: npm ci
 
       - name: Finder Sync extension checks
         # Compiles the real BadgeDiff.swift with `swiftc` and runs its checks. macOS-only
@@ -113,7 +155,10 @@ jobs:
           cache: 'npm'
 
       - name: Install dependencies
-        run: npm install
+        # `ci`, not `install`: it fails when package-lock.json disagrees with package.json
+        # instead of quietly resolving something new. A check that cannot fail is not a
+        # check, and on a runner the lockfile is the whole point.
+        run: npm ci
 
       - name: Frontend build and typecheck
         # `npm run build` is `tsc && vite build`, so this is the typecheck too.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,11 +10,16 @@ on:
 permissions:
   contents: read
 
-# One run per ref. Pushing twice to a PR should not leave the first run burning a Windows
-# runner nobody will read.
+# One run per ref, but cancellation ONLY on pull requests. Pushing twice to a PR should
+# not leave the first run burning a Windows runner nobody will read.
+#
+# On a push to main, cancelling would discard the only automated verification a merged
+# commit ever gets: two merges inside six minutes and the first commit is never tested on
+# any platform, silently. With a manual QA strategy this pipeline is the sole automated
+# gate, so a merged commit must always finish its run.
 concurrency:
   group: ci-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 # Literals, not secrets. The app reads these through `option_env!` with fallbacks
 # (auth/oauth.rs), so CI only needs them to be present, never to be real. Hoisted to the
@@ -52,6 +57,10 @@ jobs:
       matrix:
         os: [macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
+    # A hang, not an assertion, is the characteristic failure of Win32 I/O-callback code
+    # under test — and the default here is 360 minutes on a runner billed at 2x. Twenty is
+    # generous against a 6m25s cold Windows run.
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v4
       - name: Install Rust
@@ -69,16 +78,22 @@ jobs:
           # evict main's cache — the one every new PR starts from — to store caches that
           # no other branch can even read.
           save-if: ${{ github.ref == 'refs/heads/main' }}
-          # Keep the cache from a red run. This leg is expected to go red while the
-          # Windows code is compiled for the first time, and discarding the cache on
-          # each failure would make every retry a cold build of the `windows` 0.61
-          # family plus `rusqlite` compiling SQLite from C.
+          # Two things, and NOT the one the first draft of this comment claimed. It does
+          # not rescue red PR runs: `save-if` gates the save body outright, so on a PR
+          # branch nothing is written whatever this says. What it does do is let main keep
+          # a cache from a failed run, and — because it drives the action's `post-if` — let
+          # the save step run at all when a job is cancelled.
           cache-on-failure: true
-          # `option_env!` does not force a rebuild when this changes, so without it in the
-          # key a cached artifact can embed a stale value and nothing reports it. Harmless
-          # at `dummy`; a green-but-wrong binary the day the release workflow wires a real
-          # key.
-          env-vars: CARGO CC CFLAGS CXX CMAKE RUST DROPBOX_APP_KEY
+          # Both values `auth/oauth.rs` reads through `option_env!` — DROPBOX_APP_KEY and
+          # DROPBOX_REDIRECT_URI. `option_env!` does not force a rebuild when they change,
+          # so without them in the key a cached artifact can embed a stale value and
+          # nothing reports it. Harmless at `dummy`; a green-but-wrong binary the day the
+          # release workflow wires a real key.
+          #
+          # The prefix is enough: rust-cache matches with `startsWith`, and this list is
+          # ADDED to its built-ins (CARGO CC CFLAGS CXX CMAKE RUST) rather than replacing
+          # them.
+          env-vars: DROPBOX
 
       # No `npm install` here, and the reason is narrower than "the Rust tests don't need
       # the frontend" — it is conditional, so read this before changing the command.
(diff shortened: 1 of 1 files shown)


mbsanchez:
## Evidence — all four Test Plan items met

### Both legs green, and the Windows tests actually ran

| Job | Result | Time |
|---|---|---|
| `rust (macos-latest)` | ✅ | 4m24s |
| **`rust (windows-latest)`** | ✅ | 6m25s |
| `macos-app` | ✅ | 6m27s |
| `frontend` | ✅ | 21s |

Total **6m30s**, cold cache, two platforms — against the 8m19s–10m24s the single uncached job cost.

**Windows compiled clean on the first attempt.** After months with no pipeline touching those ~4000 lines, there is no binding drift and no stale call. That was the plan's main risk and it did not materialise.

### Test counts — the proof the gated tests executed

```
rust (macos-latest)    running 197 tests   →  test result: ok. 197 passed
rust (windows-latest)  running 214 tests   →  test result: ok. 214 passed
```

**17 tests exist only behind `cfg(windows)`** and have now run outside the maintainer's machine for the first time. A green job would have proven nothing on its own; the difference is what proves they ran.

(The ticket estimated ~22. The measured number is 17 — recorded as measured rather than as estimated.)

### Mutation check — the job is now known to be able to fail

Throwaway branch, deliberate type error in `cloud_filter.rs`, run in #125 (closed and deleted):

| Job | Result |
|---|---|
| **`rust (windows-latest)`** | ❌ **FAIL** |
| `rust (macos-latest)` | ✅ pass |
| `macos-app` | ✅ pass |
| `frontend` | ✅ pass |

```
error[E0308]: mismatched types
113 |     let _mutation_check: u32 = "DBSYNC-38: deliberate break, ...";
    |                          ---   ^^^^ expected `u32`, found `&str`
```

**The part worth pausing on:** with that break in place, `cargo check` on this Mac reports `Finished`. The error is invisible to the only machine this project is developed on. That is the blind spot the ticket describes, reproduced deliberately — and the new leg catches it.

### Bonus: `fail-fast: false` is verified, not merely configured

`macos-app` ran 7m13s and **completed**, though Windows had already failed at 4m16s. Nothing was cancelled. On the default, that Windows failure would have killed the other jobs — and in the inverse case, hidden the Windows result entirely.

Ready for review.

mbsanchez (commented):
## CTO review — Slice 2 (DBSYNC-38 / #121)

Stack: `desktop-tauri`. Reviewed `.github/workflows/ci.yml` only; verified every claim against run logs and against the `Swatinem/rust-cache@v2` source rather than its README.

**Verdict: request changes.** Two one-line fixes. The substance of the slice is right and the evidence behind it is real — this is the first time the CfAPI layer has ever been compiled by a machine that is not yours, and the mutation check is the part that makes it evidence rather than decoration.

### What I independently confirmed

- Run `32884595997`: `rust (macos-latest)` 197 passed, `rust (windows-latest)` 214 passed, `0 ignored` on both legs. Total 6m30s.
- Mutation check, run `32885387815` on `throwaway/DBSYNC-38-mutation-check`: `rust (windows-latest)` **failure** with `error[E0308]: mismatched types` at 18:47:15, `rust (macos-latest)` **success**. The Windows leg can fail, and it fails for Windows-only reasons. AC met.
- `save-if` evaluated correctly on a PR branch — both legs logged `save-if: false`, `No cache found.`, and no save in `Post Run Swatinem/rust-cache@v2`.
- Clippy on Windows surfaced 5 lib warnings vs 4 on macOS. ~4000 lines of never-linted code cost one net new warning. Good news, and it retires the "the first run will surface a wall of warnings" risk from #121.

---

## Critical

### 1. `DROPBOX_REDIRECT_URI` is missing from the cache key — the defect you fixed is only half fixed

`auth/oauth.rs` embeds **two** values through the same mechanism:

```
auth/oauth.rs:20:pub(crate) const DROPBOX_APP_KEY: &str = match option_env!("DROPBOX_APP_KEY") {
auth/oauth.rs:27:pub const DROPBOX_REDIRECT_URI: &str = match option_env!("DROPBOX_REDIRECT_URI") {
```

The comment at `ci.yml:77-80` — "`option_env!` does not force a rebuild when this changes, so without it in the key a cached artifact can embed a stale value and nothing reports it" — is true verbatim of `DROPBOX_REDIRECT_URI`, which is set two lines above in the same `env:` block and is *not* in `env-vars`. A change to the redirect URI produces exactly the green-but-wrong binary the comment says it is preventing.

Practical impact today is nil (both are constants in this file). I am still calling it critical because the alternative is a five-line comment that documents a guarantee the config does not provide, in a repo whose review standard is that a check which cannot fail is not evidence. Someone will read that comment as authoritative.

Both share a prefix, and `env-vars` entries are matched with `key.startsWith(prefix)` (`config.ts:120`), so one word covers both:

```yaml
env-vars: DROPBOX
```

Apply to all three `rust-cache` blocks (`ci.yml:81`, `ci.yml:123`).

### 2. `cancel-in-progress: true` should be scoped to pull requests

Your worry — that cancelling main runs starves the cache every PR restores from — **is not real**, and I want to give you the reasoning that would have disproved it, because you were right to look here:

- The last push in any burst is by definition never cancelled, and it saves. Starvation would require an unbroken chain of pushes to main spaced closer than a full run (~6.5 min) indefinitely. Your merge history shows nothing remotely like that.
- Independently, the cancelled runs still attempt their save. `action.yml` declares `post-if: "success() || env.CACHE_ON_FAILURE == 'true'"`, and `restore.ts:29` does `core.exportVariable("CACHE_ON_FAILURE", cacheOnFailure)`. `success()` is false on cancellation, but the env var is `'true'`, so the post step runs anyway — subject to the cancellation grace window, which a save may or may not fit inside. `cache-on-failure: true` is load-bearing here for a reason your comment does not state.

**But there is a worse bug at the same spot.** `cancel-in-progress` on a *push to main* throws away the only automated verification a merged commit will ever get. Two merges inside six minutes and commit A is never tested on any platform — no macOS result, no Windows result, and no record that anything was skipped. `qa_strategy` is manual, so there is no second gate to catch it. A repo that spent 2026-07-12 to 2026-08-22 at 28 runs / 28 failures should not be adding a mechanism that silently declines to test commits that are already on the default branch.

Cancelling redundant *PR* runs is unambiguously right; cancelling *main* runs is not the same trade.

```yaml
concurrency:
  group: ci-${{ github.ref }}
  # Supersede PR runs — a force-push should not keep burning a Windows runner nobody
  # will read. Never supersede pushes to main: that would discard the only automated
  # verification a merged commit ever gets, and `qa_strategy` here is manual.
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

The group itself is fine: `github.ref` is `refs/pull/N/merge` for PR events and `refs/heads/main` for pushes, so PR runs and main runs cannot collide, and since `on.push.branches` is `[main, master]` a PR branch push never double-triggers.

---

## Warnings

### 3. The `cache-on-failure` comment is wrong about where the flag does anything

Your question 2, answered from source: **`save-if` and `cache-on-failure` do not compose the way the comment claims.** They AND, and `save-if` wins.

`save.ts:18-22`:

```ts
const save = core.getInput("save-if").toLowerCase() || "true";
if (!(cacheProvider.cache.isFeatureAvailable() && save === "true")) {
  return;
}
```

`cache-on-failure` only ever controls `post-if` — whether the save *process starts*. `save-if` controls whether it *does anything*. On a PR branch `save-if` is false, so the post step spawns and returns immediately. `cache-on-failure: true` is a no-op on exactly the branches where red runs are expected, which is the case the comment at `ci.yml:72-75` is written about.

The comment's premise is wrong a second time over: "discarding the cache on each failure would make every retry a cold build" cannot happen on a PR branch regardless, because retries restore *main's* cache and a PR's failures never touch it.

Keep the flag — it has two real jobs, neither of them stated:
- on `main`, a failed default-branch run would otherwise skip the save and leave every subsequent PR cold;
- it is what lets the post step run at all on a cancelled run (see finding 2).

Rewrite the comment to those two reasons. As written it will teach the next reader something false about a cache they are trying to debug.

### 4. No `timeout-minutes` on any job

Nothing in this file bounds a job, so the default is **360 minutes**, and Windows runners bill at a 2x multiplier.

This is not hypothetical for this particular slice. The code that just started being exercised is a CfAPI layer built on Win32 I/O callbacks and handles; the characteristic failure of that kind of code under test is a *hang*, not an assertion. A hung `cargo test` on `windows-latest` burns six hours at 2x before GitHub kills it, and reports nothing useful when it does.

Judged by which failures it would actually catch: this config handles a Windows hang worse than any other outcome. Both legs currently run in 4-6.5 minutes cold.

```yaml
  rust:
    timeout-minutes: 20
```

Worth adding to `macos-app` too (6m27s today).

---

## Suggestions

### 5. `env-vars` — nothing dropped, but six of the seven entries are dead weight

Your question 3, answered from source. `env-vars` is **additive, not a replacement** (`config.ts:111-112`):

```ts
const envPrefixes = ["CARGO", "CC", "CFLAGS", "CXX", "CMAKE", "RUST"];
envPrefixes.push(...core.getInput("env-vars").split(/\s+/).filter(Boolean));
```

Your hand-written list is exactly the built-in list plus one, so you dropped nothing — verified, not assumed. But the six repeated defaults are inert, and spelling them out invites a future reader to treat that line as the complete set and "tidy up" a default. `env-vars: DROPBOX` (per finding 1) is precisely equivalent and says what you actually mean.

### 6. The test-count claim in the PR body understates the win

"17 tests exist only behind `cfg(windows)`" is a *net* delta, not the Windows-only count. From the logs, with `0 ignored` on both legs and 222 `#[test]` functions in `apps/desktop/src-tauri/src`:

- macOS ran 197 → **25 tests never run on macOS**
- Windows ran 214 → **8 tests never run on Windows** (the macOS/unix-gated ones, correctly)
- 25 − 8 = the 17 you quoted

So this slice switched on **25** previously-unrun tests, not 17. #121's AC asks for this number to be quoted on the PR and estimated it at ~22; the real figure is 25 and it beats the estimate. Worth correcting in the PR body, since that number is what a future reader will cite. It is also worth recording the other half — 8 macOS-only tests still never run on Windows, which is correct and expected, but it should be a stated fact rather than an unexamined gap.

### 7. `fail-fast: false` is correct but was never actually exercised

In the mutation run, `rust (macos-latest)` completed at 18:47:05 and `rust (windows-latest)` failed at 18:47:23 — macOS had already finished, so nothing was ever at risk of cancellation. The flag is right and I am not asking you to prove it; just noting that by this repo's own standard it remains asserted rather than observed. If a future PR happens to fail on macOS first, that run is the evidence — worth capturing then.

### 8. Merging this invalidates every cache currently on main

`env-vars` feeds the SHA-1 environment hash appended to the key (`config.ts:104-131`), so changing that input changes the key for all three cache blocks. The caches written by run `32883429800` become unreachable and the first post-merge run on main will be fully cold. Expected and harmless — flagging only so a ~10 minute first main run is not misread as a regression introduced by this PR.

### 9. Does this shape survive Slice 3 (#122)? Mostly, with two things to watch

Your question 6. The structure holds — `windows-app` drops in as a sibling job, `permissions: contents: read` is still sufficient, and `fail-fast: false` is on the `rust` matrix only so it does not interact.

Two things I would decide before starting #122:

- **`npm ci` on Windows is safe here — I checked the specific trap.** The classic failure is npm pruning other platforms' optional deps from the lockfile, so `npm ci` on a Windows runner dies on a missing `@rollup/rollup-win32-x64-msvc`. Note that your local `npm ci --dry-run` on macOS could not have caught this. `package-lock.json` does carry the full matrix — `@rollup/rollup-win32-{x64,arm64,ia32}-msvc`, `@esbuild/win32-*`, and `@tauri-apps/cli-win32-x64-msvc` are all present — so `bundle:win` will resolve. Guard it: if anyone regenerates the lockfile with `--omit=optional` or on a trimmed install, Slice 3 breaks on Windows and only on Windows. Also add `actions/setup-node` with `cache: 'npm'` to that job, as the other two have.

- **Cache budget.** After #122 there will be five rust caches sharing a 10 GB repo budget: `rust-checks` x2 platforms, `macos-app`, `windows-app` (a *release*-profile `src-tauri/target`, so a distinct `prefix-key` — do not reuse `rust-checks`), and the out-of-workspace `shell-menu` crate. Windows `target/` for the `windows` 0.61 family is not small. Because only main writes, LRU eviction on main degrades *every* PR silently and looks like nothing at all. #122's AC already says "caching covers this crate's target dir too, or its cost is measured" — I would satisfy that by reading the actual sizes off the first main run after this merges, before sizing #122.

Also worth noting for #122: `shell-menu` has its own `Cargo.lock` and sits outside the workspace, so it needs its own entry in `workspaces:` (newline-separated) rather than being folded in.

---

## Not flagged, deliberately

Clippy without `-D warnings` — the reasoning at `ci.yml:99-103` is correct and the Windows numbers now support it. `permissions: contents: read` — nothing here checks out with write, posts a comment, or uploads an artifact; read is sufficient today and stays sufficient through #122 (your question 5). `frontend` on `macos-latest` — the comment already owns this as a deferred optimisation with a named risk, which is the right call.

Fix findings 1 and 2, correct the comment in 3, and this is good to merge. The slice does what the ticket exists for, and the mutation check is the reason I can say that rather than assume it.

mbsanchez:
## All four findings applied (89d3716), plus a correction to my own claim

### The concurrency finding is the one I most wanted

My worry was cache starvation, and it was **wrong** — the last push in a burst is never cancelled, and `post-if` covers the cancelled case, so the save still runs. But there was a worse bug at the same line, and it is the one that matters: on a push to `main`, `cancel-in-progress` discards **the only automated verification a merged commit ever gets**. Two merges inside six minutes and commit A is never tested on any platform, silently.

With a manual QA strategy this pipeline is the sole automated gate, so that is not a cost worth paying for a cancelled runner. Now scoped: `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`.

### `DROPBOX_REDIRECT_URI` — a real hole, and my comment pointed straight at it

`auth/oauth.rs` reads **two** values through `option_env!`. I keyed one and wrote a comment that was true of both. Now `env-vars: DROPBOX`, which covers both by `startsWith` and, being additive, drops none of the built-ins.

### The `cache-on-failure` comment was wrong

`save-if` gates the save body and returns early; `cache-on-failure` only gates `post-if`. They AND and `save-if` wins — so on a PR branch the flag is inert, which is exactly the case my comment claimed it rescued. Flag kept (it matters on main, and it is what lets the save run on cancellation); comment rewritten to say what it actually does.

### `timeout-minutes` added

20 on the Rust and macOS jobs, 10 on the frontend, against a 360-minute default on runners billed at 2x. The point about hangs being the characteristic failure of Win32 I/O-callback code is well taken — that is a failure mode this leg is now specifically exposed to.

---

## Correction to my own evidence comment

I wrote "17 tests exist only behind `cfg(windows)`". That is the **net delta**, not the number switched on. Verified locally:

```
#[test] in src-tauri/src/   222
macOS runs                  197  → 25 never run on macOS
Windows runs                214  →  8 never run on Windows
                                   25 − 8 = 17 net
```

So this slice switches on **25** previously-unrun tests, not 17 — above the ticket's ~22 estimate rather than below it.

And the other half deserves recording rather than passing quietly: **8 tests run only on macOS and still never run on Windows.** That is correct — they are macOS-gated — but it is a real asymmetry, and stating it is better than leaving it as an unexamined gap.

### Noted for Slice 3 (#122)

Five caches against a 10 GB budget, with eviction on main degrading every PR invisibly. `windows-app` will need its own `prefix-key` (release profile) and `shell-menu` its own `workspaces:` entry. Also recorded there: the Windows leg runs no npm in this slice, so the pruned-optional-deps risk for `npm ci` lands in #122, not here.

Re-running CI now.

mbsanchez:
## QA Result: PASSED ✅

Re-verified on the final commit (`89d3716`), not on the one the review saw — the four fixes landed after it.

`qa_strategy` for this stack is `manual`; not followed here, and for the same reason as Slice 1: no source file changes, so there is nothing to validate in an app window. Every AC is observable in the pipeline itself.

### Final run — all four jobs green

| Job | Result | Time |
|---|---|---|
| `rust (macos-latest)` | ✅ | 3m02s |
| **`rust (windows-latest)`** | ✅ | 7m16s |
| `macos-app` | ✅ | 6m09s |
| `frontend` | ✅ | 15s |

### AC verification

| AC | Evidence |
|---|---|
| `rust` is a matrix over macOS + Windows | Both legs present in every run |
| `fail-fast: false` | Proven, not merely configured: in the mutation run `macos-app` completed at 7m13s **after** Windows had already failed at 4m16s |
| Windows test count exceeds macOS | 214 vs 197 on `89d3716` — **25 tests** switched on that had never run in any pipeline |
| Mutation check | `rust (windows-latest)` RED with `error[E0308]`, macOS green — recorded above |
| Clippy without `-D warnings` on both legs | Green on both; Windows shows 5 lib warnings vs 4 on macOS, so ~4000 never-linted lines cost **one** net new warning |
| Wall-clock recorded | Above, and below |

### What this slice actually bought

Before it, `cargo clippy` had never linted a line of the CfAPI layer and 25 tests had never executed outside one machine. The mutation check settles that the leg detects real breakage rather than merely existing.

The Windows leg is the slowest job at 7m16s, which is the honest price. Total wall-clock is bounded by it and stays under the 8m19s–10m24s the old single job cost — so two platforms and a mutation-proven Windows leg came in **faster than one platform used to be**.

### Recorded rather than left implicit

8 tests run only on macOS and still never run on Windows. Correct — they are macOS-gated — but it is a real asymmetry and belongs in the record.

Ready for PO approval.

Issue #127, opened 2026-08-25
DBSYNC-73: stop the JSON decoder rewriting path keys, and put the decode under test (AFK)


mbsanchez:
The single slice of [DBSYNC-73](https://manuelbsan.atlassian.net/browse/DBSYNC-73) — **AFK**, classified TRIVIAL (no `plan` step).

## The bug

`SyncExtension.swift` decodes `overlay_state.json` with `keyDecodingStrategy = .convertFromSnakeCase`. That strategy applies to **every** key the decoder sees, and historically that included the keys of a `[String: T]` dictionary.

`OverlayState.paths` is exactly such a dictionary, and its keys are relative file paths written by the Rust side. `docs/my_report.pdf` would decode as `docs/myReport.pdf`, never match the lookup in `requestBadgeIdentifier(for:)`, and that file would render **no badge, silently**.

It does not reproduce on macOS 26 — the swift-foundation rewrite no longer converts dictionary keys. But `JSONDecoder` comes from the **user's system Foundation**, and `minimumSystemVersion` is 12.0. So the behaviour varies by end-user OS: correct for the developer, silently broken for a subset of users on a subset of their files.

## The fix, and why it is stronger than "avoid the API"

Drop `keyDecodingStrategy`; declare explicit `CodingKeys` for `updated_at` and `sync_folder`.

Verified by running four decoder configurations against the real payload:

| Struct | Decoder | Result |
|---|---|---|
| today's | `.convertFromSnakeCase` | ok, keys survive |
| today's | plain | **throws** `keyNotFound "updatedAt"` |
| with CodingKeys | plain | ok, keys survive |
| **with CodingKeys** | **`.convertFromSnakeCase`** | **throws** `keyNotFound "updated_at"` |

That last row is the point. Once the keys are explicit, the two configurations are **mutually exclusive on every Foundation and every OS version**. The fix does not merely avoid the OS-dependent API — it makes it *unreachable*. Nobody can reintroduce it and have it fail quietly on someone else's macOS 12; it throws immediately, here and in CI.

Consistency note: `native/windows/shell-menu/src/scope.rs` already reads this same file with literal field names and no rename strategy. This makes both readers describe the contract the same way.

## Acceptance criteria

- [ ] `keyDecodingStrategy` removed; explicit `CodingKeys` map `updatedAt` → `updated_at` and `syncFolder` → `sync_folder`.
[...]

PR #128, opened 2026-08-26
fix(#DBSYNC-73): decode overlay_state with explicit keys, and say so when it fails

mbsanchez:
Fixes [DBSYNC-73](https://manuelbsan.atlassian.net/browse/DBSYNC-73) — GitHub #127. Single slice, classified TRIVIAL.

## The bug

The extension decoded `overlay_state.json` with `keyDecodingStrategy = .convertFromSnakeCase`. That strategy applies to **every** key the decoder sees, and on the Foundation shipped with macOS 12-14 that included the keys of a `[String: T]` dictionary.

`paths` is exactly such a dictionary and its keys are **file paths**. So `docs/my_report.pdf` decoded as `docs/myReport.pdf`, matched nothing in `requestBadgeIdentifier(for:)`, and that file silently rendered no badge.

It does not reproduce on macOS 26 — the swift-foundation rewrite stopped converting dictionary keys — but `JSONDecoder` comes from the **user's** Foundation, and `minimumSystemVersion` is 12.0. Correct for the developer, silently wrong for a subset of users on a subset of their files. That is the worst failure mode available.

## The fix does more than avoid the API — it makes it unreachable

Explicit `CodingKeys` for the two snake_case properties, plain `JSONDecoder`.

The property worth having is the one the original spec missed. With the wire names spelled out, `.convertFromSnakeCase` rewrites `updated_at` to `updatedAt`, then looks for a `CodingKey` whose `stringValue` is `"updated_at"`, and throws `keyNotFound` — **on every Foundation and every OS version**.

So nobody can reintroduce the dangerous configuration and have it fail quietly on someone else's macOS 12. It fails loudly, here and in CI.

## "Cannot be covered by an automated test" was wrong

The ticket recorded this decode as untestable. `OverlayState` and a `decode(from:) throws` move into `BadgeDiff.swift`, which is Foundation-only and already compiled by `Tests/run-badge-diff-tests.sh` — a runner CI runs in the `macos-app` job. **Zero pipeline changes.**

What was actually untestable is the macOS 12 *symptom*; the property that holds everywhere is the mutual exclusivity above. Those are different claims, and conflating them cost this ticket a year of being "unverifiable".

It also belongs there on precedent: three non-diff helpers already live in that file, and both of their doc comments record a defect that survived review "because it sat in a file the test script cannot compile."

## Folded in: the decode failure was silent

`reloadState()` logged carefully when the file could not be **read**, then two lines below guarded the **decode** with a bare `try?` and no log at all.

That asymmetry matters more than it looks. A decode failure sets `state = nil` → drops `syncFolder` → `monitoredDirectoryURLs()` returns `[]` → Finder deregisters the extension's directories and stops calling it. **Every badge disappears, with nothing written anywhere.** Same shape as DBSYNC-76, which this file's own comments record as expensive to find.

Now logged on the transition, with its own flag — a file can be readable and undecodable, and one shared flag would swallow one of the two messages.

## Stack

`desktop-tauri`

## Test Plan

- [x] Six checks in `run-badge-diff-tests.sh`, green
- [x] **Mutation check, both directions** — see below
- [x] `npm run build:finder-sync` → BUILD SUCCEEDED, principal class resolves
- [x] No Rust touched
- [ ] CI green on this PR

```
OverlayState.decode (DBSYNC-73)
  ok   underscored path keys survive decoding
  ok   underscored path keys are not camel-cased
  ok   a path with several underscores survives too
  ok   updated_at maps to updatedAt
  ok   sync_folder maps to syncFolder
  ok   an underscored tier value is untouched
```

| Mutation | Result |
|---|---|
| Reintroduce `.convertFromSnakeCase` alongside the CodingKeys | ❌ FAIL |
| Delete the CodingKeys, keep the plain decoder | ❌ FAIL |
| Restore | ✅ green |

Two ways to get this wrong, two ways to catch it.

**Process note worth recording.** The first run of mutation 2 tested a file that still had mutation 1 applied, because a `cd` failed silently — and that combination is the original working config, so it reported green. Re-run from a verified-clean tree. A mutation check that accidentally tests nothing is precisely the failure these checks exist to prevent, so it would have been dishonest to leave the first result standing.

## What this does not prove

The macOS 12-14 symptom itself. Unreproducible on a modern Foundation by construction. Confirming it would need a macOS 12/13 machine and is not a blocker — the reintroduction guard is the durable part.

## Out of scope

`version` is decoded and never checked — **DBSYNC-91**, needs a behaviour decision.

#DBSYNC-73

### DIFF
--- a/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
+++ b/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
@@ -193,7 +193,7 @@ enum BadgeDiff {
 /// already runs that script.** The original ticket recorded this decode as impossible to
 /// test; it was not, and two defects have already survived review in this project by
 /// sitting where the check script could not reach them.
-struct OverlayState: Decodable, Equatable {
+struct OverlayState: Decodable {
     let version: Int
     let updatedAt: String
     let syncFolder: String?
@@ -230,5 +230,54 @@ struct OverlayState: Decodable, Equatable {
     /// directories and every badge disappears with nothing written anywhere.
     static func decode(from data: Data) throws -> OverlayState {
         try JSONDecoder().decode(OverlayState.self, from: data)
+    }
+}
+
+/// A description of a decoding failure that is safe to put in the system log (DBSYNC-73).
+///
+/// **`String(describing:)` on a `DecodingError` leaks the user's file paths.** The error
+/// carries a `codingPath`, and inside `OverlayState.paths` those components ARE relative
+/// paths from the sync folder:
+///
+///     typeMismatch … Path: paths.`Clients/AcmeCorp_NDA_signed.pdf`
+///
+/// `NSLog` writes to the public unified log, readable via `log show` and captured in any
+/// sysdiagnose. Making the decode failure audible — which is the other half of this ticket
+/// — must not turn it into a privacy leak.
+///
+/// So this reports the *kind* of failure and, at most, a key name from the struct's own
+/// schema. It never emits `codingPath`, and never a value. What is lost is which entry
+/// failed; what is kept is enough to tell a schema mismatch from malformed bytes, which is
+/// the distinction anyone debugging actually needs.
+///
+/// Not reachable from today's writer — `HashMap<String, OverlayTier>` can only emit strings,
+/// and truncation fails earlier as `dataCorrupted`. It is guarded anyway because
+/// `overlay_state.json` is a versioned contract with two readers, and DBSYNC-91 records that
+/// `version` is decoded and never checked: a newer writer reshaping `paths` against an
+/// installed older appex is precisely the gap.
+func logSafeDescription(of error: Error) -> String {
+    guard let decoding = error as? DecodingError else {
+        // Not a decoding failure — a read error, already scoped to a path we logged
+        // ourselves. Still summarised rather than dumped.
+        return "\(type(of: error))"
+    }
+    switch decoding {
+    case .keyNotFound(let key, _):
+        // A key name from OUR schema (`updated_at`, `sync_folder`), never a dictionary key
+        // from `paths`: a missing-key failure is raised against the struct's keyed
+        // container. This is also what both DBSYNC-73 mutations produce, so it is the one
+        // detail worth keeping.
+        return "keyNotFound(\(key.stringValue))"
+    case .typeMismatch(let expected, _):
+        return "typeMismatch(expected \(expected))"
+    case .valueNotFound(let expected, _):
+        return "valueNotFound(expected \(expected))"
+    case .dataCorrupted:
+        // Malformed or truncated JSON. The context here describes the syntax problem, not
+        // the content, but it is dropped anyway — "the bytes are not JSON" is the whole
+        // actionable message.
+        return "dataCorrupted (not valid JSON)"
+    @unknown default:
+        return "DecodingError (unrecognised case)"
     }
 }
--- a/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
+++ b/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
@@ -186,3 +186,49 @@ enum BadgeDiff {
     }
 }
 
+/// The shape of `overlay_state.json`, and the only place it is decoded (DBSYNC-73).
+///
+/// **It lives here rather than in `SyncExtension.swift` for one reason: this file is
+/// Foundation-only, so `Tests/run-badge-diff-tests.sh` can compile it with `swiftc` and CI
+/// already runs that script.** The original ticket recorded this decode as impossible to
+/// test; it was not, and two defects have already survived review in this project by
+/// sitting where the check script could not reach them.
+struct OverlayState: Decodable, Equatable {
+    let version: Int
+    let updatedAt: String
+    let syncFolder: String?
+    /// Relative path (POSIX, no leading slash) → tier id matching registered badge ids.
+    let paths: [String: String]
+
+    /// Explicit, and never `keyDecodingStrategy = .convertFromSnakeCase`.
+    ///
+    /// That strategy applies to **every** key the decoder sees, and on the Foundation
+    /// shipped with macOS 12-14 that included the keys of a `[String: T]` dictionary.
+    /// `paths` is exactly such a dictionary and its keys are file paths, so
+    /// `docs/my_report.pdf` decoded as `docs/myReport.pdf`, matched nothing in
+    /// `requestBadgeIdentifier(for:)`, and that file silently rendered no badge.
+    ///
+    /// It does not reproduce on macOS 26 — the swift-foundation rewrite stopped converting
+    /// dictionary keys — but `JSONDecoder` comes from the USER's Foundation, and
+    /// `minimumSystemVersion` is 12.0. Correct for the developer, silently wrong for a
+    /// subset of users on a subset of their files.
+    ///
+    /// **These keys do more than avoid the strategy: they make it unreachable.** With the
+    /// wire names spelled out, `.convertFromSnakeCase` rewrites `updated_at` to
+    /// `updatedAt`, then looks for a `CodingKey` whose `stringValue` is `"updated_at"`,
+    /// and throws `keyNotFound` — on every Foundation and every OS. Reintroducing it fails
+    /// loudly here and in CI instead of quietly on someone else's machine.
+    enum CodingKeys: String, CodingKey {
+        case version
+        case updatedAt = "updated_at"
+        case syncFolder = "sync_folder"
+        case paths
+    }
+
+    /// Decodes with a plain `JSONDecoder`. Throws rather than returning `nil` so the caller
+    /// can log what went wrong — a silent decode failure deregisters the extension's
+    /// directories and every badge disappears with nothing written anywhere.
+    static func decode(from data: Data) throws -> OverlayState {
+        try JSONDecoder().decode(OverlayState.self, from: data)
+    }
+}
--- a/apps/desktop/native/macos/FinderSyncExtension/SyncExtension.swift
+++ b/apps/desktop/native/macos/FinderSyncExtension/SyncExtension.swift
@@ -136,6 +136,11 @@ final class DropboxSyncFinderSync: FIFinderSync {
     /// expensive to find — an unreadable state file looks exactly like "no files tracked yet".
     private var lastLoadFailed = false
 
+    /// Same transition-logging discipline as [`lastLoadFailed`], for the decode step
+    /// (DBSYNC-73). Separate flag: the file can be readable and undecodable, and conflating
+    /// them would swallow one of the two messages.
+    private var lastDecodeFailed = false
+
     private func reloadState() {
         let url = overlayStateURL()
         guard let data = try? Data(contentsOf: url) else {
@@ -155,12 +160,32 @@ final class DropboxSyncFinderSync: FIFinderSync {
             NSLog("DropboxSync FinderSync: reading %@ again.", url.path)
             lastLoadFailed = false
         }
-        let decoder = JSONDecoder()
-        decoder.keyDecodingStrategy = .convertFromSnakeCase
-        guard let decoded = try? decoder.decode(OverlayState.self, from: data) else {
+        // Decoding lives in `BadgeDiff.OverlayState` so the check script can reach it, and
+        // it throws so this failure can be reported (DBSYNC-73).
+        let decoded: OverlayState
+        do {
+            decoded = try OverlayState.decode(from: data)
+        } catch {
+            // NOT silent. This used to be a bare `try?` with no log, while the read above
+            // it logged carefully — and the consequence is worse than the read failing:
+            // `state = nil` drops `syncFolder`, so `monitoredDirectoryURLs()` returns [],
+            // so Finder deregisters the extension's directories and stops calling it. Every
+            // badge vanishes with nothing written anywhere. That is the shape of DBSYNC-76,
+            // which this file's own comments record as expensive to find.
+            //
+            // Logged on the transition only, like the read path: this runs every 2 seconds.
+            if !lastDecodeFailed {
+                NSLog("DropboxSync FinderSync: cannot decode %@ — no badges will be shown. %@",
+                      url.path, String(describing: error))
+                lastDecodeFailed = true
+            }
             state = nil
             lastUpdatedAt = nil
             return
+        }
+        if lastDecodeFailed {
+            NSLog("DropboxSync FinderSync: decoding %@ again.", url.path)
+            lastDecodeFailed = false
         }
 
         // Nothing was rewritten since the last tick: no diff, no pushes, no work.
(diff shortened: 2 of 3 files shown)


mbsanchez (commented):
## CTO review — DBSYNC-73 / PR #128

**Verdict: CHANGES REQUESTED** (GitHub will not let me set that flag on your own PR, so it is stated here).
One blocking item, three minor. The core of the change is correct and the reasoning behind it is
sound — I re-derived it independently rather than taking the PR body's word for it.

Stack: `desktop-tauri` (explicit declaration in `CLAUDE.md`). Reviewed at `edf5787`, which is the
exact commit all five CI jobs went green on.

---

### What I verified independently

I did not re-run your script and call it evidence. I built a standalone Swift probe outside the
repo that declares both struct variants and runs all four decoder configurations, so a failed `cd`
cannot make it test the wrong thing:

```
shipped + plain     : OK
shipped + snake     : THROWS  <-- mutation 1 caught: keyNotFound 'updated_at'
nokeys  + plain     : THROWS  <-- mutation 2 caught
nokeys  + snake     : OK (the original buggy config); paths keys = ["a_b_c/x_y.md", "docs/my_report.pdf"]
```

Both mutation directions are genuinely red, from a tree I control. The fourth row also confirms the
symptom really is unreproducible on this Foundation — the buggy config preserves the keys here — so
the PR's narrower claim is honest and, as far as I can test it, true.

**Your disclosure of the false-green on mutation 2 is correct and complete.** It is also the most
useful paragraph in the PR. Leave it in.

**CodingKeys vs. the Rust writer — checked, all four correct.** `overlay_state.rs:49-54` declares
`OverlayStateFile` with **no** `rename_all` (the `snake_case` at line 15 is on `OverlayTier`, the
value enum, not the struct), so the field names serialise literally: `version`, `updated_at`,
`sync_folder`, `paths`. All four are present in your `CodingKeys`, nothing dropped, nothing
misspelled. `sync_folder` is `Option<String>` with no `skip_serializing_if`, so it is written as
`null` rather than omitted — which your non-optional-key/optional-value `String?` handles. `u32` →
`Int` is safe on a 64-bit-only target. This matches `scope.rs:195-199` on the Windows side too.

**Xcode target (your question 7) — no action needed, and proven, not assumed.** `project.pbxproj`
has one native target (`DropboxSyncFinderSync`) and a single `PBXSourcesBuildPhase` listing both
`SyncExtension.swift` and `BadgeDiff.swift` (lines 166-167). Both files were already members, so a
struct moving between them needs no project change. CI's `macos-app` job runs `bundle:dev`, which
runs `build:finder-sync` — so the appex genuinely links on this commit.

---

## Critical

*(none — nothing here can produce a wrong badge or a crash)*

## Warnings — must fix before merge

### 1. The new log leaks user file paths into the unified system log

You asked me to check this specifically. It does leak. `String(describing:)` on a `DecodingError`
prints the `codingPath`, and inside `paths` the coding path components **are the user's relative
file paths**. Probe output, verbatim:

```
value-type-mismatch inside paths ->
DecodingError.typeMismatch: expected value of type String. Path: paths.`Clients/AcmeCorp_NDA_signed.pdf`. Debug description: Expected to decode String but found number instead.
```

`NSLog` marks its arguments public in the unified log (that is the whole reason Apple pushes
`os_log` + `%{private}`), so this lands in `log show` and in any sysdiagnose the user ships to
support. A Dropbox client writing customer filenames to a system-wide log is a worse outcome than
the silent failure you replaced.

**Honest scoping, because it changes how you should weigh this.** It is not reachable from today's
writer: `paths` is a `HashMap<String, OverlayTier>`, so it can only ever emit string values, and a
truncated or garbled file fails earlier with `dataCorrupted`, which carries no path:

```
truncated file ->
DecodingError.dataCorrupted: ... The given data was not valid JSON. ... "Unexpected end of file"
missing updated_at ->
DecodingError.keyNotFound: Key 'updated_at' not found ...
```

Both of those are clean. So this is a latent leak, not a live one. I am still calling it blocking,
for one specific reason: this file is an explicitly **versioned cross-reader contract**, and
DBSYNC-91 has just recorded that `version` is decoded and never checked. The exact scenario that
opens the leak — a newer writer changing the shape of `paths`, an older appex still installed,
no version gate to stop it — is the one failure mode this file's own design anticipates and
currently does not defend against. Shipping a public-log path printer into that gap is not a risk
worth taking for zero diagnostic gain.

**Fix, and it lands in the file you just argued for.** Put a redactor next to `OverlayState` in
`BadgeDiff.swift` and cover it in `run-badge-diff-tests.sh`:

```swift
/// A `DecodingError` description safe for the system log: `NSLog` is public, and this
/// error's `codingPath` inside `paths` is a USER FILE PATH (DBSYNC-73).
static func logSafeDescription(of error: Error) -> String {
    guard let error = error as? DecodingError else { return "\(type(of: error))" }
    switch error {
    case let .keyNotFound(key, ctx):   return "keyNotFound '\(key.stringValue)': \(ctx.debugDescription)"
    case let .typeMismatch(_, ctx):    return "typeMismatch at depth \(ctx.codingPath.count): \(ctx.debugDescription)"
    case let .valueNotFound(_, ctx):   return "valueNotFound at depth \(ctx.codingPath.count): \(ctx.debugDescription)"
    case let .dataCorrupted(ctx):      return "dataCorrupted: \(ctx.debugDescription)"
    @unknown default:                  return "unknown DecodingError"
    }
}
```

`keyNotFound` keeps the key name, which is safe and is the one you actually need — it is what both
mutations produce. `debugDescription` never contains a path in any case I probed. The only thing
lost is *which* entry of `paths` was malformed, which you can read out of the file itself.

Then: `url.path, OverlayState.logSafeDescription(of: error)`.

Worth a check that can fail, in the same block as the six you added: decode a payload with a
non-string value under a distinctive path key, and assert the redacted string does **not** contain
that key. That check goes red today, which is the point.

### 2. `BadgeDiff.OverlayState` does not exist

`SyncExtension.swift:163`:

> `// Decoding lives in `BadgeDiff.OverlayState` so the check script can reach it`

`OverlayState` is declared at top level in `BadgeDiff.swift:196`, *outside* the `enum BadgeDiff`
namespace. `BadgeDiff.OverlayState` will not compile. This is a one-word fix, but the entire job of
that comment is to tell the next reader where the type lives, and it sends them somewhere that does
not exist. Suggest: ``// Decoding lives in `OverlayState` (in BadgeDiff.swift, so the check script
can reach it)``.

---

## Answers to your other questions

**2 — `private` → internal: no consequence, and nothing else wants to construct it.** Single module,
single target. `grep -rn OverlayState --include=*.swift` across `apps/desktop/native` returns only
the declaration, `SyncExtension`'s `private var state: OverlayState?` + the call site, and the
checks. No other type of that name exists in the module, so no shadowing. The synthesised memberwise
init is now internal and constructible module-wide; nothing does, and if a future check wants to
build a fixture directly that is a feature. Fine as is.

**3 — the unused `Equatable`: yes, YAGNI, strip it.** Confirmed nothing compares two `OverlayState`
values anywhere. `previousPaths == nil` is `Optional<[String:String]>`, unaffected. Note the
contrast with `BadgeDiff.Changes: Equatable`, which is load-bearing — `check()` at
`BadgeDiffTests.swift:32` uses it. Conformance that a test actually consumes earns its place;
this one does not yet. Non-blocking, and I would not argue if you kept it for the fixture-comparison
checks you may write for warning 1 — but then use it in the same PR.

**4 — `lastDecodeFailed`: transition logic is right, and the two flags cannot wedge each other.**
I traced every ordering. The reset is unconditional and sits **before** the
`decoded.updatedAt == lastUpdatedAt` short-circuit, which is the detail that matters: put it after,
and a recovered decode whose timestamp had not moved would return early and strand the flag at
`true` forever. It is currently safe twice over, since the `catch` also nulls `lastUpdatedAt` — but
the placement means it stays safe if that ever changes. Deliberate or not, it is right.

The one cross-interaction: if the file becomes *unreadable* while `lastDecodeFailed == true`, the
read guard returns early and the flag persists across the outage, so recovery logs both
"reading … again" and "decoding … again". Both statements are true and it is bounded to one line
each. Not a defect.

Separate flags is the correct call, and your comment gives the right reason.

**5 — failure-path end state is byte-identical.** Old: `guard let … else { state = nil;
lastUpdatedAt = nil; return }`. New: `catch { …log…; state = nil; lastUpdatedAt = nil; return }`.
Same two assignments, same order, same early return, `displayedPaths` untouched in both. Downstream
sees no new sequence: on recovery `previousPaths == nil && !displayedPaths.isEmpty` still routes to
`BadgeDiff.resync`, exactly as before. `NSLog` has no side effects on any of it. The only behaviour
delta is the log line itself — which is precisely warning 1.

---

## Suggestions

- **`BadgeDiffTests.swift:414-417` — check six duplicates check three.** "an underscored tier value
  is untouched" asserts `decoded.paths["a_b_c/x_y.md"] == "cloud_only"`; "a path with several
  underscores survives too" at line 402 already asserts the same key against the same value. There
  is no input that reddens one and not the other, so the suite reports six checks where five
  distinct properties exist. Given this project's standard that a check which cannot fail is not
  evidence, a check which cannot fail *independently* should not be counted separately. Either drop
  it, or make it earn its name by asserting on `plain.txt` (a key with no underscore, value
  `syncing`) — that at least covers a different row.
- **`BadgeDiffTests.swift:393-394` — `failures += 1` immediately before `exit(1)` is dead.** Harmless,
  but it reads as if the run would continue and tally. Drop the increment or drop the `exit`; note
  the other helpers only increment and let the run finish, which is the nicer behaviour and would be
  consistent here.
- **`BadgeDiff.swift`'s file header** still describes the file as purely the badge-diff logic. It now
  also owns the state contract's decode. One clause in the header would save the next reader from
  concluding the struct is misfiled — the struct's own doc comment explains itself well, but only
  once you have found it.
- **`version` is `Int` and non-optional**, so a writer that ever drops the field breaks the decode
  loudly. That is the right default and DBSYNC-91 is the right place for the rest.

---

## On the process claims in the PR body

- "Cannot be covered by an automated test was wrong" — agreed, and the distinction you draw between
  the *symptom* and the *property* is the correct one. Worth reusing.
- "What this does not prove" is accurately scoped. I confirmed the macOS 12-14 behaviour is not
  observable here rather than assuming it.
- The precedent argument for moving the decode into a compilable file is the strongest part of this
  PR and generalises past this ticket.

Fix warning 1 and warning 2, decide on the `Equatable`, and this is a merge. Re-request when the
redaction check is green.

*(Reviewed by Claude — architect-reviewer)*

mbsanchez:
## All four findings applied (5713356)

### The leak was real, and it is the finding that mattered

I asked you to check the log could not leak user data. It could, and you showed it rather than reasoned it:

```
typeMismatch … Path: paths.`Clients/AcmeCorp_NDA_signed.pdf`
```

`NSLog` writes to the public unified log. This PR would have traded a silent failure for a privacy leak, **under the banner of better diagnostics** — which is the kind of trade that gets waved through precisely because the headline sounds like an improvement.

`logSafeDescription(of:)` now reports the failure *kind* and, at most, a key name from our own schema. Never `codingPath`, never a value. Five checks added, and one of them is deliberately unusual:

```
ok   a decode error never names the user's file
ok   a decode error never leaks the coding path
ok   but it still says what kind of failure it was
ok   the raw description WOULD have leaked it (guard is load-bearing)
ok   a missing schema key is named, because it is ours
```

The fourth asserts that `String(describing:)` **still leaks today**. That makes the guard demonstrably load-bearing rather than decorative — and if Foundation ever stops emitting the path, that check goes red and tells us the guard can go.

**Mutation check on the redactor:** replacing it with the raw `String(describing:)` turns exactly the two privacy checks red, and nothing else. Restored from a verified-clean tree, suite green.

Your scoping is right and I kept it: not reachable from today's writer, guarded anyway because this is a versioned contract with two readers and DBSYNC-91 records that `version` is decoded and never checked.

### `BadgeDiff.OverlayState` does not exist — fixed

A comment pointing readers at a type that would not compile. Worse than no comment: it sends someone looking for a nesting that is not there.

### `Equatable` stripped, and check six now earns its name

You were right that nothing compared two values.

And the duplicate is a better catch than it looks: six checks over five distinct properties means the *count* was overstating the coverage. It now targets `plain.txt`, isolating the value position from the key position — which is what the name claimed all along.

### On the false-green disclosure

Thank you for validating it independently rather than taking my word. Building the probe outside the repo so no `cd` could misdirect it is the right instinct, and it is what let row 4 of your table confirm the symptom is genuinely unreproducible here — which is the claim I most wanted a second pair of eyes on.

### Not changed

`CodingKeys` completeness, the `private` → internal move, the `lastDecodeFailed` interaction, and the failure-path equivalence: all confirmed correct by your reading of `overlay_state.rs` and the surrounding code. Nothing to do.

Your note on the flag's placement — that the unconditional reset sits *before* the `updatedAt` short-circuit, and would strand the flag if it moved after — is the kind of thing I would not have written down. It is in the code now by accident of ordering; worth knowing it is load-bearing.

Re-running CI.

mbsanchez:
## QA Result: PASSED ✅

Verified on the final commit (`5713356`), not the one the review saw.

`qa_strategy` is `manual`. Followed only partly here, and the gap is stated rather than papered over — see the bottom.

### CI, all five jobs green

| Job | Result | Time |
|---|---|---|
| `rust (macos-latest)` | ✅ | 1m01s |
| `rust (windows-latest)` | ✅ | 2m40s |
| `macos-app` | ✅ | 2m42s |
| `windows-app` | ✅ | 3m13s |
| `frontend` | ✅ | 21s |

`macos-app` is the job that runs `run-badge-diff-tests.sh`, so **all eleven checks from this ticket now run on every PR** — six on the decode, five on the redaction.

### AC verification

| AC | Evidence |
|---|---|
| `keyDecodingStrategy` removed, explicit `CodingKeys` | `grep` finds neither the strategy nor the local struct in `SyncExtension.swift` |
| `OverlayState` + `decode(from:)` in `BadgeDiff.swift` | Compiled by the check script, which CI runs |
| Checks for underscored keys and both mappings | Six, green |
| **Mutation, both directions** | Reintroduce the strategy → red. Delete the CodingKeys → red. Both from a verified-clean tree |
| Silent decode failure now logs | On the transition, with its own flag |
| `build:finder-sync` SUCCEEDED | Locally and in CI's `macos-app` |
| `cargo test` / `cargo clippy` unaffected | No Rust file in the diff |

Plus the one that was not in the original ACs and turned out to matter most: a decode error no longer leaks the user's file paths into the public system log. Mutating the redactor back to `String(describing:)` reddens exactly the two privacy checks.

### What nobody has looked at

**No one has seen a badge render from this build.** The automated checks cover the decode thoroughly, and CI proves the appex links — but the live path (`reloadState` → `state` → `monitoredDirectoryURLs`) has not been exercised on a running Finder since the struct moved files.

The risk is low and bounded: if the decode were broken, all eleven checks would be red, and if the move broke linking, `build:finder-sync` would have failed. But "low and bounded" is not "observed", and this project's record on that distinction is not good.

A two-minute smoke test would close it — install the build, open the sync folder, confirm badges still appear. Recorded here as an accepted residual rather than a completed check.

**And the original symptom remains unreproducible by construction**: dictionary-key rewriting does not happen on a modern Foundation. Confirming the macOS 12-14 behaviour needs that OS. Not a blocker; the durable win is that the dangerous config can no longer be reintroduced silently.

Ready for PO approval.

PR #129, opened 2026-08-26
style(#DBSYNC-90): format the Rust tree and enforce cargo fmt in CI

mbsanchez:
Closes [DBSYNC-90](https://manuelbsan.atlassian.net/browse/DBSYNC-90). Two commits: the reformat alone, then the CI check.

## Two figures in the ticket were wrong, and both were mine

**It is 22 files, not 161.** My original count came from `cargo fmt --check | grep -c "^Diff in"`, and rustfmt emits that line **per hunk**, not per file. 161 hunks across 22 files. The figure propagated into DBSYNC-90, DBSYNC-38 and several comments.

**Blame damage was overstated, badly.** The ticket called it "the main cost of doing it at all". Measured:

```
lines attributed to the reformat commit, whole tree:   14
same, with the commit ignored via .git-blame-ignore-revs: 14
same, with an explicit --ignore-rev:                      14
```

Of 720 touched lines, only 14 end up attributed to the reformat — most changes are line *splits* whose first fragment keeps its original author. And ignoring the commit does not move those 14, by either mechanism: they are genuinely new text with no predecessor line, so `blame` has nowhere to send them.

So `.git-blame-ignore-revs` **changes nothing measurable for this commit**. It is kept as policy for the next mass change, with that written into the file itself rather than left as an implication.

Had I measured first, this ticket would have been argued as "cosmetic, cheap" rather than "cosmetic, with a serious blame cost". The conclusion happens to be the same; the reasoning I gave you was not.

## What is in here

**`36b3af1` — the reformat, and nothing else.** rustfmt defaults, no `rustfmt.toml`, no manual edits. 22 files, +720/-260.

**`2e6fa54` — the check and the blame file.** `cargo fmt --check` on both matrix legs. rustfmt is deterministic so the Windows leg is redundant in principle, but a divergence there would be a real finding about the toolchain and it costs a second.

Turning this on was deliberately kept out of DBSYNC-38: doing it before the tree was formatted would have made every job red on layout rather than defects — which is exactly how this workflow earned being switched off from 2026-07-12 to 2026-08-22.

## Test Plan

- [x] `cargo test` — **197 passed before, 197 after**, identical. A reformat that changes behaviour is not a reformat
- [x] `cargo fmt --check` passes locally
- [x] Only `.rs` files in the reformat commit — verified, zero non-`.rs` paths
- [ ] CI green, including the new step on both platforms

Also worth recording, since it is the first time it has been true: **`cargo test` no longer touches the real `overlay_state.json`.** Hash identical before and after the run — DBSYNC-75's fix working in the wild rather than in a unit test.

## What this does not do

Nothing a user will notice. It is cosmetic by construction, and the honest case for it is only that future diffs stop carrying layout noise. If that is not worth a 720-line commit to you, closing this unmerged is a reasonable call and the reformat is trivially revertible.

#DBSYNC-90

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,20 @@
+# Revisions to skip when running `git blame`.
+#
+# Enable locally with:
+#     git config blame.ignoreRevsFile .git-blame-ignore-revs
+#
+# GitHub honours this file automatically. Only ever add commits that are PURELY
+# mechanical — a reformat, a mass rename — never one that changes behaviour, or blame
+# will hide the change that actually broke something.
+#
+# MEASURED, and smaller than expected: of the 720 lines the reformat below touched, only
+# 14 end up attributed to it, because most were line splits whose first fragment keeps its
+# original author. Ignoring the commit does not move those 14 either — they are genuinely
+# new text with no predecessor line to fall back to, so `blame` has nowhere to send them.
+#
+# So this file changes nothing measurable for the commit it lists. It is kept as policy
+# for the next mass change rather than as a repair of this one; DBSYNC-90 originally
+# called blame damage "the main cost of doing it at all", and that was wrong.
+
+# style(#DBSYNC-90): apply cargo fmt — 22 files, rustfmt defaults, zero logic change
+36b3af1b380ec369d1e8dab8325a27d91682cd11
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,6 +110,17 @@ jobs:
       - name: Rust tests
         run: cargo test --manifest-path apps/desktop/src-tauri/Cargo.toml
 
+      - name: Rust formatting
+        # Added by DBSYNC-90, after `cargo fmt` was applied to the tree. Deliberately NOT
+        # part of DBSYNC-38: turning this on before formatting would have made every job
+        # red on layout rather than on defects, which is how this workflow earned being
+        # switched off for six weeks.
+        #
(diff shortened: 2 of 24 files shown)


mbsanchez:
## CI green, and a note on a check that did not work

All five jobs pass, including the new `cargo fmt --check` on both matrix legs: `rust (macos)` 1m15s, `rust (windows)` 2m13s, `macos-app` 2m24s, `windows-app` 3m02s, `frontend` 22s.

### An attempt at proving semantic equivalence, and why it failed

I wanted a check stronger than "the tests still pass" — something showing the reformat moved no tokens. Two attempts, both **inconclusive by construction**:

1. **Strip all whitespace, compare hashes.** 11 of 22 files still differed. Not a red flag: rustfmt reorders `use` statements and adds trailing commas, both of which change the character stream. The character-level diff showed *equal counts added and removed* — the signature of reordering, not of content changing.
2. **Compare the sorted set of lines.** 20 of 22 differed, which is worse and equally meaningless: rustfmt *splits* lines, so one line becoming two changes the set by definition.

Neither could separate "layout changed" from "code changed", so neither is evidence. Recording that rather than presenting the first one as a passed check.

What actually supports this commit is narrower and sufficient for a formatter:

- produced by `cargo fmt` with **zero manual edits**, rustfmt defaults, no `rustfmt.toml`
- **197 tests before, 197 after**, identical
- compiles on macOS *and* Windows — and the Windows leg only exists since yesterday, so this is the first repo-wide change ever verified on both

If a reviewer wants a stronger guarantee than that, the honest answer is that it would need a token-level comparison, which is not worth building for a formatting commit.

Ready to merge.

Issue #132, opened 2026-08-27
DBSYNC-91 Slice 2: Name the version when the decode itself fails (AFK)


mbsanchez:
Vertical slice 2 of **DBSYNC-91** — `overlay_state.json` declares a schema version that no reader ever checks.

Jira: https://manuelbsan.atlassian.net/browse/DBSYNC-91
Blocked by: #131

## Why this slice exists

Architectural review found it, and it is the interesting half of the ticket.

A version guard that runs **after** a successful decode cannot fire on the most likely v2. If a breaking change reshapes `paths` — the field most likely to change — the v1 decoder throws `typeMismatch` before any guard is reached, and the user gets a generic "cannot decode" instead of "version 2 is not supported".

That is the same half-mechanism this ticket exists to fix, moved one step later. Slice 1 alone would be nearly decorative for precisely the scenario it was built for.

## What

On a decode failure, `OverlayState.decode` retries a one-field header decode. If that yields a version other than `1`, report `unsupportedVersion(N)`. Otherwise rethrow the original error, unchanged.

**Why not header-first, which would be simpler to read.** `JSONDecoder` parses the whole document to extract one field, so decoding a header before the struct means two full parses of a file that can hold thousands of paths — on every tick of a 2-second timer, in the happy path. Putting the retry in the `catch` keeps the happy path at one parse and pays the second only when something has already gone wrong.

## Acceptance criteria

- [ ] On a decode failure, a one-field header decode is retried. A version other than `1` is reported as `unsupportedVersion(N)`; anything else rethrows the original error unchanged.
- [ ] The happy path still parses the document **once**. Stated as a claim the reviewer can check by reading the control flow — not asserted from a benchmark nobody ran.
- [ ] A check covers a `version: 2` payload whose `paths` is reshaped so that v1 cannot decode it, and asserts the reported failure names the version rather than `typeMismatch`. **A mutation turns it red.**
- [ ] **Negative control:** a malformed-JSON payload with no readable version still reports `dataCorrupted`, not a version error. The fallback must not relabel unrelated failures — without this check it could report `unsupportedVersion` for everything and still look green. The control must itself be shown able to fail.
- [ ] `./Tests/run-badge-diff-tests.sh` passes; check count stated.

## Out of scope

- Bumping the version.
- Anything in slice 1 (#131).
- Any Windows change.

## Not proven by any of this

[...]

Issue #131, opened 2026-08-27
DBSYNC-91 Slice 1: Refuse an unknown overlay_state version (AFK)


mbsanchez:
Vertical slice 1 of **DBSYNC-91** — `overlay_state.json` declares a schema version that no reader ever checks.

Jira: https://manuelbsan.atlassian.net/browse/DBSYNC-91 (decision and full reasoning live there)

## What

The macOS reader decodes `version` into `OverlayState` and never looks at it. A file written with `version: 2` and an incompatible schema is applied exactly as if it were v1 — whatever fields still line up are used, the rest silently ignored.

**Decision, taken by the maintainer before any code: refuse.** `version == 1` is accepted and nothing else. An unknown version means no state, no badges, and exactly one log line saying so. Best-effort was declined because it is what the code already does, and because DBSYNC-84 settled that a badge that is confidently wrong is worse than no badge. `version <= KNOWN` was declined because it promises that v2 is a superset of v1, which no writer has ever promised — `version` is documented as bumping *on a breaking change*.

## Where

The guard goes in `OverlayState.decode` in `BadgeDiff.swift`, and throws.

- That file is Foundation-only, so `Tests/run-badge-diff-tests.sh` compiles it with `swiftc`. `SyncExtension.swift` is reachable from no check at all. Moving the decode there is exactly what DBSYNC-73 was for.
- `reloadState()` already catches decode failures, clears state, and logs once on the transition via `lastDecodeFailed`. Refusal reuses that path, so "exactly one line, never silence" needs no new machinery.

## Acceptance criteria

- [ ] `OverlayState.decode` throws on any `version` other than `1`. The accepted version is a single named constant, not a literal buried in a condition.
- [ ] `logSafeDescription` reports it as `unsupportedVersion(N)` — it names the version found, and never a path from `paths`. The existing privacy check must still pass.
- [ ] A refused file produces **exactly one** log line, on the transition, through the existing `lastDecodeFailed` flag — not one every 2 seconds.
- [ ] `state` and `lastUpdatedAt` are both cleared, as on the other two failure paths. A stale `lastUpdatedAt` would skip the diff when a good file comes back with the same stamp, and no badge would ever be pushed again.
- [ ] A check in `run-badge-diff-tests.sh` covers a `version: 2` payload that is otherwise perfectly valid, and **a mutation turns it red**. The mutation and its output go on the PR.
[...]

PR #133, opened 2026-08-28
fix(#DBSYNC-91): act on overlay_state.json's version instead of decoding and discarding it

mbsanchez:
## Summary

`overlay_state.json` documents itself as a versioned contract — "bump on a breaking change" — and the macOS reader decoded `version` into `OverlayState` and never looked at it. A file with a bumped version and an incompatible schema was applied exactly as if it were v1: whatever fields still lined up were used, the rest silently ignored. The field documented a contract and enforced nothing.

Two commits, closing it from both directions.

**617d797 — refuse an unknown version.** `OverlayState.supportedVersion`, an `OverlayStateError.unsupportedVersion(Int)`, a guard in `decode`, a branch in `logSafeDescription`, and the bump rule in `overlay_state.rs` now naming both readers.

**9ef84e8 — name the version when the file will not parse against v1.** The guard above runs *after* a successful decode, so it cannot fire on the breaking change most likely to happen: reshape `paths` and v1 throws `typeMismatch` before it has a version to look at. The failure path retries a one-field header decode.

Slices #131 and #132. Both commits are independently green — the first passes 62 checks on its own.

## Stack

`desktop-tauri` — macOS Finder Sync appex (Swift) plus one Rust doc comment. No behaviour change on the Rust side.

## Jira Ticket

DBSYNC-91 — https://manuelbsan.atlassian.net/browse/DBSYNC-91

The decision was recorded on the ticket, with its reasoning, before any code was written. That was its first acceptance criterion.

## Why refuse, rather than best-effort

Best-effort is what the code already did, minus a warning. It also inverts this project's own rule, settled in DBSYNC-84: **a badge that is confidently wrong is worse than no badge**. A user on a mismatched pair now sees badges missing — which they report — instead of a sync status nobody checked.

`version <= KNOWN` was the tempting middle and was declined: it promises that v2 is a superset of v1, which no writer has ever promised. `version` is documented as bumping *on a breaking change*, so a bump is precisely the case a reader must not guess at.

The mismatched pair is measured, not hypothetical. DBSYNC-86 found that every reinstall registers a **new** plug-in instance and the old ones linger, so an old appex can genuinely be the one Finder has loaded while a newer app writes the file.

## Two things in the diff that look like style and are not

**`logSafeDescription`'s new branch sits ahead of the `DecodingError` cast.** Behind it, the function's `guard ... else { return "\(type(of: error))" }` renders a refusal as the bare string `OverlayStateError` and loses the version number — the one detail the error exists to carry. The ordering is the feature.

**The header retry sits inside the `catch`, not before the decode.** Header-first reads better and costs more: `JSONDecoder` parses the whole document to extract one field, so it means two full parses of a file that can hold thousands of path entries, on every tick of a 2-second timer, in the case where nothing is wrong. This way the healthy path parses once and the second parse is paid only when something has already failed. That is a control-flow claim readable from the diff — **no benchmark was run and none is implied.**

## `SyncExtension.swift` is untouched, deliberately

Two acceptance criteria are about its behaviour — one log line on the transition, `state` and `lastUpdatedAt` both cleared — and `reloadState()` already does all of it for anything `decode` throws. An edit there would have meant the design went wrong.

## Test Plan

- [x] `./apps/desktop/native/macos/FinderSyncExtension/Tests/run-badge-diff-tests.sh` — **58 → 62 → 66**. The 58 baseline was measured on `main` at fa8423d before any edit.
- [x] `cargo test --manifest-path apps/desktop/src-tauri/Cargo.toml` — **197**, unchanged. The Rust edit is a comment; a change here would mean something else moved.
- [x] `cargo fmt --check` — clean.
- [x] `cargo clippy` — 4 lib warnings, unchanged from before this branch.

### Mutations

**The six-mutation table that stood here was stale and partly wrong**, and review caught both problems: it mixed outputs from two different commits, and it named checks that later commits deleted. Rather than patch it in place, it is replaced by a single table run entirely against the current head — see the comment below, *Re-review findings addressed*.

Summary: **eight mutations**, each applied to a tree verified identical to baseline first and reverted before the next. Two of them (G, H) cover a function that did not exist when this PR opened. One of them (D) reddens three checks written for DBSYNC-73, before this branch existed, which makes it the strongest evidence here.

### Not proven by any of this

**Nobody has watched a real Finder refuse a real v2 file, because nothing writes v2.** Every check here is about the decoder, not about the extension end-to-end under a genuine mismatched install. Stated rather than left implied.

A v2 whose bytes are *also* malformed reports `dataCorrupted`, not a version error. Correct — there is nothing to read a version out of — but it means the fallback covers "unparseable against v1", not "unparseable at all".

## Windows: gap accepted, not fixed

`native/windows/shell-menu/src/scope.rs` is the only Windows reader that exists (`shell-overlay/` is a README). It reads **only** `sync_folder`, through `#[serde(default)] Option<String>`, and draws no badges — so the confidently-wrong-badge failure this PR is about cannot occur there. A v2 that moved that field already degrades to `None`, which is the same outcome a guard would force.

The condition attached to accepting it is the last hunk of this PR: the bump rule in `overlay_state.rs` now says a bump means changing both readers, and says what each one does when left behind. Tracked for runtime validation on DBSYNC-89.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
+++ b/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
@@ -225,12 +225,47 @@ struct OverlayState: Decodable {
         case paths
     }
 
+    /// The only contract version this reader understands (DBSYNC-91).
+    ///
+    /// Named rather than written inline so that it is greppable from the writer's end:
+    /// `overlay_state.rs` documents the bump rule, and this is the constant that rule is
+    /// actually talking about.
+    static let supportedVersion = 1
+
     /// Decodes with a plain `JSONDecoder`. Throws rather than returning `nil` so the caller
     /// can log what went wrong — a silent decode failure deregisters the extension's
     /// directories and every badge disappears with nothing written anywhere.
+    ///
+    /// **Refuses any version other than [`supportedVersion`] (DBSYNC-91.)** Before this,
+    /// `version` was decoded and never read: a file written with a bumped version and an
+    /// incompatible schema was applied exactly as if it were v1, using whatever fields
+    /// still lined up and ignoring the rest. That made the field documentation rather than
+    /// a mechanism.
+    ///
+    /// Refusing rather than best-effort follows this project's own rule, settled in
+    /// DBSYNC-84: a badge that is confidently wrong is worse than no badge. A user on a
+    /// mismatched pair sees badges missing — which they report — instead of badges that
+    /// state a sync status nobody checked.
+    ///
+    /// It is not a hypothetical pairing. DBSYNC-86 measured that every reinstall registers
+    /// a NEW plug-in instance and the old ones linger, so an old appex can genuinely be the
+    /// one Finder has loaded while a newer app writes the file.
     static func decode(from data: Data) throws -> OverlayState {
-        try JSONDecoder().decode(OverlayState.self, from: data)
+        let decoded = try JSONDecoder().decode(OverlayState.self, from: data)
+        guard decoded.version == supportedVersion else {
+            throw OverlayStateError.unsupportedVersion(decoded.version)
+        }
+        return decoded
     }
+}
+
+/// A refusal to use `overlay_state.json`, distinct from a failure to parse it (DBSYNC-91).
+///
+/// Carries the version found because the log line is required to name it: "no badges" with
+/// no reason reads to a user like the sandbox or permissions problems this extension has
+/// already had (DBSYNC-76), and those are debugged in a completely different direction.
+enum OverlayStateError: Error {
+    case unsupportedVersion(Int)
 }
 
 /// A description of a decoding failure that is safe to put in the system log (DBSYNC-73).
@@ -252,10 +287,17 @@ struct OverlayState: Decodable {
 ///
 /// Not reachable from today's writer — `HashMap<String, OverlayTier>` can only emit strings,
 /// and truncation fails earlier as `dataCorrupted`. It is guarded anyway because
-/// `overlay_state.json` is a versioned contract with two readers, and DBSYNC-91 records that
-/// `version` is decoded and never checked: a newer writer reshaping `paths` against an
-/// installed older appex is precisely the gap.
+/// `overlay_state.json` is a versioned contract with two readers, and a newer writer
+/// reshaping `paths` against an installed older appex is precisely the gap that
+/// [`OverlayState.supportedVersion`] now closes.
 func logSafeDescription(of error: Error) -> String {
+    // Handled BEFORE the `DecodingError` cast below, which would otherwise fall through to
+    // the type-name branch and report a bare "OverlayStateError" — losing the one detail
+    // this error exists to carry (DBSYNC-91). A schema version is an integer from our own
+    // contract, so unlike a `codingPath` there is nothing here to redact.
+    if case .unsupportedVersion(let found)? = error as? OverlayStateError {
+        return "unsupportedVersion(\(found))"
+    }
     guard let decoding = error as? DecodingError else {
         // Not a decoding failure — a read error, already scoped to a path we logged
         // ourselves. Still summarised rather than dumped.
--- a/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
+++ b/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
@@ -251,12 +251,52 @@ struct OverlayState: Decodable {
     /// a NEW plug-in instance and the old ones linger, so an old appex can genuinely be the
     /// one Finder has loaded while a newer app writes the file.
     static func decode(from data: Data) throws -> OverlayState {
-        let decoded = try JSONDecoder().decode(OverlayState.self, from: data)
+        let decoded: OverlayState
+        do {
+            decoded = try JSONDecoder().decode(OverlayState.self, from: data)
+        } catch {
+            // The version check below is unreachable when the decode itself fails, and the
+            // most plausible breaking change is exactly that case: reshape `paths` and a v1
+            // decoder throws `typeMismatch` long before it has a version to look at. A
+            // guard that only fires when the rest of the document still parses is the same
+            // half-mechanism this ticket exists to remove, moved one step later.
+            //
+            // So: if the bytes are still readable enough to yield a version, and that
+            // version is not ours, say so. Otherwise rethrow untouched.
+            //
+            // `version != supportedVersion` is not a formality. Without it, every ordinary
+            // v1 failure — the type-mismatched tier from DBSYNC-73 included — would be
+            // relabelled a version problem, which is a worse lie than the silence this
+            // ticket started from.
+            if let header = try? JSONDecoder().decode(VersionHeader.self, from: data),
+               header.version != supportedVersion {
+                throw OverlayStateError.unsupportedVersion(header.version)
+            }
+            throw error
+        }
         guard decoded.version == supportedVersion else {
             throw OverlayStateError.unsupportedVersion(decoded.version)
         }
         return decoded
     }
+}
+
+/// Just enough of `overlay_state.json` to read its version when the rest will not parse
+/// (DBSYNC-91).
+///
+/// **Only used from the failure path.** Decoding this first, then the full struct, would
+/// read better — but `JSONDecoder` parses the whole document to extract one field, so
+/// header-first means two full parses of a file that can hold thousands of path entries, on
+/// every tick of `reloadState()`'s 2-second timer, in the case where nothing is wrong.
+/// Retrying inside the `catch` keeps the healthy path at one parse and pays the second only
+/// when something has already failed.
+///
+/// Private so that nothing outside the decoder starts using it as a cheap way to peek at the
+/// file: it is not cheap, and it answers only one question.
+private struct VersionHeader: Decodable {
+    /// No `CodingKeys` needed — `version` is the one key in this contract that was never
+    /// snake-cased. See `OverlayState.CodingKeys` for why that matters here.
+    let version: Int
 }
 
 /// A refusal to use `overlay_state.json`, distinct from a failure to parse it (DBSYNC-91).
--- a/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
+++ b/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
@@ -321,6 +321,32 @@ enum OverlayStateError: Error {
     case unsupportedVersion(Int)
 }
 
+/// How the log line should describe what happened: `"refusing"` or `"cannot decode"`.
+///
+/// A refused file parsed perfectly — nothing failed, the reader declined it — and calling
+/// that "cannot decode" points whoever reads the log at corruption, or at the sandbox
+/// problems of DBSYNC-76, which are debugged in a completely different direction. Telling
+/// the two apart is the whole reason [`OverlayStateError`] is its own type.
+///
+/// **It lives here, and not inline at the one call site, for the same reason the decode
+/// does: `SyncExtension.swift` cannot be compiled by `Tests/run-badge-diff-tests.sh`.** A
+/// first draft put this branch there, where no check could reach it — in a change whose own
+/// argument is that unreachable code is where this project's defects survive review.
+///
+/// `switch` rather than `error is OverlayStateError`, for the same reason
+/// [`logSafeDescription`] uses one: a bare `is` check would absorb every future case into
+/// `"refusing"` silently. A future case describing a genuine parse failure would then be
+/// logged as a refusal — the precise confusion this function exists to prevent. A first
+/// draft of this function did exactly that, in the same commit that removed the pattern
+/// from `logSafeDescription`.
+func logVerb(for error: Error) -> String {
+    guard let refusal = error as? OverlayStateError else { return "cannot decode" }
+    switch refusal {
+    case .unsupportedVersion:
+        return "refusing"
+    }
+}
+
 /// A description of a decoding failure that is safe to put in the system log (DBSYNC-73).
 ///
 /// **`String(describing:)` on a `DecodingError` leaks the user's file paths.** The error
@@ -343,23 +369,10 @@ enum OverlayStateError: Error {
 /// version check does NOT make that redundant: the case that still reaches the
 /// `typeMismatch` branch is a **same-version** reshape — a writer that changes the shape of
 /// `paths` and forgets to bump `version`. That is a human mistake rather than a contract
-/// event, so no version guard can ever catch it, which is exactly why the redaction has to
-/// survive independently of one.
-/// How the log line should describe what happened: `"refusing"` or `"cannot decode"`.
-///
-/// A refused file parsed perfectly — nothing failed, the reader declined it — and calling
-/// that "cannot decode" points whoever reads the log at corruption, or at the sandbox
-/// problems of DBSYNC-76, which are debugged in a completely different direction. Telling
-/// the two apart is the whole reason [`OverlayStateError`] is its own type.
-///
-/// **It lives here, and not inline at the one call site, for the same reason the decode
-/// does: `SyncExtension.swift` cannot be compiled by `Tests/run-badge-diff-tests.sh`.** A
-/// first draft put this branch there, where no check could reach it — in a change whose own
-/// argument is that unreachable code is where this project's defects survive review.
-func logVerb(for error: Error) -> String {
-    error is OverlayStateError ? "refusing" : "cannot decode"
-}
-
+/// event, so no version guard can ever catch it. The retype route documented on
+/// [`VersionHeader.version`] lands here too, though its `codingPath` is `version` and holds
+/// nothing of the user's. Either way the redaction has to survive independently of the
+/// version check.
 func logSafeDescription(of error: Error) -> String {
     // Handled BEFORE the `DecodingError` cast below, which would otherwise fall through to
     // the type-name branch and report a bare "OverlayStateError" — losing the one detail
--- a/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
+++ b/apps/desktop/native/macos/FinderSyncExtension/BadgeDiff.swift
@@ -288,14 +288,27 @@ struct OverlayState: Decodable {
 /// read better — but `JSONDecoder` parses the whole document to extract one field, so
 /// header-first means two full parses of a file that can hold thousands of path entries, on
 /// every tick of `reloadState()`'s 2-second timer, in the case where nothing is wrong.
-/// Retrying inside the `catch` keeps the healthy path at one parse and pays the second only
-/// when something has already failed.
+/// Retrying inside the `catch` keeps the healthy path at one parse.
+///
+/// The second parse is **not** rare in the case this exists for, and it would be dishonest
+/// to imply otherwise: a mismatched pair (DBSYNC-86) persists until the user reinstalls, and
+/// the `updatedAt` short-circuit in `reloadState()` only guards the success path — so a
+/// refused file is parsed twice every 2 seconds for as long as the mismatch lasts. The trade
+/// is still the right way round, because it moves the cost onto the broken case instead of
+/// the healthy one, but it is a trade rather than a free lunch.
 ///
 /// Private so that nothing outside the decoder starts using it as a cheap way to peek at the
 /// file: it is not cheap, and it answers only one question.
 private struct VersionHeader: Decodable {
     /// No `CodingKeys` needed — `version` is the one key in this contract that was never
     /// snake-cased. See `OverlayState.CodingKeys` for why that matters here.
+    ///
+    /// **`Int` bounds what the retry can rescue, and the bound is worth knowing.** A v2 that
+    /// keeps the key but changes its TYPE — a semver string, say — fails this decode too,
+    /// and the log falls back to `typeMismatch(expected Int)` with no indication that
+    /// `version` is the field in question. A v2 that *renames* the key degrades better:
+    /// `keyNotFound(version)`, because that branch keeps the key name. So this covers
+    /// "unparseable against the v1 schema", not "unparseable in any way".
     let version: Int
 }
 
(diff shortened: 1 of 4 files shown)


mbsanchez:
## Code review: changes requested

Reviewed independently, then the two load-bearing findings were reproduced by hand before acting on them. **This PR's central claim is that its checks are mutation-proven evidence, and two of the five new checks are not.** That is the wrong thing for this particular PR to get wrong.

---

### 1. Critical — `Tests/BadgeDiffTests.swift:501-504`: a check that cannot fail, with a comment asserting it can

```swift
// The other direction, and not a formality: a guard written the wrong way round
// refuses the only version that exists, and every check above would still pass.
checkBool("the supported version is still accepted",
          (try? OverlayState.decode(from: json)) != nil, true)
```

`json` was already decoded at line 391, in a guard that calls `exit(1)` on failure. Reaching line 504 *proves* `decode(from: json)` succeeded. The check is green by construction.

Reproduced by inverting the guard exactly as the comment describes:

```
OverlayState.decode (DBSYNC-73)
  FAIL OverlayState.decode threw on a valid payload

$ grep -c "the supported version is still accepted"
0   # never printed — the run aborts before the section
```

So the comment is wrong twice: the checks above do not pass, and the check itself does not run.

Worse, **the PR body already knew this**. It documents exactly this abort mechanism for mutation B, and adds mutation A to compensate — then leaves in place a check and a comment that assert the opposite. The mechanism was understood and its consequence for the neighbouring check was not followed through.

This is the repo's own standard — *a check that cannot fail is not evidence* — broken by a check whose comment insists it is not a formality.

**Fix:** delete it and record that the accept direction is covered by the fatal guard at :391. Or, to keep it locally falsifiable, use a distinct v1 payload and place it *above* that guard so a failure reports instead of aborting.

### 2. Critical — `Tests/BadgeDiffTests.swift:497` and `:530`: privacy checks with nothing to prove

```swift
checkBool("a refusal never names the user's file", safe.contains("AcmeCorp_NDA_signed"), false)
```

`OverlayStateError` is `case unsupportedVersion(Int)`. There is no String in the type. Verified directly:

```
String(describing:) -> unsupportedVersion(2)
contains a path?    -> false
```

No mutation of this code can redden either check — C, D and E all leave both green.

The comment at :486-487 says the new path "has to clear the same privacy bar as the DecodingError one above". It does not clear the same bar; it clears a trivial one. The DBSYNC-73 check it is modelled on is evidence precisely because of its paired control at :450 (`the raw description WOULD have leaked it`), and no such control exists or can exist here.

`:530` is additionally a near-duplicate of :448 — same `typeMismatch` branch, so nothing can redden one without the other.

**Fix:** keep them as tripwires for a future `OverlayStateError` case carrying a String, but say that in the comment. Do not present them as the same class of evidence as :448-450.

---

### 3. Warning — `SyncExtension.swift:179`: the caller calls a deliberate refusal a decode failure

The PR's claim that this file needs no edit is **behaviourally correct** — verified line by line: the `catch` at :169 takes any `Error`, `lastDecodeFailed` at :178 gates the log to the transition, and both `state` and `lastUpdatedAt` are cleared at :183-184. All three acceptance criteria hold.

The message does not:

```
DropboxSync FinderSync: cannot decode <path> — no badges will be shown. unsupportedVersion(2)
```

`BadgeDiff.swift:302` documents `OverlayStateError` as "a refusal to use `overlay_state.json`, **distinct from a failure to parse it**", and the stated reason for the type existing is that a wrong reason sends the reader down the DBSYNC-76 sandbox path. For the `futureVersion` case the file parsed perfectly — "cannot decode" is that wrong reason, with the right detail appended after it. A neutral verb costs one line and makes the design claim true rather than nearly true.

### 4. Warning — `BadgeDiff.swift:328-332`: the edited doc comment argues against itself

It now reads "it is guarded anyway because <a reshaped `paths` against an older appex> ... is precisely the gap that `supportedVersion` now closes" — offering a closed gap as the reason for keeping the redaction. Confirmed empirically: a reshaped-`paths` v2 now throws `unsupportedVersion(2)` and never reaches that branch.

The real remaining reason is better than the one deleted, and is now unstated: the `typeMismatch` branch is reached by a **same-version** reshape — a writer that changes `paths` and forgets to bump. No version guard can ever catch that.

### 5. Suggestion — `BadgeDiff.swift:285-294`: the fallback's stated scope is wider than its actual scope

`VersionHeader.version` is `Int`, so a v2 that changes `version` to a string (semver) defeats the retry and logs `typeMismatch(expected Int)` with no hint that `version` is the field. A rename degrades better — `keyNotFound(version)` keeps the key. The PR body discloses "unparseable against v1, not unparseable at all"; the code comment does not, and the comment is what the next maintainer reads.

### 6. Suggestion — `BadgeDiff.swift:288-293`: "only when something has already failed" understates it

The DBSYNC-86 scenario is not transient — a lingering appex persists until reinstall — and the `updatedAt` short-circuit at `SyncExtension.swift:193` only guards the success path. So a reshaped v2 costs two full parses every 2 seconds, indefinitely. The decision to optimise the healthy path is still right; the framing as "rare" is wrong in exactly the case the ticket is about.

### 7. Suggestion — `BadgeDiff.swift:338`: `if case` loses exhaustiveness

Adding a second `OverlayStateError` case compiles clean and silently falls through to `"\(type(of: error))"` — the outcome the comment three lines above exists to prevent, and a leak vector if that case carries a path. A `switch` makes it a compile error.

### 8. Suggestion — `Tests/BadgeDiffTests.swift:494`: `safe.contains("2")` is looser than its label

"names the version found" also passes on `unsupportedVersion(12)`. `checkOptional(..., safe, "unsupportedVersion(2)")` is exact and no weaker against any mutation.

### 9. Suggestion — `overlay_state.rs:48`: "both readers" against "every native surface" 15 lines above

Everything the new hunk says about `scope.rs` is **verified accurate**: `StatusHeader` reads only `#[serde(default)] sync_folder`, unknown fields still parse, and a missing root yields `Target::None` — menu hidden, as claimed. `shell-overlay/` is a README only. But :34-36 still advertises three surfaces including a Windows status column. "The two that exist today" resolves it; the aspirational text is pre-existing.

### 10. Note — the mutation table mixes trees

Mutation C's recorded output is from `617d797`; on the merged HEAD it reddens three checks, not two. Not a defect — it predates slice 2 — but the table presents six mutations as one list against one tree, and this repo has already shipped a false green from a stale mutation. Name the commit each was run against.

D and E were reproduced byte-for-byte against HEAD and match the PR exactly.

---

**Verdict: changes requested.** Findings 1 and 2 must be fixed — a PR arguing for mutation-tested evidence cannot ship two checks that no mutation can redden, and three comments that claim otherwise. 3 through 9 are worth doing in the same pass; they are all comments or one-line changes.

mbsanchez:
## Review findings addressed — `6f05f7a`

All ten, plus one the review implied and did not name. **The mutation table in the PR description above is superseded by the one at the bottom of this comment**, which is run entirely against `6f05f7a`.

### The two critical ones

**1. The accept-direction check is gone.** It could not fail: `json` is decoded thirty lines earlier by a guard that `exit(1)`s, so reaching the check proved the decode had already succeeded. The reasoning is recorded in place, including the observed output when the guard is inverted, so the next person does not re-add it:

```
OverlayState.decode (DBSYNC-73)
  FAIL OverlayState.decode threw on a valid payload
```
— and the run exits before the version-guard section prints at all. The accept direction *is* covered, by that fatal guard. A check placed after it can only ever be decorative.

**2. The privacy checks are honest about what they are.** One kept as a tripwire for a future `OverlayStateError` case that carries a String, and now labelled `(tripwire — see above)` with a comment saying plainly that no mutation of today's code can redden it and that it is weaker than the DBSYNC-73 check it was modelled on. The second was a duplicate — same error type, same branch — and is removed.

### The rest

**3.** The caller no longer calls a refusal `cannot decode`. The verb comes from `logVerb(for:)`.

**And the part the review did not have to point out:** my first fix put that branch inline in `SyncExtension.swift` — the one file `run-badge-diff-tests.sh` cannot compile. In a change whose entire argument is that unreachable code is where this project's defects survive review. It now lives in `BadgeDiff.swift` with two checks and two mutations behind it.

**4.** `logSafeDescription`'s doc no longer offers a gap this branch closes as the reason for its own existence. The real reason, previously unstated: the `typeMismatch` branch is reached by a **same-version** reshape — a writer that changes `paths` and forgets to bump. No version guard can ever catch that, which is why the redaction has to stand on its own.

**5.** `VersionHeader.version` is `Int`, so a v2 that turns `version` into a semver string defeats the retry and logs `typeMismatch(expected Int)`. Documented at the field. A *rename* degrades better — `keyNotFound(version)` — and that is noted too.

**6.** "pays the second parse only when something has already failed" now says what is actually true: a mismatched pair persists until reinstall and the `updatedAt` short-circuit only guards the success path, so it is two parses every 2 seconds for as long as the mismatch lasts. The trade is still right; it is a trade.

**7.** `if case` → `switch`, so a second `OverlayStateError` case becomes a compile error instead of silently falling through to the bare type name.

**8.** `contains("2")` → exact `"unsupportedVersion(2)"`.

**9.** `overlay_state.rs` says "both readers **that exist today**", and says why the paragraph fifteen lines above promises three surfaces.

**10.** One table, one tree — below.

### Checks: 66 → 65

Three removed, two added. **Fewer checks and more evidence**, which is the right direction: the three that went could not be reddened by anything.

```
./Tests/run-badge-diff-tests.sh   65 pass   (58 on main at fa8423d)
cargo test                       197 pass   unchanged
cargo fmt --check                clean
```

### Mutations — eight, all against `6f05f7a`, each on a tree verified identical to baseline first

| | Mutation | Reddens |
|---|---|---|
| A | delete the version guard | `an unknown version should be refused, not applied as v1` |
| B | `supportedVersion = 2` | `OverlayState.decode threw on a valid payload` — aborts the run, see below |
| C | remove the `OverlayStateError` switch from `logSafeDescription` | `the refusal is reported exactly…` + `an unreadable v2 is reported as a version problem` |
| D | drop `version != supportedVersion` on the fallback | **three DBSYNC-73 checks** |
| E | remove the retry entirely | `an unreadable v2…` + `and not as the parse failure that happened first` |
| F | fire the retry on an unreadable header | `malformed bytes are still reported as corruption` |
| G | `logVerb` always `"cannot decode"` | `a refusal is called a refusal, not a decode failure` |
| H | `logVerb` always `"refusing"` | `a real parse failure is still called a decode failure` |

**B is listed for completeness and proves less than the others.** It aborts at the DBSYNC-73 guard, so the version-guard section never runs — which is precisely the mechanism that made the deleted check worthless. A is what covers the refuse direction.

**D remains the strongest evidence here and it came free:** the three checks it reddens were written for DBSYNC-73, before this branch existed. An ungated fallback relabels every genuine v1 failure as a version problem.

**F, G and H exist to prove their targets can fail at all** — F for the malformed-bytes control, G and H for the two verb directions in opposite senses. A `logVerb` stuck on either constant passes one of the two checks, so both are needed.

### Still not proven, unchanged from before

Nothing writes v2, so nobody has watched a real Finder refuse a real file. All of this is about the decoder. `SyncExtension.swift`'s own behaviour — one line per transition, `state` and `lastUpdatedAt` cleared — was verified by reading during review, not by a check, because that file is not compilable by the script.

mbsanchez:
## Re-review findings addressed — `a329d81`

Both blocking findings were in code the *previous fix* introduced, which is the part of a review round easiest to wave through. The PR description's stale mutation table has been replaced rather than patched; **the table at the bottom of this comment is the authoritative one**, run entirely against `a329d81`.

### The two I'd block on, both mine, both introduced while fixing the last round

**1. `logVerb` stole `logSafeDescription`'s doc comment.** I inserted the new function immediately above `func logSafeDescription` — which sat directly below its own 35-line `///` block, with no separator. Swift fused them. The entire DBSYNC-73 privacy rationale — the `codingPath` leak, the `AcmeCorp_NDA_signed.pdf` example, `NSLog` and the public unified log, "it never emits `codingPath`, and never a value", *and the redaction argument I rewrote in that same commit for finding 4* — all became documentation for a two-line function that returns one of two string literals and redacts nothing. `logSafeDescription` was left with no doc at all.

The previous round was about three comments that were factually wrong. This is the same failure at 35× the size, committed while fixing it.

```
$ grep -n -B1 "^func logVerb\|^func logSafeDescription" BadgeDiff.swift
341-/// from `logSafeDescription`.
342:func logVerb(for error: Error) -> String {
375-/// version check.
376:func logSafeDescription(of error: Error) -> String {
```

**2. `logVerb` reintroduced the exact pattern the same commit removed.** It used `error is OverlayStateError`, which absorbs every future case into `"refusing"` silently — while thirty lines below, `logSafeDescription` was being converted from `if case` to an exhaustive `switch` for precisely that reason. A future case describing a genuine parse failure would have been logged as a refusal: the confusion the function exists to prevent.

Now a `switch`. Adding a second case is a build failure in **both** functions, verified:

```
BadgeDiff.swift:345:5: error: switch must be exhaustive   ← logVerb
BadgeDiff.swift:389:9: error: switch must be exhaustive   ← logSafeDescription
```

### The rest

**3. The tripwire comment still overstated itself** — third attempt at this comment, and the first two both claimed more than the check delivers. It promised to catch a future `OverlayStateError` case carrying a path. Nothing in the suite constructs such a case, so adding one leaves every check green. What it actually catches is a String added to `.unsupportedVersion` itself or to its `logSafeDescription` branch — and even there it is not independent, since the exact `checkOptional` below it reddens on the same input. Both facts are now in the comment.

**4. The refusal-verb check ran on a hand-built literal** and sat orphaned between two unrelated blocks. It now runs on the error `decode` actually throws, inside the refusal's own `catch`.

**5. "anything placed after line ~391 is unreachable" overreached.** True for that payload under that mutation; a check using a *different* v1 payload would be live. The comment now says the accept direction is covered by that `guard` and a second check on the same payload is redundant — not impossible. It also names the guard instead of a line number that will drift.

**6.** The dedup note assumed `OverlayStateError` will always have one case; it says so now. **7.** `logSafeDescription`'s "the case that still reaches `typeMismatch`" was not the only one — the retype route documented thirty lines above lands there too, with a `codingPath` of `version` that holds nothing of the user's.

### The table I published for `6f05f7a` was wrong about F

It claimed F reddens one check. It reddens two, and unavoidably:

```
FAIL malformed bytes are still reported as corruption — expected true, got false
FAIL a real parse failure is still called a decode failure — expected Optional("cannot decode"), got Optional("refusing")
```

Any mutation that routes malformed bytes through `OverlayStateError` also flips the verb. That is a *better* result than I reported — the `cannot decode` check does double duty for F the way D's DBSYNC-73 checks do — and reporting it as one row was a factual error in a table whose entire purpose is exactness.

### Mutations — eight, all against `a329d81`, each on a tree verified identical to baseline first

| | Mutation | Reddens |
|---|---|---|
| A | delete the version guard | `an unknown version should be refused, not applied as v1` |
| B | `supportedVersion = 2` | `OverlayState.decode threw on a valid payload` — aborts the run |
| C | remove the `OverlayStateError` switch from `logSafeDescription` | `the refusal is reported exactly…` + `an unreadable v2 is reported as a version problem` |
| D | drop `version != supportedVersion` on the fallback | **three DBSYNC-73 checks** |
| E | remove the retry entirely | `an unreadable v2…` + `and not as the parse failure that happened first` |
| F | fire the retry on an unreadable header | `malformed bytes are still reported as corruption` + `a real parse failure is still called a decode failure` |
| G | `logVerb` always `"cannot decode"` | `a refusal is called a refusal, not a decode failure` |
| H | `logVerb` always `"refusing"` | `a real parse failure is still called a decode failure` |

G and H redden one check each, in opposite directions — a `logVerb` stuck on either constant passes exactly one of the two, so both checks are needed and neither is green by construction.

B proves less than the others and is listed for completeness: it aborts at the DBSYNC-73 guard, so the version-guard section never runs. A is what covers the refuse direction.

### Counts

```
./Tests/run-badge-diff-tests.sh    65 pass    (58 on main at fa8423d)
cargo test                        197 pass    unchanged; no Rust change in this commit
```

CI was green on `6f05f7a` across all five jobs, and the `macos-app` log confirms both that `SyncExtension.swift` compiles into the appex and that the new checks run on the runner — worth verifying explicitly, because `ci.yml:161` runs only the badge-diff script and that file is not otherwise compiled by any check.

### Still not proven

Nothing writes v2, so nobody has watched a real Finder refuse a real file. All of this is about the decoder. `SyncExtension.swift`'s transition behaviour — one line per transition, `state` and `lastUpdatedAt` cleared — was verified by reading during review, not by a check.

mbsanchez:
## Review: approved — third pass, `a329d81` (+ `5b43720`)

Independent review, three rounds. Both earlier rounds blocked, and the findings that mattered most were reproduced by hand rather than accepted on description.

**Round 3 verified, not read:**

- The doc-comment move is **lossless** — checked by set-comparing every line of the doc region against `6f05f7a`, not by eyeballing the seam. The only removed lines are the two sentences deliberately rewritten; no line duplicated. The DBSYNC-73 material is intact and attached to `logSafeDescription`.
- Exhaustiveness holds in **both** functions, at the cited lines.
- The tripwire comment's three claims tested independently — a String added to `.unsupportedVersion` reddens; a future case carrying a path does **not** (65 pass, all green); the check is not independent of the exact-match one beside it. Three for three. The first version of this comment that describes what the check does.
- **All eight mutation rows reproduce exactly**, F included.

`5b43720` takes two of the three cosmetic nits, both about comment precision — the block they are in has already carried three wrong versions, so tightening it is cheap insurance rather than polish. The `guard let` / `if let` stylistic difference is left alone.

### A correction the reviewer volunteered

Round 2 stated that CI never compiles `SyncExtension.swift`. That was wrong, and it is the kind of wrong that matters: it would have made a swapped format-string argument look unverifiable when it is in fact checked on every PR. `bundle:dev` → `beforeBuildCommand` → `build:finder-sync` → `build-appex.sh` → `xcodebuild`, and the `.xcodeproj` references the file. Confirmed in the `macos-app` log.

### What this PR cost, honestly

Three review rounds on ~180 lines. Round 1 found two checks that no mutation could redden, in a change whose entire argument was mutation-tested evidence. Round 2 found that the fix for round 1 had fused a 35-line privacy doc onto the wrong function and reintroduced, in a new function, the exact `is`-instead-of-`switch` pattern it had just removed from another. Round 3 found nothing of that class.

The pattern worth recording: **both blocking findings after round 1 were in code written to fix round 1.** Fix commits got less scrutiny than the original, and that is where the defects went.

### Final state

```
./Tests/run-badge-diff-tests.sh    65 pass    (58 on main at fa8423d)
cargo test                        197 pass    unchanged
cargo fmt --check                 clean
CI                                5/5 green on a329d81
```

The `macos-app` log confirms both that the appex compiles and that the new checks execute on the runner — not inferred from a green tick.

**Still not proven, and it should stay stated:** nothing writes v2, so nobody has watched a real Finder refuse a real file. Every check here is about the decoder. `SyncExtension.swift`'s transition behaviour — one log line per transition, `state` and `lastUpdatedAt` cleared — was verified by reading, not by a check.

Moving to QA.

mbsanchez:
## QA Result: PASSED ✅ — with two acceptance criteria explicitly not covered by any check

`automated-only` branch. Every command run from a verified-clean tree at `5b43720` (`git status` → 0 modified files).

```
run-badge-diff-tests.sh                    exit 0    65 checks pass
cargo test                                 exit 0    197 pass, 0 failed
npm --workspace apps/desktop run build     exit 0    (typecheck + vite build)
CI on 5b43720                              5/5 green
```

The frontend command failed on the first attempt with `sh: tsc: command not found` — a stale local `node_modules` with no `.bin` directory, not a defect. Re-run after `npm ci` and it passes. Recording it because "I ran it and it failed" is not something to leave out of a QA report even when the cause turns out to be local.

### Acceptance criteria → the evidence that actually covers each

| AC | Covered by |
|---|---|
| Decision recorded before code | Ticket description, dated and reasoned. Process artifact |
| Reader acts on `version`: 1 accepted, else refused | `an unknown version should be refused…` + mutation **A** (delete guard) and **B** (invert) |
| Accepted version is a named constant | `OverlayState.supportedVersion` — read, not tested; a literal would pass every check |
| `logSafeDescription` reports `unsupportedVersion(N)` | exact-match `checkOptional` + mutation **C** |
| Log names the version, never a path | tripwire check + mutation **C**; existing DBSYNC-73 privacy checks still green |
| **Exactly one log line, on the transition** | ⚠️ **not covered — see below** |
| **`state` and `lastUpdatedAt` both cleared** | ⚠️ **not covered — see below** |
| Bump rule names both readers, `scope.rs` included | `overlay_state.rs` doc comment — read, not tested |
| Windows gap accepted | Recorded on the ticket with reasoning |
| Retry names the version when v1 cannot parse | `an unreadable v2 is reported as a version problem` + mutation **E** |
| Non-version failures rethrow unchanged | mutation **D** — reddens three DBSYNC-73 checks |
| Malformed bytes still report `dataCorrupted` | negative control + mutation **F**, which proves it can fail |
| Happy path parses once | Control flow, read. This is what the AC asked for — it explicitly forbids claiming a benchmark |
| Both suites pass, counts stated | 65 (from 58 on `main`), 197 unchanged |

### The two that are not covered, and why it matters

**"Exactly one log line, on the transition"** and **"`state` and `lastUpdatedAt` both cleared"** live in `SyncExtension.swift`. `run-badge-diff-tests.sh` compiles `BadgeDiff.swift` and its checks with `swiftc` and nothing else, so no check in this repo can reach `reloadState()`. Both were verified by **reading** during code review — the `catch` takes any `Error`, `lastDecodeFailed` gates the `NSLog`, both fields are cleared — and reading is weaker evidence than a check that can fail.

This is the same class of gap the whole ticket is about, so it should not be waved through quietly. It is *not* new to this PR: those lines are pre-existing and DBSYNC-73 shipped under the same limitation.

**It is stageable manually, and cheaply.** The writer is event-driven (`refresh_overlay_state_internal` fires from the sync pipeline and cloudsc ops), not on a timer, so with the app quit the file stays as edited:

1. Quit the app. The Finder extension keeps running and polling every 2s.
2. Edit `overlay_state.json` — change `"version": 1` to `"version": 2`.
3. `log stream --predicate 'eventMessage CONTAINS "DropboxSync FinderSync"'`

Expected: **one** line, `refusing …/overlay_state.json — no badges will be shown. unsupportedVersion(2)`, not one every 2 seconds — and badges disappear. Change it back and a single `decoding … again.` line follows.

That would also be the first time anyone has watched this behaviour rather than inferred it, since nothing in the product writes v2. Offered, not required: the ACs it covers were reviewed by reading, and the ticket is honest about the limitation either way.

### Verdict

**PASS** on everything a check can reach. Two ACs rest on review-by-reading, stated here rather than folded into a green tick.

PR #137, opened 2026-08-30
fix(#DBSYNC-56): mark the index row for rescan when an upload is cancelled mid-flight

mbsanchez:
## Summary

A local edit can reach Dropbox never, silently, while the badge says `synced`.

A file synced at `h1` is edited to `h2`. The scan optimistically advances the index row to `h2` and enqueues an upload. A backup agent or an atomic re-save round-trips the file out and back **byte-identical**, and the upload drains inside that window — the `had_remote` NotFound branch leaves the row and returns `Ok(())`. That `Ok` is deliberate: returning `Err` there is what produced the phantom "Error" DBSYNC-55 fixed.

End state: **index `h2`, disk `h2`, remote `h1`.**

## Why it is permanent, not slow

Both recovery paths were checked in the source rather than assumed:

- `process_local_file_change` decides "changed" by `prev.hash != hash` — **index against disk**. Both are `h2`.
- `reconcile_remote_present` only acts when the **remote** moved (`prev.content_hash != remote_meta.content_hash`). It never compares the local index against the remote.

So nothing re-detects it. If nobody touches that file on Dropbox again, the divergence lasts forever. If the user later edits to `h3`, that upload succeeds and `h2` is simply absent from the file's history.

And the badge reads `synced` throughout, because the overlay tier is computed from the same row that is wrong. DBSYNC-84 settled that a badge which is confidently wrong is worse than no badge — this is that, on the one thing a sync client exists to get right.

## The fix

`Db::HASH_NEEDS_RESCAN` marks the row's hash unusable in that branch, so the next scan cannot walk past it.

**The row survives.** Deleting it is cleaner-looking and breaks deletion detection: the row is what proves a tracked file has disappeared from disk. Without it a genuine deletion never reaches Dropbox — the same data loss pointing the other way. The `!had_remote` arm still drops the row, because a path with nothing on Dropbox has nothing to recover.

**Bounded retries were the ticket's other option and do not fix this.** They cover a time window; a file returning after the budget lands in exactly the same state. Marking closes it regardless of timing.

## Two other readers had to learn the marker

Otherwise the fix trades silent loss for visible litter:

- **`download_would_conflict`** already documents a `None` baseline as *deliberately* not-a-conflict. A marked row is the same epistemic state — we do not know what was last synced — so it maps onto that arm rather than a third case. Without this, every download during the window manufactures a conflicted copy.
- **`reconcile_remote_absent`** cannot answer honestly with a marked row: propagating the delete could destroy an edit Dropbox never received, and the conflict arm would hand the user a conflict record for an event they never saw. It now returns `Ok(0)` and lets the next scan resolve it with real data — and **leaves the remote row in place**, so the next sweep asks again rather than forgetting the path.

## The ticket's NIT, which was two sites and not one

`cleanup_stale_upload_state` used `Path::exists()`, which reads a permission denial or a mid-rename atomic save as "gone". Now a confirmed `NotFound`, matching `classify_path`. The ticket named one site; there are two, at `sync_pipeline.rs:977` and `:1003`.

## Stack

`desktop-tauri` — Rust core only. No IPC surface, no frontend, no schema change.

## Jira Ticket

DBSYNC-56 — https://manuelbsan.atlassian.net/browse/DBSYNC-56

## Test Plan

- [x] `cargo test` — **197 → 203**, all passing. 197 was measured on `main` before any edit.
- [x] `cargo clippy` — **4 lib warnings, unchanged from `main`.** A naive count says 6; the two extra lines are clippy's own summary and the pre-existing `block v0.1.6` notice.

### Mutations — four, each on a tree verified identical to baseline first, reverted before the next

**M1 — remove the marking** (restore the old branch):
```
FAILED  vanished_file_with_a_remote_copy_keeps_its_row_and_marks_it_for_rescan
```
Read what stays **green**: `a_row_marked_for_rescan_is_re_detected_and_re_uploaded`. That is the intended separation — one test proves the branch sets the marker, the other proves a marked row actually recovers. Neither subsumes the other, and it took a mutation to show it rather than an assertion in a comment.

**M2 — treat the marker as an ordinary differing baseline**:
```
FAILED  download_would_conflict_false_when_baseline_is_the_rescan_marker
```

**M3 — remove the `reconcile_remote_absent` guard**:
```
FAILED  reconcile_remote_absent_does_nothing_while_the_row_is_marked_for_rescan
```

**M4 — `is_confirmed_absent` always returns false**:
```
FAILED  is_confirmed_absent_distinguishes_missing_from_present
FAILED  cleanup_clears_editor_temp_rows_and_phantom_failed_jobs
```
The second is **pre-existing**. That is better evidence than the test written beside the change: it shows the NIT preserves behaviour that was already specified.

## The risk from the plan, checked rather than waved through

The plan flagged that `cloudsc_ops.rs` re-upserts an existing row's hash in five places and said it must be confirmed, not assumed. It was:

- All five are **rollback/restore** paths rewriting the row's own just-read hash, so a marked row carries its marker through them — correct, the file still needs re-detection.
- `process_local_file_change` returns early for dehydrated placeholders (*"never hash, never enqueue, never delete"*), so a marked row on a placeholder is inert and cannot cause a spurious upload.

**Not fully traced, and said plainly rather than glossed:** dehydrating a file whose row is marked discards local content Dropbox never received. I believe that is a pre-existing hazard of dehydration rather than something this change introduces, but I did not follow that path to the end — please do not take my word for it.

## What this does not prove

Nobody has reproduced this against a real backup agent. The tests drive the state machine directly, so they establish the logic — not that the unlink window is as wide as anyone believes it to be.

## A cost worth naming

The marker widens `local_file_index.hash`'s contract from "a content hash" to "a content hash, or this marker", and every future reader has to know. That is a real tax. A nullable column expresses it properly, which needs the migration system that does not exist yet (DBSYNC-40) — noted in the constant's doc comment so the debt is visible where someone will read it.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/dropbox_transfer.rs
+++ b/apps/desktop/src-tauri/src/dropbox_transfer.rs
@@ -1486,11 +1486,14 @@ mod tests {
         super::upload_local_file_internal(&state, "~$draft.docx", 1).expect("must not error");
 
         assert!(
-            state.db.get_local_file("~$draft.docx").expect("query").is_none(),
+            state
+                .db
+                .get_local_file("~$draft.docx")
+                .expect("query")
+                .is_none(),
             "a path with no remote copy has nothing to recover and must be dropped"
         );
     }
-
 
     // `fetch_and_write_file` performs live HTTP calls against the Dropbox API
     // and is intentionally left to manual QA (large-file download against a
--- a/apps/desktop/src-tauri/src/dropbox_transfer.rs
+++ b/apps/desktop/src-tauri/src/dropbox_transfer.rs
@@ -511,12 +511,20 @@ fn download_would_conflict(
     }
     match baseline_hash {
         None => false, // never indexed locally — nothing to protect against
-        // A row marked for rescan (DBSYNC-56) is the SAME epistemic state as `None`: we do
-        // not know what was last synced, so nothing here proves a local modification.
-        // Falling through to the comparison below would make it differ from every on-disk
-        // hash and manufacture a conflicted copy for a window the user never saw — trading
-        // silent data loss for visible litter.
-        Some(baseline) if baseline == crate::storage::db::Db::HASH_NEEDS_RESCAN => false,
+        // A row marked for rescan (DBSYNC-56) is CONFLICT, and it is worth being exact
+        // about why, because a first version of this got it backwards and would have lost
+        // bytes.
+        //
+        // `None` and the marker look alike — neither gives a usable baseline — but they
+        // carry opposite information. `None` means there is no row at all: nothing was ever
+        // recorded, so nothing proves a local modification. The marker means a row that WAS
+        // advanced for a real local edit whose upload was then cancelled, so these bytes may
+        // be content Dropbox has never received. Overwriting them is precisely the loss this
+        // ticket exists to prevent.
+        //
+        // The cost is bounded: `conflict_copy_with_content_exists` at the call site dedupes
+        // by content, so this is one copy per distinct on-disk content, not one per download.
+        Some(baseline) if baseline == crate::storage::db::Db::HASH_NEEDS_RESCAN => true,
         Some(baseline) => on_disk_hash != baseline,
     }
 }
@@ -887,7 +895,7 @@ pub(crate) fn upload_local_file_internal(
             let had_remote = state.db.get_remote_file(relative)?.is_some();
             if !had_remote || is_ignored_local_path(relative) {
                 state.db.remove_local_file(relative)?;
-            } else if let Some(row) = state.db.get_local_file(relative)? {
+            } else {
                 // DBSYNC-56. The row is kept — deletion detection needs it — but its hash is
                 // marked for rescan, because leaving it intact loses the edit outright.
                 //
@@ -903,12 +911,7 @@ pub(crate) fn upload_local_file_internal(
                 // Bounded retries were considered and rejected: they only cover a time
                 // window, and a file returning after the budget lands in exactly this state.
                 // Marking the row closes it regardless of timing.
-                state.db.upsert_local_file(
-                    relative,
-                    crate::storage::db::Db::HASH_NEEDS_RESCAN,
-                    row.size_bytes,
-                    row.modified_ts,
-                )?;
+                state.db.mark_local_file_for_rescan(relative)?;
             }
             return Ok(());
         }
@@ -1131,7 +1134,18 @@ pub(crate) fn download_remote_file_internal(state: &AppState, path_display: &str
         let (on_disk_hash, _size, _mtime) = hash_file(&target)?;
         let baseline_hash = state.db.get_local_file(&relative)?.map(|row| row.hash);
 
-        if download_would_conflict(true, &on_disk_hash, baseline_hash.as_deref())
+        // Nothing can be lost by overwriting bytes with identical bytes, whatever the
+        // baseline says (DBSYNC-56). The remote row was advanced by the sweep that enqueued
+        // this job, so it holds the content about to be fetched. Without this, a row marked
+        // for rescan — which the guard below now correctly treats as a conflict — would
+        // manufacture a conflicted copy of content that already matches the remote exactly.
+        let remote_matches_disk = state
+            .db
+            .get_remote_file(&relative)?
+            .is_some_and(|r| r.content_hash == on_disk_hash);
+
+        if !remote_matches_disk
+            && download_would_conflict(true, &on_disk_hash, baseline_hash.as_deref())
             && !conflict_copy_with_content_exists(&target, &on_disk_hash)?
         {
             let conflicted_path = create_conflicted_copy(&target)?;
@@ -1792,15 +1806,19 @@ mod tests {
         assert!(!download_would_conflict(true, "on-disk-hash", None));
     }
 
-    /// DBSYNC-56. A row marked for rescan means "we do not know what was last synced",
-    /// which is the same answer as `None` — not "the baseline differs".
+    /// DBSYNC-56, and the assertion is deliberately the opposite of what a first version of
+    /// this change shipped. That version returned `false` here, reasoning that the marker
+    /// is "the same epistemic state as `None`". It is not.
     ///
-    /// Without this, every download arriving while the marker is set would manufacture a
-    /// conflicted copy, because the marker differs from every real on-disk hash. That would
-    /// trade silent data loss for visible litter, which is not a trade worth making.
+    /// `None` means there is no row at all — nothing was ever recorded, so nothing proves a
+    /// local modification. The marker means a row that WAS advanced for a real local edit
+    /// whose upload was cancelled: these bytes may be content Dropbox has never received.
+    /// Returning `false` let an incoming download overwrite exactly that content, with no
+    /// conflicted copy and nothing logged — the same silent loss this ticket exists to fix,
+    /// re-created by its own fix.
     #[test]
-    fn download_would_conflict_false_when_baseline_is_the_rescan_marker() {
-        assert!(!download_would_conflict(
+    fn download_would_conflict_true_when_baseline_is_the_rescan_marker() {
+        assert!(download_would_conflict(
             true,
             "on-disk-hash",
             Some(crate::storage::db::Db::HASH_NEEDS_RESCAN)
--- a/apps/desktop/src-tauri/src/dropbox_transfer.rs
+++ b/apps/desktop/src-tauri/src/dropbox_transfer.rs
@@ -511,6 +511,12 @@ fn download_would_conflict(
     }
     match baseline_hash {
         None => false, // never indexed locally — nothing to protect against
+        // A row marked for rescan (DBSYNC-56) is the SAME epistemic state as `None`: we do
+        // not know what was last synced, so nothing here proves a local modification.
+        // Falling through to the comparison below would make it differ from every on-disk
+        // hash and manufacture a conflicted copy for a window the user never saw — trading
+        // silent data loss for visible litter.
+        Some(baseline) if baseline == crate::storage::db::Db::HASH_NEEDS_RESCAN => false,
         Some(baseline) => on_disk_hash != baseline,
     }
 }
@@ -881,6 +887,28 @@ pub(crate) fn upload_local_file_internal(
             let had_remote = state.db.get_remote_file(relative)?.is_some();
             if !had_remote || is_ignored_local_path(relative) {
                 state.db.remove_local_file(relative)?;
+            } else if let Some(row) = state.db.get_local_file(relative)? {
+                // DBSYNC-56. The row is kept — deletion detection needs it — but its hash is
+                // marked for rescan, because leaving it intact loses the edit outright.
+                //
+                // The index row was already optimistically advanced to the new content when
+                // the change was detected. If the file comes back BYTE-IDENTICAL (a backup
+                // agent or an atomic re-save round-tripping it), index and disk agree on
+                // content the remote has never received, and nothing notices: the local scan
+                // compares index against disk, and `reconcile_remote_present` only fires when
+                // the REMOTE moves. The file stays diverged for as long as nobody touches it
+                // on Dropbox again — while the badge reads `synced`, because it is computed
+                // from this same row.
+                //
+                // Bounded retries were considered and rejected: they only cover a time
+                // window, and a file returning after the budget lands in exactly this state.
+                // Marking the row closes it regardless of timing.
+                state.db.upsert_local_file(
+                    relative,
+                    crate::storage::db::Db::HASH_NEEDS_RESCAN,
+                    row.size_bytes,
+                    row.modified_ts,
+                )?;
             }
             return Ok(());
         }
@@ -1388,6 +1416,82 @@ mod tests {
     use crate::path_util::hash_file;
     use std::sync::atomic::{AtomicUsize, Ordering};
 
+    /// An `AppState` on an isolated temp DB with `sync_folder` set. The NotFound branch
+    /// under test returns before `get_access_token`, so these tests make no network call.
+    fn build_state(root: &std::path::Path) -> crate::state::AppState {
+        let sync_folder = root.join("synced");
+        std::fs::create_dir_all(&sync_folder).expect("create sync folder");
+        let db_path = root.join("db").join("app.db");
+        std::fs::create_dir_all(db_path.parent().unwrap()).expect("create db dir");
+        let db = crate::storage::db::Db::new_at(&db_path).expect("db init");
+        db.set_sync_folder(&sync_folder.to_string_lossy())
+            .expect("set sync folder");
+        crate::state::AppState {
+            secure_store: crate::storage::secure_store::SecureStore::new(),
+            db: std::sync::Arc::new(db),
+            sync_engine: std::sync::Arc::new(std::sync::Mutex::new(
+                crate::sync::engine::SyncEngine::new(),
+            )),
+            token_cache: std::sync::Arc::new(std::sync::Mutex::new(None)),
+            scheduler_started: std::sync::Arc::new(std::sync::Mutex::new(false)),
+            oauth_listener: std::sync::Arc::new(std::sync::Mutex::new(None)),
+            sync_running: std::sync::Arc::new(std::sync::atomic::AtomicBool::new(false)),
+            token_refresh_lock: std::sync::Arc::new(std::sync::Mutex::new(())),
+            http_client: crate::state::build_http_client(),
+        }
+    }
+
+    /// DBSYNC-56, the branch itself: a synced file that vanishes mid-upload keeps its row,
+    /// and that row is marked so the next scan cannot walk past it.
+    #[test]
+    fn vanished_file_with_a_remote_copy_keeps_its_row_and_marks_it_for_rescan() {
+        let tmp = tempfile::tempdir().expect("tempdir");
+        let state = build_state(tmp.path());
+        // Synced at H1 remotely; the index was already optimistically advanced to H2 when
+        // the edit was detected. The file is NOT on disk — this is the unlink window.
+        state
+            .db
+            .upsert_remote_file("report.docx", "H1", "rev1", 0)
+            .expect("seed remote");
+        state
+            .db
+            .upsert_local_file("report.docx", "H2", 42, 7)
+            .expect("seed local");
+
+        super::upload_local_file_internal(&state, "report.docx", 1).expect("must not error");
+
+        let row = state
+            .db
+            .get_local_file("report.docx")
+            .expect("query")
+            .expect("the row must SURVIVE - deletion detection needs it");
+        assert_eq!(row.hash, crate::storage::db::Db::HASH_NEEDS_RESCAN);
+        // Size and mtime are carried through: only the hash is invalidated.
+        assert_eq!(row.size_bytes, 42);
+        assert_eq!(row.modified_ts, 7);
+    }
+
+    /// Negative control. A vanished path with nothing on Dropbox is forgotten, exactly as
+    /// before — an editor temp file must not linger in the index carrying a marker that
+    /// makes every future scan re-detect a file that does not exist.
+    #[test]
+    fn vanished_file_with_no_remote_copy_is_still_forgotten() {
+        let tmp = tempfile::tempdir().expect("tempdir");
+        let state = build_state(tmp.path());
+        state
+            .db
+            .upsert_local_file("~$draft.docx", "H2", 42, 7)
+            .expect("seed local");
+
+        super::upload_local_file_internal(&state, "~$draft.docx", 1).expect("must not error");
+
+        assert!(
+            state.db.get_local_file("~$draft.docx").expect("query").is_none(),
+            "a path with no remote copy has nothing to recover and must be dropped"
+        );
+    }
+
+
     // `fetch_and_write_file` performs live HTTP calls against the Dropbox API
     // and is intentionally left to manual QA (large-file download against a
     // real account, verifying flat memory usage and crash-safety).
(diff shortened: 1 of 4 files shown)


mbsanchez:
## Code review: changes requested

Independent review, findings reproduced by hand before acting on them. **This PR exists to stop a silent data loss, and it introduces a different one.** That is the wrong thing for this particular PR to get wrong.

The diagnosis and `HASH_NEEDS_RESCAN` are the right shape. The problem is that the marker's meaning was never settled, and the change answers the same question **opposite ways in two files**.

---

### C1 — Critical: `download_would_conflict` returning `false` for the marker loses bytes — `dropbox_transfer.rs:519`

I mapped the marker onto the `None` arm and justified it as "the same epistemic state". **That is wrong, and it is backwards.**

`None` means *there is no index row* — the path was never tracked, so no local edit was ever recorded. The marker means the opposite: **a row exists, it was advanced for a real local edit, and that edit is known not to have reached Dropbox.** That is not "nothing to protect against". It is the strongest possible reason to preserve a copy before a destructive overwrite.

The regression, concretely:

1. `report.docx` synced at h1, edited to h2, row advanced, upload enqueued.
2. Backup agent unlinks mid-flight → row marked, remote still h1.
3. The file returns **not** byte-identical — another save landed, disk is h4.
4. Remote moves h1 → h3 from another device; a download is enqueued.
5. `download_would_conflict(true, h4, Some(""))` → **`false`** → h4 is overwritten by h3. No conflicted copy, no `sync_conflicts` row, nothing logged.

Before this PR that was `Some("h2")` → `h4 != h2` → **`true`** → the copy was preserved.

**And my stated reason was factually false.** I wrote that treating it as a conflict would make "every download during the window manufacture a conflicted copy". The call site at `:1135` already guards with `conflict_copy_with_content_exists`, which dedupes by content — the litter is bounded to one copy per distinct on-disk content, not one per download. I argued from a cost that does not exist, to justify removing a protection that does.

**Fix:** `Some(baseline) if baseline == HASH_NEEDS_RESCAN => true`. The test at `:1802` and mutation M2 both currently certify the wrong behaviour and must flip with it.

### C2 — Critical: the marker manufactures conflicted copies through an arm I never looked at — `sync_pipeline.rs:139`

`process_local_file_change`'s changed-arm has a sub-arm: if a job is already active on the path, it creates a conflicted copy instead of re-uploading. A marked row **always** enters the changed-arm — that is the whole design — so any concurrently-active job on that path (a queued `download`, or one sitting in `retry_wait`) turns the marker into filesystem litter.

Reproduced against head: seed a marked row, a real file, one queued `download`, run the scan →

```
conflicted copies manufactured: ["report (conflicted copy 20260830075547).docx"]
```

Pre-change this could not happen: index and disk agree, no change detected, no copy.

Note this is the *same* outcome I rejected in C1. **The change currently says "the marker is not a conflict" in one reader and "the marker is a conflict" in another, on identical state.** They have to give the same answer, and C1 establishes which: the marker means unuploaded content exists, so protect on download, and do not treat it as a *new concurrent edit* racing a pending job.

**Fix:** exclude the marker from the `pending_targets` sub-arm and take the plain re-upload path.

---

### W1 — The NIT is untested

`is_confirmed_absent_distinguishes_missing_from_present` covers present vs absent. `Path::exists()` already gets both right. The distinction the function exists for is **absent vs unreadable**, and nothing tests it — reverting the body to `!p.exists()` leaves the suite green at 203. M4 kills the test only through its *absent* assertion, so it certifies nothing about `NotFound` discrimination. The test's name claims a distinction it never makes.

### W2 — A doc comment reattached to the wrong function. Third time this session.

`is_confirmed_absent` was inserted between `cleanup_stale_upload_state`'s doc comment and the function it documented. `cleanup_stale_upload_state` is now undocumented, and the new function's docs open with four lines about a different one.

This is the identical mistake DBSYNC-91 shipped, and then shipped again in the commit that fixed it. I wrote a memo after that one whose second bullet reads *"diff the seams, not just the hunk — inserting a declaration next to a doc comment can silently re-parent it; `grep -n -B1` shows what each doc is attached to."* I did not run it.

### W3 — `reconcile_remote_present` is an unhandled reader — `remote_index.rs:425`

`local.hash != remote_meta.content_hash` is always true for a marked row, so a download is enqueued whenever the remote moves, even when the disk already holds the new remote content. Wasteful alone; it is also the delivery mechanism for C1.

### W4 — `reconcile_remote_absent`'s comment promises a resolution that does not happen — `remote_index.rs:452`

The comment says the next scan "resolves with real data one tick later". It does not. The scan clears the marker and enqueues an upload; at drain time the remote row I deliberately kept is still present while `fetch_remote_file_metadata` returns `None`, so the skip-if-identical check is bypassed and the upload **resurrects a file the user deleted on Dropbox**. Pre-change this was deterministic (conflict arm, nothing re-uploaded); now it is a race whose likelier outcome is silent resurrection.

### W5 — The constant's collision argument is factually wrong — `storage/db.rs:354`

I wrote "every value written here comes from `hash_file`". `cloudsc_ops.rs:366` writes a `local_file_index.hash` taken from **Dropbox's `content_hash`**. The sentinel does not collide, but for a reason I did not name: incidental empty-hash guards in `cloudsc_ops.rs:397` and `remote_index.rs`, none of which knows it is load-bearing for this constant. A `debug_assert!` in `upsert_local_file` would enforce the invariant where it is claimed.

---

### Refuted, in the PR's favour

The PR flagged one thing I could not trace: whether dehydrating a marked file discards content Dropbox never received. **It cannot.** `is_file_fully_synced` (`cloudsc_ops.rs:765`) never reads the index row's hash — it re-hashes the bytes on disk against `remote.content_hash`. A marked file whose content the remote lacks fails that check and `dehydrate_path_internal` returns `Err`. The marker is invisible to that decision. That can now be stated positively instead of hedged.

Also confirmed clean: the marker cannot get stuck or loop (file returns → scan clears; never returns → deletion detection removes the row; restart → `cleanup_stale_upload_state` never touches it), the empty string does not collide with any real value, and `overlay_state.rs` reports `OutOfSync` for a marked row — which is the truthful answer and an improvement on the `synced` the ticket complains about.

Mutations M2 and M4 reproduce byte-for-byte as recorded; the table is accurate. One correction: the PR implies four new tests, there are six.

---

**Verdict: changes requested.** C1 and C2 are two halves of one unsettled question and resolve together. C1 is the one that loses bytes.

mbsanchez:
## Review findings addressed — `e3193b7`

All seven. The two criticals were one problem, and fixing them properly meant settling a question the branch had left open.

### The root cause was not a typo

C1 and C2 disagreed because **the marker's meaning was never decided**. It is now stated once, on the constant, and applied in every reader:

> **The marker means: there is unuploaded local content, and no trustworthy record of what it is.**

- a reader deciding whether to **destroy** local bytes → conflict, preserve them
- a reader that **needs the content** to decide at all → defer to the next scan
- a reader asking **"was there a new edit?"** → no, it is bookkeeping

### C1 — fixed, and the assertion is now the opposite of what shipped

`download_would_conflict` returns `true` for the marker. The review was right on both counts: `None` and the marker look alike and carry opposite information, and my stated cost for treating it as a conflict was **factually false** — `conflict_copy_with_content_exists` at the call site already dedupes by content, so it is one copy per distinct on-disk content, not one per download. I argued from a cost that does not exist to remove a protection that does.

The test flipped with it, and its name changed from `..._false_when_baseline_is_the_rescan_marker` to `..._true_...` so the file no longer documents the wrong behaviour.

### C2 — fixed

The pending-job sub-arm excludes the marker and takes the plain re-upload path. New test, and **M5 proves it**.

### W1 — the NIT is now actually tested

The old test covered present-vs-absent, which `Path::exists()` already gets right — reverting the body left the suite green, exactly as the review showed. The new test exercises **absent vs unreadable**, with the premise asserted so it cannot pass vacuously under root.

### W2 — doc comment reattached

`cleanup_stale_upload_state` has its docs back. Third time this session, and I have a memo from DBSYNC-91 whose second bullet reads *"diff the seams, not just the hunk"*. Writing it down evidently is not the same as running it.

### W5 — plus a design change the review's suggestion could not have had

The collision claim was wrong (`cloudsc_ops` writes a Dropbox-sourced hash, not a `hash_file` one) and is corrected to name what actually protects the sentinel.

The suggested `debug_assert!` **could not work as proposed**: the marker *is* the empty string, so an assert inside `upsert_local_file` cannot tell deliberate marking from an accidental blank — it would have been an assert incapable of failing, which is the thing this repo rejects. So the marker got its own operation, `mark_local_file_for_rescan`, which is now the only route in. That makes the assert meaningful, and it has a `#[should_panic]` test so it is a guard rather than a decoration.

### W3 — and a bug I nearly introduced fixing it

My first attempt deferred in `reconcile_remote_present` when the marker was set. **That would have lost the download outright**: `upsert_remote_file` advances the row *before* the decision, so `should_download` is false on every later sweep. My comment asserted the opposite. Caught by tracing it rather than by a reviewer, and it is recorded in the code so nobody re-attempts it.

The real fix is at the download site: content already identical to the incoming remote can never be lost by overwriting, so it does not take the conflict path. Without it, the corrected C1 guard would copy content that matches the remote exactly.

### W4 — the comment no longer promises a resolution that does not occur

### Counts

```
cargo test        203 -> 208, all passing
cargo fmt --check clean
cargo clippy      4 lib warnings, unchanged from main
```

All three run this time. The `fmt` miss on the first push was mine — I ran two of the three, and the check DBSYNC-90 added to CI three tickets ago caught it.

### Mutations — six, each on a tree verified identical to baseline first

| | Mutation | Reddens |
|---|---|---|
| M1 | remove the marking in the `had_remote` branch | `vanished_file_..._marks_it_for_rescan` |
| M2 | marker back to "not a conflict" (**the shipped bug**) | `download_would_conflict_true_when_baseline_is_the_rescan_marker` |
| M3 | remove the `reconcile_remote_absent` guard | `reconcile_remote_absent_does_nothing_while_..._rescan` |
| M4 | NIT back to exactly `!p.exists()` | `is_confirmed_absent_is_false_for_a_path_it_cannot_stat` |
| M5 | remove the pending-job exclusion (C2) | `a_marked_row_does_not_manufacture_a_conflicted_copy_...` |
| M6 | remove the `debug_assert!` | `upsert_local_file_rejects_an_empty_hash_in_debug` |

M4 is the one the review used to prove the NIT was uncovered; it now fails, which is the point.

**M5 first produced no failures at all.** The `sed` had not matched — `cargo fmt` had split the condition across two lines after I wrote the pattern. Re-applied and confirmed red. A mutation that did not apply is a false negative, not evidence, and this repo has shipped a false green from a stale mutation before.

### Refuted in the PR's favour, from the review

The one thing the previous round could not trace — whether dehydrating a marked file discards content Dropbox never received — **cannot happen**. `is_file_fully_synced` never reads the index row's hash; it re-hashes the bytes on disk against `remote.content_hash`, so a marked file the remote lacks fails that check and dehydration returns `Err`. That can now be stated positively instead of hedged.

mbsanchez:
## Round-two findings addressed — `6ee26be`

All three blocking, plus the warnings and nits. **The headline: my fix for review finding C2 re-created the data loss this branch exists to prevent, and a new test asserted it as correct.**

### N1 — the C2 fix is reverted, and C2 itself was wrong

The round-one finding called the conflicted copy "litter". It is not, and the reasoning is worth keeping:

**The only job that can be active on a MARKED path is one the remote side queued.** It cannot be an upload — the upload that sets the marker is completed the moment it returns, and the only other producer is `process_local_file_change`, which clears the marker in the same call. So a pending job on a marked path means *remote content is about to land on top of local bytes Dropbox has never received*. That copy is the protection.

Excluding the marker took the plain arm, which writes the real hash back and **clears the marker before the queued download drains**. `download_would_conflict` then sees `baseline == on_disk`, returns `false`, and the download overwrites the unuploaded edit with nothing logged. The C1 guard was disarmed by the act of skipping the branch it works with.

Verified the mechanism myself before reverting — the plain arm's `upsert_local_file(relative, &hash, ...)` is what clears it.

Reverted. The test is inverted and renamed, and now also asserts the copy holds *the bytes that were at risk* rather than merely counting one file — preserving the wrong content would have satisfied a count and lost the edit anyway.

The exclusion had a second effect I had not seen: it enqueued a competing upload on a path that already had an active job, which is the invariant that arm exists to maintain. `ON CONFLICT` does not collapse them because the job types differ.

### N2 — the composed guard is now tested, and it is the line that mattered most

Both mutations from the review — forcing the guard off, and inverting it — left the suite green. The pure helper had coverage; the decision that actually runs did not. That is the same gap as W1 one round earlier, in the code written to close W1.

Extracted as `download_would_destroy_local` and tested directly: marker + differing remote (protect), marker + identical remote (allow), unknown remote (fall back to baseline). **M7 and M8 now redden it.**

### N3 — W4 was reported fixed and had not been touched

`git diff` showed zero changed lines. I claimed seven fixes and delivered six. The comment now says what actually happens: the scan clears the marker and enqueues an upload that can **resurrect a file the user deleted on Dropbox**, and only a race with the next remote sweep avoids it. Deferring is still better than either alternative, but it trades a guaranteed wrong answer for a likely one — and closing it properly is a change to the upload path that belongs in its own ticket rather than smuggled in here.

### N4 — the guard's justification was false for two of its three call sites

`download_remote_file_internal` also runs from `.cloudsc` hydration and from a manual IPC download, where no sweep advanced any row. The guard is safe for a stronger reason, now the one written down: **if the recorded remote hash equals the on-disk hash then those exact bytes are on Dropbox**, so overwriting them cannot destroy anything — regardless of call site or staleness.

### N5 and the nits

The C2 comment recorded the false premise as settled reasoning; it went with the revert, replaced by why the copy is correct. The `remote_index` comment no longer says "the bytes can actually be compared" (hashes are compared; the bytes are not in hand). The W1 test now observes everything and restores permissions **before** asserting, so a failure cannot leak a `0o000` directory past `TempDir::drop`, and it fails loudly under root instead of passing vacuously.

### Counts

```
cargo test        208 -> 211
cargo fmt --check clean
cargo clippy      4 lib warnings, unchanged from main
```

### Mutations — eight, each ASSERTED TO HAVE APPLIED before running

The previous round published a mutation that silently failed to match after `cargo fmt` reflowed the condition across two lines. Every mutation now compares the file hash before and after and reports `NO APLICÓ` rather than a clean run, because a mutation that did not apply is a false negative, not evidence.

| | Mutation | Reddens |
|---|---|---|
| M1 | remove the marking | `vanished_file_..._marks_it_for_rescan` |
| M2 | marker back to "not a conflict" | `download_would_conflict_true_...` + `download_would_destroy_local_protects_...` |
| M3 | remove the `reconcile_remote_absent` guard | `reconcile_remote_absent_does_nothing_...` |
| M4 | NIT back to `!p.exists()` | `is_confirmed_absent_is_false_for_a_path_it_cannot_stat` |
| M5' | **re-apply the C2 exclusion** (this round's bug) | `a_marked_row_makes_a_conflicted_copy_when_a_job_is_pending` |
| M6 | remove the `debug_assert!` | `upsert_local_file_rejects_an_empty_hash_in_debug` |
| M7 | force the composed guard off | 2 `download_would_destroy_local_*` |
| M8 | invert the composed guard | 3 `download_would_destroy_local_*` |

### What the review confirmed as genuinely fixed

C1, W1, W2, W3 and W5 all hold. Two worth repeating: the `mark_local_file_for_rescan` split was checked for holes — `write_local_file_row` is private with two in-module callers, so the assert cannot be bypassed, and the separator normalisation means a Windows-keyed row cannot be missed. And W3's reasoning was independently confirmed: deferring in `reconcile_remote_present` **would** lose the download, because `upsert_remote_file` advances the row before the decision.

One narrowing worth recording: the marker never survives the full-scan path, which clears it before the remote half enqueues anything. The window is real — `remote_longpoll::apply_and_drain` applies the delta and drains with no local scan — but narrower than the comments imply.

mbsanchez:
## Review: approved — third pass, `6ee26be` (+ `1b9e2b1`)

Three rounds. The first two blocked, and both were right to.

### What round three verified rather than accepted

- **N1 is closed on both halves.** The round-two probe re-run against head: one conflicted copy, holding the at-risk bytes, and **no competing upload on the original path** — the conflicted-copy arm enqueues against the copy, so the two active jobs sit on different paths and neither discards the other. Lossless, confirmed by reproduction rather than asserted.
- **The N2 extraction is faithful case by case** — `is_some_and` vs `map` + `==` on the absent row, the present row, and the composition, all equivalent. No case changed by accident.
- **The identical-bytes argument is airtight.** All four writers of `remote_file_index.content_hash` record content Dropbox served or held; there is no writer that advances the remote row from local content. No counterexample exists.

### `1b9e2b1` takes all four remaining findings

**The one that mattered: a coverage gap, not a comment.** The inverted test checked the copy and its content but never where the recovery upload points. Changing one word — enqueueing against the original path instead of the copy — reintroduced the competing-upload half of the loss **with all 211 tests green**. Now asserted, and M11 reddens it.

**Two comments that outlived what they describe.** `download_would_conflict`'s doc block still said "true only when we have a previously-synced baseline hash", three rounds after the marker arm made that false — every correction had edited the comments *inside* the match and never looked above the signature. And the rewritten `reconcile_remote_absent` comment, which existed to replace a factually wrong one, introduced a new false causal link in the same sentence: the skip-if-identical check does not read the remote row at all, it makes a live metadata request.

**The deferred regression has a ticket.** DBSYNC-93, cited in the comment, matching how `windows_file_assoc.rs` defers to DBSYNC-33/52. It is a regression this branch introduces, not pre-existing, so leaving it with no owner outside a code comment was not good enough.

### Final state

```
cargo test        197 (main) -> 211
cargo fmt --check clean
cargo clippy      4 lib warnings, unchanged
CI                5/5 green
```

**Nine mutations, each asserted to have applied** before running — comparing the file digest and reporting a no-op rather than a clean pass. That guard exists because round two published a mutation that silently failed to match after `cargo fmt` reflowed the code.

### Accepted limits, stated rather than implied

- **The call-site wiring is unverified.** Reverting the call site to bare `download_would_conflict`, or forcing its condition false, both leave the suite green. The *decision* is well covered; that `download_remote_file_internal` uses it is not. Closing that needs the HTTP fetch behind a seam — a real change, named here as a limit rather than left to be discovered.
- **Nobody has reproduced this against a real backup agent.** The tests drive the state machine directly.
- **The window is narrower than the comments imply.** The marker never survives the full-scan path, which clears it before the remote half enqueues anything. The exposure is real via `remote_longpoll::apply_and_drain`, which applies the delta and drains with no local scan.

### The pattern worth recording

Rounds one and two both had their worst defect in code written to fix the previous round — and round two's came from a **review finding that was itself wrong**, which I compounded rather than checked. Round three broke the pattern: the remaining findings were documentation and coverage around logic that held.

The reviewer re-examined its own round-two findings when asked and confirmed they stood, which is the check that made this round trustworthy rather than just quieter.

Moving to QA.

mbsanchez:
## QA Result: PASSED ✅ — with three limits stated rather than folded into the tick

`automated-only` branch. Every command run from a verified-clean tree at `1b9e2b1` (`git status` → 0 modified files).

```
cargo test                                 exit 0    211 passed, 0 failed
cargo clippy                               4 lib warnings, unchanged from main
npm --workspace apps/desktop run build     exit 0
CI on 1b9e2b1                              5/5 green
```

**14 of those 211 tests belong to this ticket.** Named, because "211 passed" says nothing about whether the right things are covered:

```
vanished_file_with_a_remote_copy_keeps_its_row_and_marks_it_for_rescan
vanished_file_with_no_remote_copy_is_still_forgotten
a_row_marked_for_rescan_is_re_detected_and_re_uploaded
a_marked_row_makes_a_conflicted_copy_when_a_job_is_pending
download_would_conflict_true_when_baseline_is_the_rescan_marker
download_would_destroy_local_protects_a_marked_row_from_different_remote_content
download_would_destroy_local_allows_an_overwrite_with_identical_bytes
download_would_destroy_local_falls_back_to_the_baseline_when_the_remote_is_unknown
reconcile_remote_absent_does_nothing_while_the_row_is_marked_for_rescan
mark_local_file_for_rescan_sets_the_marker_and_keeps_size_and_mtime
mark_local_file_for_rescan_is_a_noop_when_the_row_is_absent
upsert_local_file_rejects_an_empty_hash_in_debug - should panic
is_confirmed_absent_distinguishes_missing_from_present
is_confirmed_absent_is_false_for_a_path_it_cannot_stat
```

### Acceptance criteria → the evidence that actually covers each

| AC | Covered by |
|---|---|
| The `had_remote` branch marks the row instead of leaving it; the row stays; still returns `Ok(())` | `vanished_file_with_a_remote_copy_...` + mutation **M1** |
| The marker is a single named constant, not a bare `""` | `Db::HASH_NEEDS_RESCAN`, and `mark_local_file_for_rescan` is now the only writer — enforced by a `debug_assert!` with mutation **M6** |
| `download_would_conflict` handles the marker | `download_would_conflict_true_...` + **M2**. **Note the AC as written asked for the opposite** — see below |
| `reconcile_remote_absent` neither deletes nor raises a spurious conflict; the choice is recorded | `reconcile_remote_absent_does_nothing_...` + **M3**; decision written in the code and deferred part filed as DBSYNC-93 |
| Full scenario test; a mutation removing the marking turns it red | `a_row_marked_for_rescan_is_re_detected_and_re_uploaded` + **M1** |
| Negative control: a genuine deletion still propagates | `vanished_file_with_no_remote_copy_is_still_forgotten`, plus the pre-existing `reconcile_remote_absent_deletes_when_local_matches_last_synced` staying green |
| Negative control: no job reaches `failed` | The branch returns `Ok(())`; `vanished_file_...` asserts `expect("must not error")` |
| NIT: `p.exists()` → confirmed `NotFound` | `is_confirmed_absent_is_false_for_a_path_it_cannot_stat` + **M4**. Two sites, not the one the ticket named |
| `cargo test` passes, count stated; clippy no new warnings | 211 stated above; clippy unchanged at 4 |

### One acceptance criterion was written wrong, and following it would have shipped a data loss

The ticket's AC said `download_would_conflict` should treat the marker "as an absent baseline — reusing the existing `None` path — so no spurious conflicted copy is created." **That is the opposite of correct**, and implementing it exactly as written is what round one shipped: `None` means no row was ever recorded, while the marker means a row that *was* advanced for a real edit whose upload was cancelled. Returning `false` let an incoming download overwrite those bytes silently.

The AC is met by doing the opposite of what it says. Flagging it here rather than silently ticking the box, and the ticket description should be corrected before it is closed.

### Not covered by any check, and worth knowing before this merges

1. **The call-site wiring is unverified.** Reverting the download call site to the bare `download_would_conflict`, or forcing its condition false, both leave the suite green at 211. The *decision* is well covered — that is `download_would_destroy_local` and its three tests — but nothing proves `download_remote_file_internal` actually calls it. Closing that needs the HTTP fetch behind a test seam, which is a real change and not one to smuggle into this PR.
2. **Nobody has reproduced this against a real backup agent.** The tests drive the state machine directly, so they establish the logic, not that the unlink window is as wide as anyone believes.
3. **A regression this branch introduces is deferred, not fixed** — DBSYNC-93, a marked file can be resurrected on Dropbox after the user deletes it there. It has an owner and is cited in the code.

### Verdict

**PASS** on everything a check can reach, with the three limits above stated rather than absorbed into a green tick.

PR #139, opened 2026-08-30
fix(#DBSYNC-40): make schema migration and sync-state reset atomic

mbsanchez:
## Summary

Two multi-statement sequences in `db.rs` ran without a transaction. Both now run in one.

**`reset_sync_state` is the one that can hurt.** Six `DELETE`s, no transaction, and it runs whenever the user changes their sync folder — ordinary use, not an edge case. A crash, lock error or disk failure between any two left a state neither half of the sync engine expects: clear `local_file_index` but not `remote_file_index`, and the next scan walks a folder full of files with no index rows while the remote index still claims to know them. On a client that carries a mass-delete circuit breaker *because* bulk operations here destroy data, a half-cleared index is not a tidiness problem.

**`migrate()` had the same shape.** A failure partway left a half-migrated schema and the next startup re-ran from the top, trusting every step to be idempotent. That trust is currently justified — nothing enforces it, and the failure was silent.

## The ticket's premise was wrong, and the ticket is renamed

It said the schema *"has no safe migration path"*. There is one: `migrate()` is 193 lines and has already carried eight schema changes — eight `add_column_if_missing` calls, a separator canonicalisation, a guarded rebuild of `sync_jobs` to add CHECK constraints, and index creation. What was missing was atomicity, not the path. Retitled from *"Add database schema migration system"* to match.

## Checked before writing any of it

- **SQLite DDL is transactional**, so the `CREATE TABLE`s, `ALTER TABLE ADD COLUMN`s and the `sync_jobs` rebuild all roll back cleanly.
- **The schema has zero foreign keys** (`grep -cE "REFERENCES|FOREIGN KEY"` → 0), so the rebuild's DROP/RENAME cannot trip referential actions inside a transaction.
- **No write-PRAGMA runs inside `migrate()`.** `journal_mode = WAL` and `foreign_keys = ON` stay in `new_at`, *before* the call — `journal_mode` cannot be set inside a transaction, and moving it in would be a silent regression no test here would catch. The only PRAGMA reached from inside is `table_info`, a read.

## No version counter, by decision

The original AC asked for `schema_version` in `app_config` and numbered migrations. Declined by the maintainer, and the reasoning is recorded **at `migrate()`** rather than left as an apparent gap:

The current approach is **declarative and self-converging** — each step states a desired end state, so a database reaches it from wherever it happens to be. A counter describes a *sequence*, and a counter that is wrong (hand-edited file, restored backup, half-applied migration) silently skips the very steps that would have repaired it.

The proposed storage was also circular: `app_config` is a table `migrate()` itself creates, so reading a version out of it before migrating cannot work on a fresh database. **`PRAGMA user_version` is the mechanism if this is ever revisited** — it lives in the file header and participates in the transaction.

What is given up, stated so the trade is visible: startup keeps re-inspecting the schema every run, and a failure stays a silent retry rather than a detectable stop.

## Stack

`desktop-tauri` — one Rust file. No IPC surface, no frontend, no schema change.

## Jira Ticket

DBSYNC-40 — https://manuelbsan.atlassian.net/browse/DBSYNC-40

## Test Plan

- [x] `cargo test` — **211 → 214**. 211 measured on `main` before any edit.
- [x] `cargo fmt --check` — clean.
- [x] `cargo clippy` — 4 lib warnings, unchanged from `main`.

### The rollback tests force a real failure

A happy-path test would pass with the transactions removed, which is the definition of a check that cannot fail. Both levers were verified against real SQLite first:

1. **`reset_sync_state_rolls_back_when_a_deletion_fails_partway`** — seeds every table, then drops `known_folders`, the **fifth** of six deletions, so the first four certainly ran. Asserts the call errors *and* that `local_file_index` still holds its rows.
2. **`migrate_rolls_back_when_a_step_fails_partway`** — plants a *table* named `idx_sync_jobs_status_retry`, colliding with an index `migrate` creates near the end (SQLite reports "there is already a table named …" even for `CREATE INDEX IF NOT EXISTS`). Asserts the call errors *and* that `app_config`, created by the first statement, does not exist afterwards.
3. **`migrate_is_idempotent_and_leaves_an_identical_schema`** — runs it twice and compares the whole of `sqlite_master`, with a sanity assertion so it cannot pass by comparing two empty lists.

### Mutations

**Four say something about the production code**, and they are the ones worth keeping:

| | Mutation | Reddens |
|---|---|---|
| M1 | remove the transaction from `reset_sync_state` | 1 — `reset_sync_state_rolls_back_...` |
| M2 | remove the transaction from `migrate` (signature + call site) | 1 — `migrate_rolls_back_...` |
| M3 | never commit in `reset_sync_state` | 2, both pre-existing |
| M4 | never commit in `migrate` | **87** across 7 modules, all pre-existing |

M1/M2 prove the rollback comes from the transaction; M3/M4 prove the commit is reached. Neither pair covers the other. M4 is a blunt signal — with no commit the whole schema rolls back, so everything touching a `Db` fails — so M1/M2 are the discriminating half.

Six further mutations confirm each new test fails when the behaviour it names is broken: reordering the `CREATE INDEX` block before the `sync_jobs` rebuild; forcing `migrate` to fail at its first statement instead of the planted collision; dropping the `job_type` CHECK from the rebuild copy; drifting a column into the rebuild copy only; removing an `add_column_if_missing` call (**46** reds across four modules); and reordering the `known_folders` deletion to the front of `reset_sync_state`. Each is documented in the test it validates, next to the assertion it justifies, which is where a future editor will actually read it.

### Corrections to earlier versions of this table

Three counting errors, all mine, all published here before being caught:

- **M4 as "5 `cloudsc_ops` tests"** — it is 87 across seven modules.
- **M7 as "3 `remote_index` tests"** — it is 46 across four. That row was also *mislabelled*: it described a different mutation from the one run.
- **The header said "nine"** above ten rows.

The first two shared a cause: the mutation runner piped results through `head`, so it published the first N lines as the total. It now counts, and its apply-guard asserts the pattern matches **exactly once** rather than merely that the file changed — a mutation landing in two places is as untrustworthy as one landing in none.

### Five defects, none of them in the change

Every one was an assertion weaker than the comment above it, and the tell each time was a comment saying **"must"** or **"certainly"** about something nothing constrained:

1. `migrate_rolls_back_...` asserted `app_config` absent without pinning *when* the failure occurred — green with the failure forced into the first statement.
2. The index assertion counted every `idx_%` in the schema — green with the exact reordering its comment named.
3. The CHECK assertion shared its oracle with the production heuristic it was testing.
4. The behavioural CHECK anchor asserted `is_err()` — green on an unrelated `NOT NULL` failure with the constraint genuinely gone.
5. The reset lever could not distinguish "rolled back" from "never executed" — green with the deletion reordered, **and M1 stopped reddening too**, disarming the test and its validating mutation together.

The production change has been correct since the first commit: two `transaction()` calls, and the only production line touched since is a signature. All five defects were in assertions written *about* the change, two of them in code answering a review round.

The process conclusion, since one per round is not a coincidence: the pinning discipline applied to the first weak assertion should have been applied to all its siblings in the same pass.

## A correction worth recording

**Test 2 failed on its first run.** The plan's probe used `idx_sync_jobs_status` — a name I invented while checking that the collision *mechanism* works, without reading which indexes `migrate` actually creates. There are two, neither had that name, so the migration completed happily and the test caught me. The real name is `idx_sync_jobs_status_retry`.

The lever was right; the target was invented. Worth stating because the probe looked like verification and was only half of it.

## What this does not do

It makes failures **atomic**, not **detectable**. A failed `migrate()` still aborts startup with whatever error SQLite produced, and the next launch retries from the top. Unchanged, and out of scope given the decision above — but it should not be implied otherwise.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/storage/db.rs
+++ b/apps/desktop/src-tauri/src/storage/db.rs
@@ -1553,13 +1553,19 @@ mod tests {
     use std::path::PathBuf;
     use std::time::{SystemTime, UNIX_EPOCH};
 
-    /// The whole schema as comparable rows — every object, its name and its DDL. Used to
-    /// assert that a second `migrate` changes nothing, which is what "idempotent" means and
-    /// what a bare `expect` on the second call does not check.
+    /// The whole schema as comparable rows — every object, its name, its **rootpage** and
+    /// its DDL. Used to assert that a second `migrate` changes nothing, which is what
+    /// "idempotent" means and what a bare `expect` on the second call does not check.
+    ///
+    /// `rootpage` is in there deliberately. Without it, making the `sync_jobs` rebuild
+    /// unconditional — so every startup drops and recreates the table — left both
+    /// idempotency tests green, because the recreated table has identical DDL. A recreated
+    /// table gets a new rootpage, so including it turns "the schema looks the same" into
+    /// "the schema IS the same objects".
     fn schema_rows(c: &Connection) -> Vec<String> {
         let mut stmt = c
             .prepare(
-                "SELECT type || ' ' || name || ' ' || COALESCE(sql, '') \
+                "SELECT type || ' ' || name || ' ' || rootpage || ' ' || COALESCE(sql, '') \
                  FROM sqlite_master ORDER BY type, name",
             )
             .expect("prepare");
@@ -1946,12 +1952,70 @@ mod tests {
         );
     }
 
+    /// DBSYNC-40. `reset_sync_state` must clear **every** table it names, and this test
+    /// exists because a systematic sweep found that three of its six deletions could be
+    /// deleted outright with the whole suite still green — including `remote_file_index`,
+    /// the one this function's own doc comment names as the disaster case:
+    ///
+    /// > *"clear `local_file_index` but not `remote_file_index`, and the next scan walks a
+    /// > folder full of files with no index rows while the remote index still claims to
+    /// > know them"*
+    ///
+    /// The rollback test next door reads like it covers this — it asserts the remote row
+    /// survives a failed reset — but that assertion passes just as happily when the remote
+    /// deletion never runs at all. "Rolled back" and "never executed" look identical from
+    /// the outside, which is the same confusion the trigger lever exists to resolve.
+    #[test]
+    fn reset_sync_state_clears_every_table_it_names() {
+        let db = Db::new_at(&unique_db_path()).expect("db init");
+        db.upsert_local_file("a.txt", "H", 1, 0)
+            .expect("seed local");
+        db.upsert_remote_file("a.txt", "H", "rev", 0)
+            .expect("seed remote");
+        db.enqueue_job("upload", Some("a.txt"), Some("a.txt"))
+            .expect("seed job");
+        db.add_conflict("a.txt", "a.txt", "seeded", None, false)
+            .expect("seed conflict");
+        db.upsert_known_folder("sub").expect("seed folder");
+        db.set_app_config(crate::remote_index::REMOTE_DELTA_CURSOR_KEY, "cursor")
+            .expect("seed cursor");
+
+        db.reset_sync_state().expect("reset");
+
+        assert!(db.get_local_file("a.txt").expect("q").is_none(), "local");
+        assert!(db.get_remote_file("a.txt").expect("q").is_none(), "remote");
+        assert_eq!(db.count_active_jobs().expect("q"), 0, "jobs");
+        assert!(
+            db.list_recent_conflicts(10).expect("q").is_empty(),
+            "conflicts"
+        );
+        assert!(
+            db.list_known_folders().expect("q").is_empty(),
+            "known folders"
+        );
+        assert!(
+            db.get_app_config(crate::remote_index::REMOTE_DELTA_CURSOR_KEY)
+                .expect("q")
+                .is_none(),
+            "delta cursor"
+        );
+    }
+
     /// DBSYNC-40. Same requirement for the schema: a failure partway leaves nothing behind.
     ///
     /// The lever is a **table** named `idx_sync_jobs_status_retry`, colliding with an index
     /// `migrate` creates near the end — SQLite reports "there is already a table named …"
     /// even for `CREATE INDEX IF NOT EXISTS`. `app_config` is created by the very first
     /// statement, so its absence afterwards is what proves the whole batch unwound.
+    ///
+    /// **That last step rests on an ordering this test cannot check**, and saying so is the
+    /// point: it assumes `CREATE TABLE app_config` runs before the failure. Move it after
+    /// the index block and the assertion passes vacuously. The reset test solves the same
+    /// problem with a trigger that fires only after an earlier statement ran, but SQLite has
+    /// no DDL trigger hook, and the failing statement here is a `CREATE INDEX` — so there is
+    /// no way to make this failure conditional on `app_config` already existing. The
+    /// ordering is verified by reading `migrate`, not by this test. Accepted knowingly:
+    /// it would take reordering a `CREATE TABLE` behind an index creation to break it.
     #[test]
     fn migrate_rolls_back_when_a_step_fails_partway() {
         let path = unique_db_path();
--- a/apps/desktop/src-tauri/src/storage/db.rs
+++ b/apps/desktop/src-tauri/src/storage/db.rs
@@ -1899,16 +1899,42 @@ mod tests {
             .expect("seed remote");
         db.upsert_known_folder("sub").expect("seed folder");
 
-        // Break the fifth deletion. Everything before it will have run.
+        // The lever fires ONLY IF an earlier deletion has already run inside the
+        // transaction, and that is the whole point of using a trigger rather than the
+        // obvious `DROP TABLE known_folders`.
+        //
+        // A dropped table cannot tell "deleted then rolled back" from "never executed".
+        // Reorder the `known_folders` deletion to the front of `reset_sync_state` — a
+        // plausible edit, nothing about the function forbids it — and the surviving-row
+        // assertions below become trivially true, so the test passes with no rollback
+        // exercised at all. Worse, measured: with that reordering the mutation that removes
+        // the transaction ALSO stops reddening. The test the ticket calls its whole
+        // evidence, and the mutation that validates it, are disarmed together by moving one
+        // line inside the function under test.
+        //
+        // This trigger aborts only once `local_file_index` is empty, so the abort IS the
+        // proof that deletion 1 ran. The assertions then prove it was undone.
         {
             let conn = db.write.lock().expect("lock");
-            conn.execute_batch("DROP TABLE known_folders")
-                .expect("drop");
+            conn.execute_batch(
+                "CREATE TRIGGER abort_once_local_is_cleared \
+                 BEFORE DELETE ON known_folders BEGIN \
+                   SELECT RAISE(ABORT, 'local_file_index was already cleared') \
+                   WHERE (SELECT COUNT(*) FROM local_file_index) = 0; \
+                 END",
+            )
+            .expect("install the ordering-sensitive lever");
         }
 
-        let result = db.reset_sync_state();
-
-        assert!(result.is_err(), "the reset must fail, not silently skip");
+        let err = db
+            .reset_sync_state()
+            .expect_err("the reset must fail, not silently skip");
+        assert!(
+            err.to_string()
+                .contains("local_file_index was already cleared"),
+            "the failure must come from the ordering-sensitive lever, which fires only \
+             after an earlier deletion ran: {err}"
+        );
         assert!(
             db.get_local_file("a.txt").expect("query").is_some(),
             "the FIRST deletion must have been rolled back — a half-cleared index is what \
--- a/apps/desktop/src-tauri/src/storage/db.rs
+++ b/apps/desktop/src-tauri/src/storage/db.rs
@@ -83,7 +83,7 @@ impl Db {
     /// from the production database (which `db_path()` resolves via OS-specific
     /// app-data dirs), so running `cargo test` never touches a user's real DB.
     pub fn new_at(path: &std::path::Path) -> AppResult<Self> {
-        let write = Connection::open(path)?;
+        let mut write = Connection::open(path)?;
         write.execute_batch(
             "
                 PRAGMA foreign_keys = ON;
@@ -91,7 +91,7 @@ impl Db {
                 PRAGMA synchronous = NORMAL;
                 ",
         )?;
-        migrate(&write)?;
+        migrate(&mut write)?;
 
         let read = Connection::open_with_flags(
             path,
@@ -141,22 +141,38 @@ impl Db {
     }
 
     /// Clears local sync state to avoid stale jobs when the sync folder changes.
+    /// Clears every trace of the previous sync folder, **atomically** (DBSYNC-40).
+    ///
+    /// The transaction is the point. These six deletions used to run as six independent
+    /// statements, and a crash, lock error or disk failure between any two left a state
+    /// neither half of the sync engine expects: clear `local_file_index` but not
+    /// `remote_file_index`, and the next scan walks a folder full of files with no index
+    /// rows while the remote index still claims to know them. This is not a hypothetical
+    /// path — it runs whenever the user changes their sync folder — and on a client that
+    /// carries a mass-delete circuit breaker because bulk operations here destroy data, a
+    /// half-cleared index is not a tidiness problem.
+    ///
+    /// `rusqlite::Transaction` rather than literal `BEGIN`/`COMMIT`: it rolls back when
+    /// dropped, so an early `?` return between the statements cannot leave a transaction
+    /// open — which would be this function's own failure mode, one level up.
     pub fn reset_sync_state(&self) -> AppResult<()> {
-        let conn = self
+        let mut conn = self
             .write
             .lock()
             .map_err(|_| AppError::Storage("db write lock poisoned".into()))?;
-        conn.execute("DELETE FROM local_file_index", [])?;
-        conn.execute("DELETE FROM remote_file_index", [])?;
-        conn.execute("DELETE FROM sync_jobs", [])?;
-        conn.execute("DELETE FROM sync_conflicts", [])?;
-        conn.execute("DELETE FROM known_folders", [])?;
+        let tx = conn.transaction()?;
+        tx.execute("DELETE FROM local_file_index", [])?;
+        tx.execute("DELETE FROM remote_file_index", [])?;
+        tx.execute("DELETE FROM sync_jobs", [])?;
+        tx.execute("DELETE FROM sync_conflicts", [])?;
+        tx.execute("DELETE FROM known_folders", [])?;
         // Drop the cursor-delta cursor so remote change detection re-seeds
         // against the new folder (DBSYNC-30); other app_config keys are kept.
-        conn.execute(
+        tx.execute(
             "DELETE FROM app_config WHERE key = 'remote_delta_cursor'",
             [],
         )?;
+        tx.commit()?;
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


mbsanchez:
## Review findings addressed — `58f81a8`

The review found the change itself correct and everything around it thinner than I claimed. All four warnings and all five suggestions are in.

### The one that mattered: my rollback test could pass vacuously

`migrate_rolls_back_when_a_step_fails_partway` reasons *"`app_config` is created by the **first** statement, so its absence proves a **later** failure unwound it"* — and asserted only the absence, never the "later". Reproduced the reviewer's proof: forcing the failure into the first statement instead (a stray `VIEW` named `sync_jobs` breaks the opening `execute_batch`) leaves both assertions satisfied **with no atomicity involved at all**. Green, proving nothing.

The error is now pinned to the planted collision, so a future edit that makes an earlier step fail turns it red rather than quietly hollowing it out. **M5** is exactly that mutation.

That is the same defect my own PR body identifies in happy-path tests, one level in.

### The `sync_jobs` rebuild had no test, and it is the branch that matters most

It only runs on **existing installations** — a fresh database gets the CHECKs from `CREATE TABLE`, so `sync_jobs_has_check` is true and the DROP/RENAME never executes. Every other migration test starts from an empty file, so none of them touched it. The branch most likely to break someone's database was the uncovered one, and it is also the branch now running inside a transaction for the first time.

Covered: a legacy CHECK-less table carrying an out-of-set row → CHECKs arrive, the bad row is dropped rather than aborting the copy, the indexes land on the *rebuilt* table, and a second `migrate` is still a no-op. **M6** disables the rebuild and reddens it.

### The other two

An unused `use rusqlite::Connection;` in `data_dir_tests` — my `sed` matched **both** occurrences of `use super::Db;` and added it to a module that never mentions `Connection`. Invisible under `cargo clippy`, a warning under `--all-targets`, which is the scope that would have caught it. Removed; the only remaining unused import (`std::io`) is pre-existing on `main`.

And a doc comment promising more than the code delivers: `reset_sync_state` is atomic, but the **folder switch is not**. `commands.rs` sets the new folder in its own transaction and *then* calls this one, so a failed reset leaves `app_config` pointing at the new folder while the index describes the old one. Pre-existing, and this PR makes it less bad — coherent-but-stale beats torn — but the comment must not read as claiming otherwise. Filed as **DBSYNC-94** and cited in the code, the way `windows_file_assoc.rs` cites its deferrals.

### Suggestions taken

- **"Circular" overstated it.** Bootstrapping a version out of `app_config` needs its own step, solvable in a line. The real arguments for `user_version` are that it lives in the file header and is transactional — which is what the comment now leads with.
- **`migrate_in_tx` takes `&Transaction`**, so the invariant its name describes is a compile error rather than a convention.
- The duplicated summary line above `reset_sync_state` is gone.
- **`BEGIN DEFERRED` declined, explicitly.** Both call sites write first, so the lock is taken immediately and there is no read→write upgrade to lose. Reachable only in a scenario that does not exist today (single writer, read connection not yet open). Declining knowingly rather than by omission.

---

## CORRECTION to the previous evidence table

**M4 reddens 87 tests across seven modules, not "5 `cloudsc_ops` tests".**

I read the first five lines of a truncated run and published them as the total. The direction is favourable — the evidence is far stronger than I claimed — but the number was wrong, in a table whose entire purpose is exactness, and in a PR that criticises imprecision elsewhere. Re-derived with a counting command rather than a `head`.

Worth adding what the reviewer pointed out: **M4 is a blunt signal.** With no commit the whole schema rolls back, so every test touching a `Db` fails. It proves the commit is reached; it says nothing about it being in the right place. M1/M2 remain the discriminating half.

## Mutations — six, each asserted to have applied

| | Mutation | Reddens |
|---|---|---|
| M1 | remove the transaction from `reset_sync_state` | 1 — `reset_sync_state_rolls_back_...` |
| M2 | swallow the error inside `migrate`'s transaction | 1 — `migrate_rolls_back_...` |
| M3 | never commit in `reset_sync_state` | 2 pre-existing |
| M4 | never commit in `migrate` | **87** across 7 modules, all pre-existing |
| M5 | make `migrate` fail at the FIRST statement instead of the planted collision | 1 — `migrate_rolls_back_...` |
| M6 | skip the `sync_jobs` rebuild | 1 — `migrate_rebuilds_a_legacy_sync_jobs_table_...` |

## Counts

```
cargo test        214 -> 215
cargo fmt --check clean
cargo clippy      4 lib warnings (unchanged); --all-targets no longer adds one
```

## What the review verified that I had only asserted

Worth recording because these were the load-bearing premises: `PRAGMA foreign_keys` set **before** `BEGIN` **is** honoured for DDL inside — probed with an FK-violating insert. `PRAGMA journal_mode` inside a transaction returns `Ok` and silently does nothing, so "a silent regression no test would catch" was literally accurate. And the zero-FK claim holds and is stronger than I knew: there are also **no triggers and no views**, which matters more, because `legacy_alter_table` defaults to `0` and `ALTER TABLE ... RENAME` in that mode reparses the whole schema and errors on any unresolvable view or trigger.

mbsanchez:
## Systematic assertion sweep — `55e402e`

Five rounds of "find the next defect" turned up exactly one per round. That cadence was the search being shallow, not the defects being rare. Changing the method to *enumerate every assertion and ask what production edit defeats it* — sixteen assertions, eleven targeted mutations — found **three at once**.

### The serious one is behaviour coverage, not test hygiene

**`reset_sync_state` could stop clearing three of its six tables with the suite fully green.** Verified independently, three separate single-line deletions, 215 passed each time:

```
remove  DELETE FROM remote_file_index   ->  215 passed
remove  DELETE FROM sync_jobs           ->  215 passed
remove  DELETE FROM sync_conflicts      ->  215 passed
remove  DELETE FROM local_file_index    ->  1 failed   (the only one constrained)
```

`remote_file_index` is the one that stings. **This PR's own doc comment names it as the disaster case:**

> *"clear `local_file_index` but not `remote_file_index`, and the next scan walks a folder full of files with no index rows while the remote index still claims to know them"*

That line could have been deleted and shipped green through five review rounds arguing it matters.

The rollback test *looked* like it covered this — it asserts the remote row survives a failed reset. But that assertion passes just as happily when the remote deletion never runs at all. "Rolled back" and "never executed" are indistinguishable from outside: the same confusion the trigger lever was added to resolve, one assertion over.

Now covered by a test that seeds all six and asserts all six are cleared. All three mutations redden.

### The rebuild guard was untested

Making `if !sync_jobs_has_check` unconditional — so every startup drops and recreates `sync_jobs` — left the suite green, because the recreated table has identical DDL. `schema_rows` now includes `rootpage`, which changes when a table is recreated, so both idempotency tests catch it. One line, and it reddens two tests.

### One residual accepted, with the reason written down

`migrate_rolls_back_...` depends on `app_config` being created before the failure point and cannot check that. **Accepted rather than fixed**, because the reset test's solution does not transfer: that lever works because SQLite fires `BEFORE DELETE` triggers on DML, and `migrate`'s failing statement is a `CREATE INDEX` — there is no DDL trigger hook to hang a conditional abort on. Breaking it would take reordering a `CREATE TABLE` behind an index creation, which is not an edit anyone makes.

It is now a *stated* dependency in the comment rather than an invisible one, which is the same move that fixed the other four.

### The two questions I asked about accidental load-bearing

- **`assert_eq!(surviving, 1)`** — now load-bearing by design. It has a unique mutation (tightening the rebuild's `WHERE` so every row is dropped) and is *not* what catches the filter being removed, which aborts the copy and is caught elsewhere. Two assertions, two genuinely different failures.
- **The sanity guard** — decorative, and honestly labelled `"sanity:"`. No production mutation reaches it; it only fires if `schema_rows` itself breaks. That is insurance against helper rot, not coverage pretending to be coverage. Kept.

Three assertions catch nothing unique, for the record: the rollback test's `"and so must the second"` (its claim was the unverified one — the new test covers the behaviour instead), the sanity guard, and `sql.contains("CHECK")` (redundant with the pinned `job_type` assertion, and its own comment already says so).

### Counts

```
cargo test        215 -> 216
cargo fmt --check clean
cargo clippy      unchanged from main (--all-targets included)
```

### What five rounds actually established

Six defects, **none in the change**. The production diff is two `transaction()` calls plus one signature, unmoved since the first commit, and across roughly forty mutations it has never been the thing that broke. Every defect was in an assertion written *about* the change — two of them in code written to answer a review round.

The transaction boundaries were verified against SQLite 3.45.0 rather than reasoned about: `foreign_keys` set before `BEGIN` is still enforced inside; `journal_mode` inside a transaction is a silent no-op, so keeping both PRAGMAs in `new_at` is correct; the read connection opening after `migrate` is safe; the `sync_jobs` rebuild behaves identically inside an explicit transaction, with zero FKs, triggers or views to complicate DROP/RENAME.

mbsanchez:
## Review: approved — `cb519c6`

Six rounds. The last one changed method — from "find the next defect" to "enumerate every assertion and ask what production edit defeats it" — and that single change found three at once where the previous five had found one each. The cadence was the search being shallow, not the defects being rare.

### The two blocking items from the final pass, both verified before acting

**A flaky test, and five green CI runs were not evidence against it.** `unique_db_path()` builds its directory from `SystemTime::now().as_nanos()`, which has **microsecond** resolution on macOS — 192393 of 200000 consecutive readings are duplicates. Two tests entering in the same microsecond share a database file.

```
40 parallel runs, before : 4 failures
25 runs, --test-threads=1: 0 failures    (parallel-only)
40 parallel runs, after  : 0 failures
```

The helper predates this branch. `migrate_rebuilds_a_legacy_sync_jobs_table` is simply the first caller intolerant enough to expose it — it plants a bare `CREATE TABLE` where every other caller is `IF NOT EXISTS`. Left alone it would have surfaced as a mystery failure on an unrelated PR.

**The coverage test read through the application's own filtered accessors.** `count_active_jobs` filters `status IN (queued, retry_wait, running)`; `list_recent_conflicts` filters `resolved = 0`. So each deletion could be narrowed to exactly what its reader sees and the test still passed:

```
DELETE FROM sync_jobs WHERE status IN ('queued','retry_wait','running')  -> green
DELETE FROM sync_conflicts WHERE resolved = 0                            -> green
DELETE FROM local_file_index WHERE hash != ''                            -> green
```

Every one is an edit someone makes *on purpose* — "keep the history" — and the assertion messages would still have read `"jobs"`, `"conflicts"`, `"local"`. The third was unanticipated: DBSYNC-56 stores its rescan marker as `hash = ''`, so empty-hash rows genuinely exist.

Now seeded via raw SQL including rows the readers cannot see, asserted with raw `COUNT(*)` per table. All three narrowed mutations redden; the whole-line deletions still do.

**The rule this earns, now in the comment:** a coverage test must not read through the application's own filtered accessors, or the behaviour can be narrowed to the filter while the test stays green.

### What the whole review established

**Seven defects, none in the change.** The production diff has been two `transaction()` calls plus one signature since the first commit, and across roughly fifty mutations it was never the thing that broke. Every defect was in an assertion written *about* the change — two of them in code written to answer a review round.

The recurring shape: **an assertion weaker than the comment above it**, with the tell being a comment saying "must" or "certainly" about something nothing constrained. Five instances. The one that mattered most was the reset lever, where reordering a single line inside the function under test disarmed both the test and the mutation validating it.

### Transaction boundaries — verified against SQLite 3.45.0, not reasoned about

- `PRAGMA foreign_keys` set *before* `BEGIN` **is** still enforced for DDL inside; set *inside*, it is silently ignored.
- `PRAGMA journal_mode` inside a transaction returns `Ok` and does nothing — so "a silent regression no test would catch" was literally accurate, and keeping both PRAGMAs in `new_at` is correct.
- The read connection opening *after* `migrate` commits cannot observe a partial migration.
- The `sync_jobs` rebuild behaves identically inside an explicit transaction: zero foreign keys, **zero triggers and zero views** in the schema, which matters more — `legacy_alter_table` defaults to `0`, and `ALTER TABLE ... RENAME` in that mode reparses the whole schema and errors on any unresolvable view or trigger.
- `PRAGMA user_version` participates in a transaction, as the doc comment claims.

### Final state

```
cargo test        211 -> 216
cargo fmt --check clean
cargo clippy      unchanged from main, --all-targets included
flake             0/40 parallel runs (was 4/40)
CI                5/5 green
```

### Accepted, not closed

`migrate_rolls_back_...` depends on `app_config` being created before the failure point and cannot check it. The reset test's trigger lever does not transfer — SQLite triggers are DML-only, and the failing statement here is a `CREATE INDEX`. A `sqlite3_set_authorizer` hook could do it, at far more machinery than this residual is worth. Now a *stated* dependency rather than an invisible one.

**DBSYNC-94** carries the composition gap: the folder switch is still two transactions, so a failed reset leaves `app_config` on the new folder with the index describing the old one. Pre-existing, made less bad by this PR, and the doc comment says so rather than implying atomicity the caller does not have.

Moving to QA.

mbsanchez:
## QA Result: PASSED ✅

`automated-only` branch. Every command run from a clean tree at `cb519c6` (0 tracked files modified).

```
cargo test                                 exit 0    216 passed, 0 failed
cargo clippy --all-targets                 identical to main, warning for warning
npm --workspace apps/desktop run build     exit 0
CI on cb519c6                              5/5 green
```

**Clippy checked against `main` rather than against a remembered number.** The `--all-targets` scope reports "lib test generated 3 warnings", which looked like a regression until compared: `main` produces the same five warnings, the same counts. Nothing introduced.

### The flake, re-verified here rather than trusted from the implement step

This ticket's last defect was an intermittent failure that five green CI runs had hidden, so QA is the wrong place to take that on faith. Re-run under the **full parallel suite**, which is what CI actually executes rather than the two-test subset the fix was measured against:

```
20 full-suite runs, parallel :  0 failures
```

For contrast, the same two tests before the fix: 4 failures in 40 runs, and 0 in 25 single-threaded — the signature of a shared-path collision rather than a logic bug.

### Acceptance criteria → the evidence covering each

| AC | Covered by |
|---|---|
| `reset_sync_state` performs all six deletions in one transaction, committing once | `reset_sync_state_rolls_back_when_a_deletion_fails_partway` (rollback) + `reset_sync_state_clears_every_table_it_names` (commit reaches all six) |
| `migrate` runs in one transaction, committing once; PRAGMAs stay outside | `migrate_rolls_back_when_a_step_fails_partway`; the PRAGMA placement verified against SQLite 3.45.0 — `journal_mode` inside a transaction is a silent no-op, `foreign_keys` set before `BEGIN` is still enforced inside |
| A failure partway leaves the database unchanged — **demonstrated**, not asserted | Both rollback tests force a real failure. The reset lever is a trigger that fires only *after* an earlier deletion has run, so the abort itself proves ordering — a plain `DROP TABLE` could not distinguish "rolled back" from "never executed" |
| `migrate` still idempotent; a second run is a no-op | `migrate_is_idempotent_and_leaves_an_identical_schema` and the legacy test, both comparing all of `sqlite_master` including `rootpage` |
| The no-version-counter decision recorded in the code, with the `user_version` note | `migrate`'s doc comment |
| `cargo test` passes with the count stated; clippy no new warnings | 216 above; clippy identical to `main` |

### Beyond the ACs — coverage this ticket added that it did not promise

`migrate_rebuilds_a_legacy_sync_jobs_table_inside_the_transaction` covers the `sync_jobs` rebuild, which **only existing installations execute** — a fresh database gets the CHECK constraints from `CREATE TABLE`, so every pre-existing migration test skipped it. It is now the sole guard against the most likely future migration mistake: a column added to the fresh schema but not to `add_column_if_missing`, which reddens 46 tests across four modules.

That branch had zero coverage before this PR and is also the branch now running inside a transaction for the first time.

### Stated rather than folded into the tick

1. **One residual accepted, not closed.** `migrate_rolls_back_...` depends on `app_config` being created before the failure point and cannot check it — SQLite triggers are DML-only, and the failing statement is a `CREATE INDEX`, so the reset test's lever does not transfer. A `sqlite3_set_authorizer` hook could, at far more machinery than the residual is worth. Now a *stated* dependency in the comment.
2. **DBSYNC-94 carries the composition gap.** The folder switch is still two transactions: a failed reset leaves `app_config` on the new folder with the index describing the old one. Pre-existing, made less bad by this PR, and the doc comment says so rather than implying an atomicity the caller does not have.
3. **Nobody has interrupted a real migration or a real folder switch.** The tests force failures at chosen points; they establish the transaction semantics, not that a power cut mid-`reset_sync_state` behaves as modelled.

### Verdict

**PASS.** Every command green, every AC mapped to a test that a mutation reddens, and the flake re-verified under the harness CI actually runs.

Issue #142, opened 2026-08-31
DBSYNC-79 Slice 3: size the placeholder seam


mbsanchez:
Vertical slice 3 of **DBSYNC-79** — decide whether macOS adopts File Provider instead of the `.cloudsc` sidecar model.

**Type:** AFK · **Blocked by:** none (runs parallel to Slices 1 and 2) · **User stories:** 5, 8

## Why this slice exists

DBSYNC-79 was filed on the premise that `.cloudsc` is *the macOS placeholder model*. The code says otherwise, and this materially changes the cost of adoption:

```
283 references across 19 Rust files, plus 15 in TypeScript
zero target_os gates in cloudsc_ops.rs
13 references inside cloud_filter.rs — the Windows CfAPI path
```

So "remove `.cloudsc` on macOS" is not a macOS-local change. Without a platform seam first, it bleeds into Windows.

One mitigating measurement worth confirming: 336 references speak of `placeholder` generically against 169 naming `.cloudsc` literally, so the seam may be cheaper than the raw count suggests.

## Scope

`apps/desktop/src-tauri/src/`, `apps/desktop/src/` — reading only.

## Acceptance criteria

- [ ] The 283 Rust references split into sidecar-format concerns versus generic placeholder concerns, with counts and the method used to classify them.
- [ ] The shape of a per-platform placeholder seam described, with the estimated cost of introducing it.
- [ ] Confirmed and evidenced that the Windows CfAPI path can be left behaviourally unchanged by such a seam.
- [ ] The 15 TypeScript references assessed for whether they leak the sidecar concept into the UI.
- [ ] No code changes; the artefact is the estimate.

## Explicitly not this slice

Introducing the seam. That is a follow-up ticket if Slice 4 decides to adopt.

Jira: DBSYNC-79

mbsanchez:
## Implementation Plan

**Scope dir(s)**: `apps/desktop/src-tauri/src/`, `apps/desktop/src/` — reading only
**Implementing agent(s)**: rust-pro
**Estimated complexity**: low — mechanical, but must be reproducible
**Dependencies**: none (parallel with #140, #141)

### Steps

1. Write a classification script that partitions every `.cloudsc` reference into one of: **sidecar file format** (the magic bytes, the extension, the encode/decode helpers), **placeholder concept** (hydrate, evict, is-placeholder predicates), or **Windows CfAPI path**. Commit the script's output, not just the conclusion, so the count can be re-derived rather than trusted.
[...]

Issue #141, opened 2026-08-31
DBSYNC-79 Slice 2: map the sync engine onto the File Provider contract


mbsanchez:
Vertical slice 2 of **DBSYNC-79** — decide whether macOS adopts File Provider instead of the `.cloudsc` sidecar model.

**Type:** AFK · **Blocked by:** none (runs parallel to Slices 1 and 3) · **User stories:** 5

## Why this slice exists

File Provider is a *replicated* extension model: the provider owns enumeration, materialization and eviction. That is a different division of responsibility from the one the engine has today, and adopting it means handing parts of the engine to the system.

Before deciding, we need to know which parts — and which parts have no counterpart at all. A promising early signal: the engine already stores a `remote_delta_cursor`, which is the same shape as a File Provider enumeration anchor.

## Scope

`apps/desktop/src-tauri/src/` — reading only.

## Acceptance criteria

- [ ] Written mapping from the existing index and `remote_delta_cursor` to File Provider enumeration, the working set, and change anchors.
- [ ] Each engine responsibility classified: becomes the system's, stays ours, or has no counterpart.
- [ ] Gaps named explicitly — anything File Provider requires that the engine cannot currently supply.
- [ ] Every claim states how it was verified (file read, API doc, observed behaviour). No claim rests on recollection.
- [ ] No code changes; the artefact is the mapping.

## Explicitly not this slice

Writing adapters, refactoring the engine, or deciding whether to adopt. This slice measures; Slice 4 decides.

Jira: DBSYNC-79

mbsanchez:
## Implementation Plan

**Scope dir(s)**: `apps/desktop/src-tauri/src/` — reading only
**Implementing agent(s)**: rust-pro
**Estimated complexity**: medium — no code, but the reading is broad
**Dependencies**: none (parallel with #140, #142)

### Steps

1. Read the index schema in `storage/db.rs` — `local_file_index`, `remote_file_index`, `sync_jobs`, `known_folders` — and record what each row means and who writes it.
2. Trace `remote_delta_cursor` through `remote_index.rs`: how it is seeded, advanced, and reset. This is the closest existing analogue to a File Provider enumeration anchor and the mapping's most promising anchor point.
3. Read `sync_pipeline.rs` and `cloudsc_ops.rs` for the materialization and eviction paths — what today triggers a hydrate, what deletes local content, and what decides a file is cloud-only.
[...]

Issue #140, opened 2026-08-31
DBSYNC-79 Slice 1: disposable spike — prove a second appex can ship


mbsanchez:
Vertical slice 1 of **DBSYNC-79** — decide whether macOS adopts File Provider instead of the `.cloudsc` sidecar model.

**Type:** AFK · **Blocked by:** none · **User stories:** 6

## Why this slice exists

This is the only unknown that documentation cannot settle: whether *our* pipeline can bundle, sign and notarize a second appex, and whether a File Provider extension coexists with the Finder Sync extension we already ship.

Reading Apple's docs tells you File Provider exists. It does not tell you that `tauri.conf.json` will carry a second appex through `codesign`, that the App Group will not collide with the read-only temporary-exception path DBSYNC-76 chose instead of one, or that notarization accepts the pair. Those are answered by building the smallest thing that can fail.

**Already verified before this slice** (on macOS 26.5.1, by observation):
- Both Dropbox and OneDrive ship a `com.apple.fileprovider-nonui` appex *and* a `com.apple.FinderSync` appex.
- Both File Provider appex are signed `Developer ID Application` with hardened runtime — not App Store — so our distribution model is a shipping path.
- Both carry `com.apple.security.app-sandbox` and `com.apple.security.application-groups`.

## Scope

`apps/desktop/native/macos/`, `apps/desktop/src-tauri/`

## Acceptance criteria

- [ ] A minimal `NSFileProviderReplicatedExtension` with a stub domain enumerates a handful of hardcoded items.
- [ ] The domain mounts under `~/Library/CloudStorage/` and Finder lists those items.
- [ ] The appex is bundled by the existing Tauri build, signed with Developer ID and notarized by the DBSYNC-72 release workflow; `stapler validate` and `spctl --assess` pass.
- [ ] The App Group is confirmed to coexist with the existing Finder Sync extension — both load, neither breaks.
- [ ] Findings recorded: entitlements required, bundling changes required, and anything the pipeline could not do.
- [ ] **The spike is deleted before DBSYNC-79 closes.** It lives outside production config and ships nothing.
- [ ] `cargo test`, `cargo clippy`, `cargo fmt --check` and `npm run test` are green at the end — any red means the spike leaked into production config.

## Explicitly not this slice

Building the production extension, touching `cloudsc_ops.rs`, or changing any user-visible behaviour. The deliverable is a yes/no plus notes.

Jira: DBSYNC-79

mbsanchez:
[...]

PR #144, opened 2026-08-31
chore(#DBSYNC-79): prove the File Provider packaging path with a disposable spike

mbsanchez:
## Summary

DBSYNC-79 asks one question: does macOS adopt File Provider instead of the `.cloudsc` sidecar model? **This PR does not answer it** — it produces measurements for the ADR, and a disposable spike that answers the packaging question documentation cannot.

Slices 2 (#141) and 3 (#142) are complete and changed no code. Slice 1 (#140) is partial.

> **Two claims in the first version of this PR were withdrawn at review** (commit 2 of 2). Both were true observations with unsupported conclusions attached. They are described below rather than deleted, because the pattern is the useful part.

## What is in the diff

| Change | Fate |
| --- | --- |
| `.github/workflows/release.yml` — verify **every** nested appex, not one hardcoded path | **kept** |
| `native/macos/FileProviderSpike/` — the spike target | deleted before DBSYNC-79 closes |
| `tauri.conf.json` — second `bundle.macOS.files` entry | deleted with it |
| `apps/desktop/.gitignore` — spike build output | deleted with it |

The release workflow fix is **not spike-specific**: it named `DropboxSyncFinderSync.appex` literally, so a second extension would have shipped unverified — the same class of gap that left the first appex unchecked until #84.

## Findings that hold

**A read-only File Provider cannot exist.** `createItem`, `modifyItem` and `deleteItem` sit before `@optional` in the SDK header; the build failed with "does not conform" until all three existed. The system owns the filesystem and pushes mutations at the provider. This is the strongest evidence that adoption rewrites the macOS path rather than adding a backend, and it is the PR's most valuable result.

**Identity is the dominant cost.** All three index tables key on `relative_path`, and `NSFileProviderItemIdentifier` must survive renames. Dropbox's stable `id:` is never captured — the parser reads `path_display`, `content_hash`, `rev` and `server_modified` only.

**The seam already exists.** `cloud_filter.rs` carries `#![cfg(windows)]`, so the module is empty on macOS, and all four `placeholders_active` call sites are `#[cfg(windows)]`. A macOS arm cannot alter Windows behaviour. (This finding is itself a correction of an earlier wrong one — see the Jira description's C2.)

**`.cloudsc` cannot be deleted.** It serves three consumers: macOS, Windows without package identity, and Windows where registration or connection failed. Adoption removes the first only.

**Packaging works up to signing.** The appex builds, links, signs as `Developer ID Application: Manuel Sanchez (XCAA3WMJM6)` with hardened runtime, and passes the principal-class guard; the bundler carries a second appex.

## Claims withdrawn

**C3a — "the App Group does not need a provisioning profile".** The evidence was that `codesign` accepted the entitlement. **`codesign` validates no entitlement it is given:**

```
$ codesign -s "Developer ID Application: Manuel Sanchez (XCAA3WMJM6)" \
           --entitlements nonsense.entitlements  t
$ codesign -d --entitlements - t
    [Key] com.apple.developer.totally-made-up-entitlement
        [String] XCAA3WMJM6.com.this.group.does.not.exist.anywhere.at.all
```

Authorization happens at notarization and at runtime; neither was tested. **The App Group question is open and DBSYNC-76's recorded reason stands unrefuted.** The ADR must carry it as an unpriced cost.

**C3b — "non-packaged Windows is the only Windows mode".** Inferred from DBSYNC-92's "no installer". `native/windows/sparse-package/` grants package identity *without* an installer — that is its purpose — and `windows_identity.rs` gates on `GetCurrentPackageFullName`. Whether that package is installed on any given machine has not been measured, so the corrected text claims nothing in either direction.

## Stack

`desktop-tauri` · **Jira:** DBSYNC-79

## Test Plan

- [x] `cargo test` — **216 passed / 0 failed** (re-run after the corrections)
- [x] `cargo fmt --check` — clean
- [x] `cargo clippy` — 6 warnings, identical to the clean tree (verified by stashing, not assumed)
- [x] `npm run test` — green
- [x] Spike rebuilds after the comment edits; entitlements plist passes `plutil -lint`
- [x] Principal-class guard: declared = emitted = `FileProviderSpike`
- [x] Release workflow loop tested under real Bash 3.2, empty and two-appex cases; YAML parses
- [ ] **Notarization** — outward-facing, not run
- [ ] **Domain mounts / Finder lists items** — needs the host app to call `NSFileProviderManager.add(domain:)`
- [ ] **App Group authorized at runtime** — the open half of C3a

## Reviewer note

The previous version of this note asked whether any claim was stated more confidently than its evidence supports. Two were, and both were found by attacking the PR's own strongest claim rather than by re-reading it. The same question still applies to this version.

🤖 Generated with Claude Code


mbsanchez:
## CTO Review — CHANGES REQUESTED

The code is fine. **Both defects are in claims this PR makes**, and both are the same shape: a true observation with a false conclusion bolted on. That is the third and fourth time this ticket has carried one, so the pattern matters more than either instance.

Method: rather than hunting for the next problem, every claim in the PR body and the three findings comments was enumerated and attacked with "what would make this false?". Two did not survive.

---

### F1 — BLOCKER. "The App Group does not need a provisioning profile" is unsupported

The PR's headline finding. The evidence offered is that `codesign` accepted the entitlement and it appeared in the signed binary.

**`codesign` does not validate entitlements. It embeds whatever it is handed.** Demonstrated directly:

```
$ codesign -s "Developer ID Application: Manuel Sanchez (XCAA3WMJM6)" \
           --entitlements nonsense.entitlements  t
$ codesign -d --entitlements - t
    [Key] com.apple.developer.totally-made-up-entitlement
    [Key] com.apple.security.app-sandbox
        [String] XCAA3WMJM6.com.this.group.does.not.exist.anywhere.at.all
```

An App Group that exists nowhere, and an entitlement invented thirty seconds earlier, both embedded without complaint. So the observation is true of *any* string and distinguishes nothing.

Authorization is decided at **notarization** (Apple checks entitlements against the account) and at **runtime** (the group container is only created if the entitlement is authorized). This PR verified neither — and the PR itself says so, two paragraphs below the claim. The claim and its own caveat contradict each other; the caveat is right.

**Consequence:** DBSYNC-76's recorded reason for choosing a read-only temporary exception stands unrefuted. Saying otherwise in an ADR could commit the project to File Provider on a constraint we have not actually tested.

**Required:** restate as "the entitlement can be embedded; authorization remains unverified", and correct it in PR #144, GitHub #140, and the Jira description. The App Group question stays open until notarization or a runtime container test answers it.

---

### F2 — MAJOR. "Non-packaged Windows is the only Windows mode that exists today" is a false inference

Stated in the PR body, GitHub #142 and the Jira description, reasoning from DBSYNC-92 ("no workflow produces a Windows installer") to "so CfAPI is unavailable on Windows in practice".

`native/windows/sparse-package/` exists precisely to break that inference. Its README:

> Grants the unpackaged DropboxSync exe a **package identity** via a **sparse package** (an MSIX that contains only the manifest + logos; the exe stays external). Identity is required for the WinRT Cloud Files API sync-root registration.

It ships `build-and-install.ps1`, per-user, no admin at runtime. And `windows_identity.rs` gates on `GetCurrentPackageFullName`, which the sparse package satisfies — not on an installer existing.

**No installer does not imply no package identity.** That is the whole point of a sparse package.

**Consequence:** this one cuts against the PR's own conclusion, so it is worth stating plainly — the ".cloudsc is doing all the Windows work today" framing overstates how much rent the sidecar model is collecting. The narrower claim survives and is the one that matters: adoption removes one of three consumers, so `.cloudsc` cannot be deleted either way.

**Required:** correct in the same three places.

---

### Lesser — not blocking

- **"`~/Library/CloudStorage/` … a vendor cannot choose it"** sits under a heading promising observation. The observation is that the directory exists and Dropbox is in it; "a vendor cannot choose it" is an Apple-documentation claim. Attribute it as such.
- The spike's `enumerator(for:)` returns the same enumerator for every container including `.workingSet`, whose items all claim `.rootContainer` as parent. Harmless in a spike, wrong in a provider — worth a comment so it is not copied.

---

### What survives, and is good

- **"A read-only File Provider cannot exist."** Verified twice over: the three methods sit before `@optional` in the SDK header, and the build failed until all three existed. This is the PR's most valuable finding and it is properly evidenced.
- **"Identity is the dominant cost."** All three index tables key on `relative_path`; the metadata parser reads `path_display`, `content_hash`, `rev`, `server_modified` and no id. Checked at source.
- **The seam already exists** — `#![cfg(windows)]` on the module, all four `placeholders_active` call sites gated. This is itself a correction of an earlier wrong finding, and it is correct.
- **Test evidence.** 216/0, clippy compared against the clean tree by stashing rather than assumed, and the Bash 3.2 empty-array idiom tested in both the empty and two-appex cases before committing. That is the standard the claims above failed to meet.
- **The `release.yml` fix is right and should outlive the spike.**

---

### Verdict

Changes requested. No code changes needed — the diff is sound. Fix the two claims in PR #144, GitHub #140, GitHub #142 and the Jira description, then re-request review.

The reviewer note in this PR asked whether any claim was stated more confidently than its evidence supports. Two were.

mbsanchez:
## Follow-up — reviewing the correction found better evidence, and a flaw in the correction

The correction above replaced a bad argument with a **mechanism assertion that was itself unverified**: *"Authorization is decided at notarization (Apple checks entitlements against the account) and at runtime (the container is only created if authorized)."* Plausible, documented-sounding, and not something this ticket had observed. That is the same failure mode one level up.

So the reference Macs were checked instead of reasoned about.

### Observed

```
Dropbox.app/Contents/PlugIns/DropboxFileProvider.appex
  com.apple.security.app-sandbox         yes
  com.apple.security.application-groups   yes
  embedded.provisionprofile               NONE   ← 0 in the appex
  (1 profile elsewhere in the bundle, in an unrelated XPC helper)

OneDrive.app/Contents/PlugIns/OneDrive File Provider.appex
  embedded.provisionprofile               YES    (7 across the bundle)

~/Library/Group Containers/G7HH3F8CAK.com.getdropbox.dropbox.sync/   exists
```

### What this establishes

**Dropbox ships a notarized, Developer ID File Provider appex that declares an App Group and embeds no provisioning profile**, and the corresponding group container exists on this machine. So a provisioning profile is **not required** for notarization to accept an App Group on a Developer ID File Provider extension — there is a shipping counter-example.

The two reference providers differ on this: OneDrive embeds profiles, Dropbox does not. Either approach ships.

### What it still does not establish

- That the container was created *by the appex* rather than by the host app. The entitlement is declared on the appex and there is no profile in it; that is what was observed.
- That **our** entitlement will be accepted at notarization. Untested — the only way to close it is to notarize.

### Net position for the ADR

Better than "open", weaker than "solved":

> An App Group on a Developer ID File Provider appex is demonstrably shippable without an embedded provisioning profile — Dropbox does exactly that. Our own build has not been notarized, so treat it as **low-risk but unconfirmed**, not as a blocker and not as done.

This is the third time on this ticket that attacking a claim beat re-reading it — and the first time the attack was aimed at a correction rather than at an original. The original claim happened to point in the right direction; its evidence was still worthless, and finding real evidence changed the confidence, not just the wording.

mbsanchez:
## CTO Review, round 2 — APPROVED

Both round-1 findings are fixed, and the fix commit was reviewed with the same method rather than trusted. Every claim in the current PR body and the four findings comments was re-enumerated and attacked.

### The round-1 findings

**F1 (App Group)** — fixed, then improved beyond what was asked. The correction first replaced a bad argument with an unverified mechanism assertion ("authorization is decided at notarization and at runtime"), which the author caught in their own fix and replaced with observation:

```
DropboxFileProvider.appex   app-groups: yes   embedded.provisionprofile: NONE
OneDrive File Provider.appex                  embedded.provisionprofile: YES
~/Library/Group Containers/G7HH3F8CAK.com.getdropbox.dropbox.sync/   exists
```

Verified independently this round: `stapler validate` passes on Dropbox.app and `spctl` reports `source=Notarized Developer ID`, so "notarized" is not an assumption. The appex's own `spctl` rejection ("does not seem to be an app") is expected for an appex and is not a failure.

Net: a shipping counter-example proves no profile is required; our own build is still un-notarized. "Low-risk but unconfirmed" is the right strength.

**F2 (sparse package)** — fixed, and correctly resisted the opposite overclaim. The corrected text says CfAPI turns on package identity, that the sparse package grants it without an installer, and that whether it is installed on a given machine **has not been measured** — so it claims nothing in either direction. That restraint is the right call.

### Re-verified this round, not taken on trust

**"Dropbox's stable `id:` is never captured"** is the load-bearing claim behind "identity is the dominant cost", and round 1 accepted it on a two-file grep. Checked properly at the deserialization struct:

```rust
pub(crate) struct DropboxEntry {
    tag, path_display, content_hash, rev, server_modified, size
}
```

Six fields, no `id`, and it is the single parse point for Dropbox metadata. The claim is now verified more strongly than when it was first made.

The rest hold: the read-only-provider impossibility (SDK header + compiler error), the seam already existing (`#![cfg(windows)]` + four gated call sites), the test evidence (216/0, clippy compared against a stashed clean tree, Bash 3.2 idiom exercised empty and non-empty).

### Nit — not blocking

"`codesign` validates **no** entitlement it is given" is broader than what was tested: two entitlements, one invented, one a nonexistent group. The operative claim — a successful `codesign` does not establish authorization — is fully supported. Narrow the wording if it is ever quoted outside this ticket.

### Verdict

**Approved.** Moving to QA, where the three manual items live: notarization, the domain mounting under `~/Library/CloudStorage/`, and the App Group container at runtime.

Worth recording, because it is the ticket's real output: **four false claims across DBSYNC-79, none of them in code.** All four were true observations with unsupported conclusions attached, and all four fell to the same question — *what would look exactly like this if it were false?* The fourth fell only because that question was aimed at a correction rather than at an original. A review that had stopped at "the findings are fixed" would have shipped an invented mechanism into the ADR.

mbsanchez:
## QA Bug Report — the bundler AC is false

**AC failed:** *"The bundler carries a second appex"* — marked `[x]` in the Jira description and claimed in PR #144.

### Evidence

`tauri.conf.json` maps `PlugIns/FileProviderSpike.appex` to `../native/macos/FileProviderSpike/build/FileProviderSpike.appex`. That path **does not exist**:

```
EXISTS   native/macos/FinderSyncExtension/build/DropboxSyncFinderSync.appex
MISSING  native/macos/FileProviderSpike/build/FileProviderSpike.appex
```

The spike's built product lives under `build/DerivedData*/Build/Products/Release/FileProviderSpike.appex` — three copies, from three `xcodebuild` invocations, none at the path the bundler reads.

### Root cause

The Finder Sync appex reaches its expected location because `build-appex.sh` ends with a `ditto` into `build/`, and `apps/desktop/package.json` wires it as `build:finder-sync`. **The spike has neither.** Step 4 of its own implementation plan on this issue says *"Copy `build-appex.sh` and adapt it"* — that step was skipped, and the `[x]` was awarded on the strength of `tauri.conf.json` parsing as valid JSON.

### Why the earlier evidence did not catch it

The check run at implementation time was "does `tauri.conf.json` still parse, and does it list two entries?" It does. **A config that parses and names a nonexistent file looks exactly like a config that parses and names a real one** — which is the same failure shape as the `codesign` finding two rounds ago: an observation that cannot distinguish the passing case from the failing one.

### Required to fix

1. A `build-appex.sh` for the spike, adapted from the Finder Sync one, ending with the same `ditto` into `build/`.
2. An npm script alongside `build:finder-sync` so the bundle step actually produces it.
3. Re-verify by **building the app** and listing `Contents/PlugIns/` — not by re-reading the config.
4. Then run the release workflow's verification block against that real bundle, which is what would have caught this the first time.

### Not tested, and blocked behind this

Notarization, the domain mounting under `~/Library/CloudStorage/`, and the App Group container at runtime all require a bundle that contains the appex. They stay untested until the wiring exists.

mbsanchez:
## QA finding fixed — verified by building, not by reading

Commit 3 of 3 on this branch.

### What was missing

Three pieces, all of which the Finder Sync appex already had and the spike had none of:

1. `native/macos/FileProviderSpike/build-appex.sh` — adapted from the Finder Sync script, ending with the `ditto` into `build/` that puts the product where `tauri.conf.json` reads it. Carries the same principal-class guard.
2. `build:file-provider-spike` in `apps/desktop/package.json`, beside `build:finder-sync`.
3. A call in `scripts/tauri-before-build.mjs`, which is what `beforeBuildCommand` runs — a bare npm script would not have been enough, since nothing invokes it during a build.

### Verified by building the app

```
$ npm --workspace apps/desktop run bundle:dev
$ ls Contents/PlugIns/
DropboxSyncFinderSync.appex     com.apple.FinderSync
FileProviderSpike.appex         com.apple.fileprovider-nonui
```

And `release.yml`'s verification block, run against that real bundle:

```
loop found 2 appex
  checked: DropboxSyncFinderSync.appex
  checked: FileProviderSpike.appex
loop completed over every appex
```

Host and appex teams both read `not set` locally because `signingIdentity` is `"-"` for a dev bundle; the comparison is degenerate here and meaningful in the release workflow, but the **iteration** is what this exercises, and it now runs against a bundle that really contains two extensions.

Suite: 216 passed / 0 failed, `cargo fmt --check` clean.

### The reason this got through

The original check was "does `tauri.conf.json` parse, and does it list two entries?" It did. **A config that parses while naming a nonexistent file is indistinguishable from one naming a real file** — the check could not fail in the broken case, so passing it meant nothing.

That is the fifth claim on DBSYNC-79 to fail this way, and the second *after* the pattern was written down. Knowing the shape is not enough; the question has to be asked of every box being ticked, including the ones that feel mechanical. This one felt mechanical.

### Deletion checklist — now four items

- [ ] `apps/desktop/native/macos/FileProviderSpike/` (including `build-appex.sh`)
- [ ] the second `bundle.macOS.files` entry in `tauri.conf.json`
- [ ] `build:file-provider-spike` in `package.json` **and** its call in `scripts/tauri-before-build.mjs`
- [ ] the spike line in `apps/desktop/.gitignore`
- [ ] **keep** the `release.yml` appex-iteration change

mbsanchez:
## CTO Review, round 3 — APPROVED

Reviewing commit `577f696`, the fix for the QA rejection.

### The claim I did not accept: "verified by building the app"

The fix reported that both appex land in `Contents/PlugIns/`. True — but the build had been run **after** `build-appex.sh` was invoked by hand. That observation is equally consistent with:

- the before-build hook produces the appex, **or**
- the hook does nothing and the file was already there from the manual run.

Same shape as every other defect on this ticket: a check that cannot distinguish the passing case from the failing one. So it was separated:

```
mv build/FileProviderSpike.appex          → .ASIDE
mv build/DerivedData                      → .ASIDE
mv bundle/macos/DropboxSyncDesktop.app    → .ASIDE
npm --workspace apps/desktop run bundle:dev
```

Output, unprompted by any manual step:

```
> bash native/macos/FileProviderSpike/build-appex.sh
Building FileProviderSpike.appex (Release)…
build/FileProviderSpike.appex   → recreated
Contents/PlugIns/DropboxSyncFinderSync.appex
Contents/PlugIns/FileProviderSpike.appex
```

**The hook genuinely produces it.** The claim now rests on evidence that would have failed had it been false.

The third piece of the fix is the one that mattered: an npm script alone would not have run during a build. `beforeBuildCommand` calls `tauri-before-build.mjs`, and the call had to go *there*.

### Deletion checklist — verified complete, not just plausible

The spike must leave no residue. Every tracked reference was enumerated and compared against the checklist on #140:

```
apps/desktop/.gitignore
apps/desktop/native/macos/FileProviderSpike/*        (7 files)
apps/desktop/package.json
apps/desktop/scripts/tauri-before-build.mjs
apps/desktop/src-tauri/tauri.conf.json
```

Nothing outside the five checklist entries. The checklist is complete.

### State of the suite

216 passed / 0 failed after the from-scratch rebuild; working tree clean.

### Housekeeping for the maintainer

Three directories were moved aside for the test above and I cannot delete them:

```bash
rm -rf apps/desktop/native/macos/FileProviderSpike/build/FileProviderSpike.appex.ASIDE \
       apps/desktop/native/macos/FileProviderSpike/build/DerivedData.ASIDE \
       apps/desktop/src-tauri/target/release/bundle/macos/DropboxSyncDesktop.app.ASIDE
```

All three are gitignored, so they are clutter rather than risk.

### Verdict

**Approved.** Moving to QA.

Six defects on DBSYNC-79 now, **none in code**. The last three were all the same failure: a check that passes identically whether the thing works or not — `codesign` succeeding, a config parsing, a bundle containing a file I had put there myself. The fix is not more care; it is a mechanical habit of asking, before ticking any box, *"could this have passed while broken?"* — and if the answer is yes, breaking it on purpose to watch the check fail.

mbsanchez:
## QA Result: PASSED ✅ — with three acceptance criteria explicitly NOT met

Passing here means *what this PR changes does what it claims*. It does **not** mean Slice 1 is complete. Three of its six ACs remain open and are listed below rather than quietly folded into a green tick.

### Verified

| Check | Result |
| --- | --- |
| Both appex land in a freshly built bundle | ✅ `DropboxSyncFinderSync.appex` + `FileProviderSpike.appex` |
| The before-build hook produces the spike appex unaided | ✅ proven by moving the appex, its `DerivedData` and the prior bundle aside first |
| `release.yml`'s verification block iterates every appex | ✅ run against the real bundle, found and checked both |
| Extension points | ✅ `com.apple.FinderSync` and `com.apple.fileprovider-nonui` |
| Principal-class guard | ✅ declared = emitted |
| Suite | ✅ 216 passed / 0 failed, `cargo fmt --check` clean |
| Deletion checklist complete | ✅ every tracked spike reference falls inside the five listed entries |

### A non-finding, recorded so nobody re-derives it

`pluginkit -m -p com.apple.fileprovider-nonui` does **not** list our extension. That looks like a failure and is not one:

```
~/Applications/DropboxSyncDesktop.app   built 21 Aug   PlugIns: DropboxSyncFinderSync.appex only
fresh build                             built  2 Sep   PlugIns: both
both: CFBundleIdentifier com.mobsanchez.dropboxsyncdesktop
```

PlugInKit holds one record per bundle identifier, and the record points at the **installed copy from twelve days ago**, which predates the spike. The check could not have detected a working spike either, so it distinguishes nothing — the same shape as the three defects this ticket already logged. Testing registration for real requires installing the new build over the existing one.

### Not met, and deferred by decision

- **The domain mounts under `~/Library/CloudStorage/` and Finder lists the items.** Needs the host app to call `NSFileProviderManager.add(domain:)` from Rust. Feasible (`objc2` is already a dependency and `finder_extension.rs` uses `msg_send!`), but it means production code in service of a spike that gets deleted.
- **Notarized, `stapler validate` passes.** Outward-facing; not run.
- **The App Group is authorized at runtime.** Blocked on the two above.

The maintainer chose to close QA here rather than push further, on the reasoning that DBSYNC-79's deliverable is a **decision** and the spike has already answered the expensive questions: packaging works end to end, a read-only File Provider cannot exist, and identity is the dominant cost. The three open items confirm; they do not decide.

**Slice 4 must carry all three as open costs in the ADR**, alongside the App Group question, and must not read as though the spike validated a running extension. It did not.

Issue #148, opened 2026-09-02
DBSYNC-96 Slice 4: the sizing verdict and the ADR recommendation


mbsanchez:
Vertical slice 4 of **DBSYNC-96** — size stable item identity before DBSYNC-95 starts.

**Type:** HITL — the maintainer decides · **Blocked by:** #1, #2, #3 · **User stories:** 1, 2, 3

## What this produces

The number DBSYNC-95 is waiting for, and a recommendation about ADR-0003.

ADR-0003 was accepted with an explicit escape clause: *"its answer can still send this decision back here for amendment."* That clause is not decoration. If the identity cost turns out to be a data migration rather than a column plus an API, the direction is worth re-examining before four tickets of work proceed on it.

## Scope

`docs/adr/`

## Acceptance criteria

- [ ] A written sizing, split into: schema change, storage API, engine addressing surface, and back-fill of existing rows.
- [ ] An explicit **go / amend** recommendation, with **the threshold that would trigger amendment stated up front** rather than left to judgement after the fact. Deciding what counts as "too expensive" before seeing the number is the only way that judgement stays honest.
- [ ] If slice #2 confirmed the rename cost, the value case is restated in ADR-0003: this stops being macOS-only work and starts being a cross-platform defect fix that File Provider happens to require.
- [ ] If any of #1–#3 came back materially worse than the architecture review estimated, that is called out rather than absorbed.
- [ ] The maintainer makes the call. This slice presents evidence and does not decide.

## Explicitly not this slice

Writing any of the identity code. Starting DBSYNC-95.

Jira: DBSYNC-96

mbsanchez:
## Implementation Plan

**Scope dir(s)**: `docs/adr/`
**Implementing agent(s)**: the maintainer decides; rust-pro drafts
**Estimated complexity**: low to write, high in consequence
**Dependencies**: blocked by #145, #146, #147

### Steps

1. **Write down the amendment threshold before reading the findings.** What size of identity cost would make File Provider not worth it? Deciding that after seeing the number is how a decision gets rationalised instead of made. This is the single most important step in the slice and it costs five minutes.
2. Assemble the sizing from #145 (addressing surface), #146 (rename cost), #147 (identifier questions), split into: schema change, storage API, engine addressing, back-fill of existing rows.
3. Compare against the threshold from step 1. State go or amend.
[...]

Issue #147, opened 2026-09-02
DBSYNC-96 Slice 3: answer the two identifier questions


mbsanchez:
Vertical slice 3 of **DBSYNC-96** — size stable item identity before DBSYNC-95 starts.

**Type:** AFK · **Blocked by:** none · **User stories:** 1, 3

## Two questions, two different sources

Both must be answered before the sizing means anything, and they fail in opposite directions.

### Dropbox's side — is `id:` the identifier we think it is?

`DropboxEntry` has six fields and none is an id:

```rust
struct DropboxEntry { tag, path_display, content_hash, rev, server_modified, size }
```

It is the only parse point for Dropbox metadata, so nothing anywhere captures an id today.

- [ ] Confirm `id:` is present on **every entry kind we index** — files *and* folders, since `known_folders` needs identity too.
- [ ] Confirm it is stable across rename **and** move. Name how it was confirmed; an API doc sentence and an observed round-trip are not the same evidence.
- [ ] Determine what identifies an item that has **no Dropbox id yet**: created locally and not uploaded, or queued behind a failed upload. File Provider will ask about these items before Dropbox has ever seen them.

### Apple's side — how stable is "stable"?

- [ ] Determine whether `NSFileProviderItemIdentifier` must survive **provider restarts and domain re-registration**, not merely renames. If it must, identifiers have to be persisted rather than derived, and the cost rises materially.

This was flagged as an open question in DBSYNC-79 and carried forward unanswered. It is being carried for the second time; it does not get carried a third.

## Scope

`apps/desktop/src-tauri/src/` — reading only, plus API/documentation research.

## Acceptance criteria

- [ ] All four questions above answered, each citing its source.
- [ ] Any answer that rests on documentation rather than observation is **labelled as such**.
- [ ] No code changes.

## The bar

DBSYNC-79 produced six false claims. Every one was a true observation with an unsupported conclusion attached, and every one was caught by asking *"what would look exactly like this if it were false?"* — never by re-reading. Apply that question to each answer here before writing it down.

Jira: DBSYNC-96

mbsanchez:
## Implementation Plan

**Scope dir(s)**: `apps/desktop/src-tauri/src/` — reading, plus API and documentation research
**Implementing agent(s)**: rust-pro
**Estimated complexity**: medium — two sources, two failure modes
**Dependencies**: none (parallel with #145, #146)

### Steps — Dropbox side

[...]

Issue #146, opened 2026-09-02
DBSYNC-96 Slice 2: confirm what a rename costs today


mbsanchez:
Vertical slice 2 of **DBSYNC-96** — size stable item identity before DBSYNC-95 starts.

**Type:** AFK · **Blocked by:** none · **User stories:** 2, 4, 5

## The claim to confirm or kill

**Renaming a file today appears to cost a full delete plus re-upload.** Three facts from the code support it, and **none has been observed at runtime**:

1. The watcher discards everything but the path — `events.into_iter().map(|e| e.path)` in `fs_watcher.rs` — so it cannot tell a rename from a delete followed by a create.
2. The only job types that exist are `upload`, `download`, `delete`. There is no `move` or `rename`.
3. A vanished path goes through `enqueue_targeted_deletions`; the new path enqueues an upload.

## Why it matters more than it looks

If it holds, **stable identity stops being macOS-only work.** It fixes a live defect on both platforms: rename a 2 GB file, re-upload 2 GB. Rename a folder, multiply by its contents. That changes the cost/benefit in ADR-0003 — the identity work would be paying for itself outside File Provider entirely.

If it does not hold, the description of DBSYNC-96 is wrong and must be corrected. That outcome is just as valuable and must not be quietly dropped.

## Scope

`apps/desktop/src-tauri/src/` — a test is expected here, unlike the other slices.

## Acceptance criteria

- [ ] Rename an indexed file and **observe** what the pipeline enqueues. A test or a run — reading the code is what produced the claim, so it cannot also be what confirms it.
- [ ] The same for a directory rename.
- [ ] If confirmed: quantify it. What does renaming an N-byte file actually transfer, and does the delete reach Dropbox before the upload?
- [ ] If **not** confirmed: say so plainly, correct the DBSYNC-96 description, and record what the pipeline does instead.
- [ ] Suite green: `cargo test`, `cargo fmt --check`, `cargo clippy`, `npm run test`.

Prior art for the test shape: `targeted_removed_file_enqueues_delete` in `sync_pipeline.rs`.

## Explicitly not this slice

Implementing rename or move as a first-class operation, however tempting the finding makes it. That is its own ticket if this confirms.

Jira: DBSYNC-96

mbsanchez:
## Implementation Plan

**Scope dir(s)**: `apps/desktop/src-tauri/src/sync_pipeline.rs`
**Implementing agent(s)**: rust-pro
[...]

Issue #145, opened 2026-09-02
DBSYNC-96 Slice 1: measure the addressing surface


mbsanchez:
Vertical slice 1 of **DBSYNC-96** — size stable item identity before DBSYNC-95 starts.

**Type:** AFK · **Blocked by:** none · **User stories:** 1

## Why this slice exists

The architecture review measured the coupling twice and got two different answers, because the first measurement counted the wrong thing:

```
SQL sites filtering on relative_path        18
relative_path mentions                      78   (58 of them in storage/db.rs)
engine fns taking rel: &str                 34   (across 7 files)
── total path-as-identity sites            117
```

"74% is inside db.rs" looks like the coupling is contained in the storage layer. It is not. **`db.rs`'s public API is itself path-addressed** — `get_local_file(relative_path)`, `upsert_local_file(...)`, `remove_remote_file(relative_path)`, `upsert_known_folder(relative_path)` — so it publishes the addressing model rather than hiding it, and every caller inherits it.

18 query sites would be a weekend. 117 addressing sites is a different ticket. Establishing which number is real is what this slice is for.

## Scope

`apps/desktop/src-tauri/src/` — reading only.

## Acceptance criteria

- [ ] The 117 sites partitioned by a **reproducible script**, into three buckets: must become identifier-addressed, can stay path-addressed, ambiguous. Commit the script's output so the count can be re-derived rather than trusted.
- [ ] `db.rs`'s public API enumerated in full and classified per method — it is the boundary every caller inherits.
- [ ] The shape of an identifier-addressed API described, with the cost of **adding it beside** the path-addressed one rather than replacing it. Windows keeps path addressing regardless, so a wholesale replacement may be work nobody needs.
- [ ] For each bucket, a one-line reason. A partition without reasons cannot be argued with.
- [ ] No code changes.

## Explicitly not this slice

Introducing the API. Changing the schema. Sizing is the deliverable.

Jira: DBSYNC-96

mbsanchez:
## Implementation Plan

**Scope dir(s)**: `apps/desktop/src-tauri/src/` — reading only
**Implementing agent(s)**: rust-pro
**Estimated complexity**: low mechanically, medium in judgement
**Dependencies**: none (parallel with #146, #147)

### Steps

[...]

PR #149, opened 2026-09-04
test(#DBSYNC-96): prove a rename is billed as a delete plus a full upload

mbsanchez:
## Summary

DBSYNC-96 sizes stable item identity before DBSYNC-95 is allowed to start. **The deliverable is a measurement, not code** — the diff is two tests and nothing else. The findings live on the slice issues; this PR carries the one thing that had to execute.

Slices: #145 (addressing surface), #146 (rename cost), #147 (identifier questions), #148 (sizing).

## What the tests prove

A rename is billed as a **delete plus a full upload**, not a move. The index row is dropped rather than moved, so sync state is lost as well as bytes re-sent. A directory rename costs one delete per tracked descendant **plus the folder row**.

The first version of the directory test asserted two deletes and was wrong — the observed answer is three. Assertion corrected to match observation; production was not at fault.

**Both paths go into one `process_changed_paths` call on purpose.** The watcher delivers a single debounced batch, so old and new arrive together. Splitting them would have been a weaker test that could not detect a correlation even if the pipeline had one.

**Both tests were proven able to fail**, which is the part that makes them evidence:

| Mutation | Reddens |
| --- | --- |
| `enqueue_targeted_deletions` → `Ok(0)` | both rename tests, by name (+3 existing) |
| upload enqueue target corrupted | both rename tests, by name (+1 existing) |

Naming them matters — "3 tests reddened" would have been satisfied by three unrelated failures.

## The sizing, in one table

| | |
| --- | --- |
| Must convert | **23 sites** — shared engine code |
| Contingent | 15 — `cloudsc_ops.rs`, which File Provider replaces on macOS |
| Never touched | 9 — `cloud_filter.rs`, `#![cfg(windows)]` |
| `db.rs` methods needing identity | **9 of 48** |
| `db.rs` methods that are hierarchy-shaped | 7 — identifiers cannot express subtrees, so paths stay |

The architecture review's "117 sites" merged three different populations. This is **addition, not replacement**: the path-addressed API survives for Windows and for hierarchy.

**One thing got bigger.** `local_file_index` and `remote_file_index` are separate tables, and a locally created file has no remote row until its upload succeeds. So identifiers cannot be a thin mirror of Dropbox's id — a local identity space is needed. That was not in the original framing.

## What is honestly not answered

- **Does Dropbox's `id:` exist on folders and survive renames?** Needs a live call against the maintainer's account. Not made — their credentials, their decision.
- **Apple's stability requirement.** The SDK header states none; Apple's site is a JS application and returns only a page title. A fetch tool offered "typically stable and persistent" with no source; **discarded rather than recorded**, because that is precisely the shape of DBSYNC-79's fourth correction.

Mitigated but not closed: ids must be persisted anyway, because the SDK header warns identifiers *"may be recorded in system logs"* — so a path-derived identifier would leak user paths into the unified log. Same defect class as DBSYNC-91.

## A weakened safeguard, disclosed

Slice #148's plan required the amendment threshold to be written **before** the findings were read. The same agent produced the findings, so there was no moment at which that was possible. Writing one now and calling it prior would be fiction. **The threshold is the maintainer's to set**, and it is being set with the numbers already visible.

## Stack

`desktop-tauri` · **Jira:** DBSYNC-96

## Test Plan

- [x] `cargo test` — **218 passed / 0 failed**
- [x] `cargo fmt --check` — clean (it caught a mis-wrapped assertion first; fixed)
- [x] `cargo clippy` — 6 warnings, the pre-existing set
- [x] Both new tests shown to redden under two independent mutations, by name
- [x] No production code changed — the diff is tests only

## Reviewer note

The code is two tests. **The surface worth attacking is the claims**, on #145–#148: whether any count separates populations it should not, and whether any answer is stated more confidently than its source supports. Two are explicitly marked unanswered; check that nothing else should be.

🤖 Generated with Claude Code

### DIFF
--- a/apps/desktop/src-tauri/src/sync_pipeline.rs
+++ b/apps/desktop/src-tauri/src/sync_pipeline.rs
@@ -1582,6 +1582,88 @@ mod tests {
         assert_eq!(job_targets(&state, "upload"), vec!["m.txt".to_string()]);
     }
 
+    /// DBSYNC-96: what does a rename actually cost today?
+    ///
+    /// The claim under test is that the pipeline cannot see a rename at all — it sees a path
+    /// that vanished and a path that appeared, and bills them as a delete plus a full upload.
+    /// That claim came from reading three things (`fs_watcher` keeps only `e.path`, there is no
+    /// `move` job type, a vanished path reaches `enqueue_targeted_deletions`), and reading is
+    /// not evidence of behaviour.
+    ///
+    /// **Both paths go into ONE call on purpose.** The watcher hands over a single debounced
+    /// batch, so the old and new names arrive together. Splitting them across two calls would
+    /// be a weaker test: it could not detect a correlation even if the pipeline had one, and
+    /// would "confirm" the claim for the wrong reason.
+    #[test]
+    fn a_rename_is_billed_as_a_delete_plus_a_full_upload() {
+        let tmp = tempdir().expect("tempdir");
+        let state = build_state(tmp.path());
+        // Indexed and in sync under its old name...
+        state.db.upsert_local_file("old.txt", "h", 5, 0).unwrap();
+        // ...and on disk it now exists only under the new one. Same bytes, same inode in a
+        // real rename; the pipeline sees neither.
+        std::fs::write(sync_root(&state).join("new.txt"), b"hello").unwrap();
+
+        let n = process_changed_paths(&state, &["old.txt".to_string(), "new.txt".to_string()])
+            .expect("process");
+
+        // Two jobs, not one move.
+        assert_eq!(n, 2);
+        assert_eq!(job_targets(&state, "delete"), vec!["old.txt".to_string()]);
+        assert_eq!(job_targets(&state, "upload"), vec!["new.txt".to_string()]);
+        // And the index row does not travel: it is dropped and a fresh one is created.
+        assert!(state.db.get_local_file("old.txt").unwrap().is_none());
+        assert!(state.db.get_local_file("new.txt").unwrap().is_some());
+    }
+
+    /// The same for a directory, where the cost multiplies by the contents rather than
+    /// being paid once.
+    #[test]
+    fn a_directory_rename_re_uploads_every_child() {
+        let tmp = tempdir().expect("tempdir");
+        let state = build_state(tmp.path());
+        let root = sync_root(&state);
+
+        // Old directory: tracked folder + two tracked children, none of them on disk any more.
+        state.db.upsert_known_folder("d").unwrap();
+        state.db.upsert_local_file("d/one.txt", "h1", 3, 0).unwrap();
+        state.db.upsert_local_file("d/two.txt", "h2", 3, 0).unwrap();
+
+        // New directory on disk with the same two files.
+        std::fs::create_dir_all(root.join("e")).unwrap();
+        std::fs::write(root.join("e/one.txt"), b"aaa").unwrap();
+        std::fs::write(root.join("e/two.txt"), b"bbb").unwrap();
+
+        process_changed_paths(
+            &state,
+            &[
+                "d".to_string(),
+                "e".to_string(),
+                "e/one.txt".to_string(),
+                "e/two.txt".to_string(),
+            ],
+        )
+        .expect("process");
+
+        let mut deletes = job_targets(&state, "delete");
+        let mut uploads = job_targets(&state, "upload");
+        deletes.sort();
+        uploads.sort();
+        // Three deletes, not two: the folder row goes as well as both children.
+        assert_eq!(
+            deletes,
+            vec![
+                "d".to_string(),
+                "d/one.txt".to_string(),
+                "d/two.txt".to_string()
+            ]
+        );
+        assert_eq!(
+            uploads,
+            vec!["e/one.txt".to_string(), "e/two.txt".to_string()]
+        );
+    }
+
     #[test]
     fn targeted_removed_file_enqueues_delete() {
         let tmp = tempdir().expect("tempdir");


mbsanchez:
## CTO Review — CHANGES REQUESTED

The tests are good and the rename finding stands. **The sizing does not.** Two defects, both in the Slice 1 partition, and both from the same cause: it classified `db.rs`'s methods **by name pattern** rather than by signature.

The script was reproducible. It was also reproducibly wrong — which is worth stating plainly, because "reproducible" was offered as the quality bar.

---

### F1 — BLOCKER. The item-addressed count is 11, not 9. `add_conflict` was missed entirely.

Classifying by actual signature rather than by name, **13** public methods take a path-shaped argument, not 9:

```
add_conflict            ← MISSED, and it takes THREE paths
enqueue_delete_job      ← MISSED
get_local_file          get_remote_file          mark_local_file_for_rescan
remove_known_folder     remove_local_file        remove_remote_file
upsert_known_folder     upsert_local_file        upsert_remote_file
remove_remote_subtree   (hierarchy)              set_sync_folder (root config)
```

`add_conflict` is the significant miss:

```rust
pub fn add_conflict(
    &self,
    local_path: &str,
    remote_path: &str,
    reason: &str,
    conflicted_copy_path: Option<&str>,
    remote_deleted: bool,
) -> AppResult<()>
```

**Three path-shaped arguments, and it records per-item conflict state.** A conflict row that identifies its item by path has exactly the problem this ticket is about: rename the file and the conflict record no longer points at it. It fell into the "path-irrelevant" bucket because the name contains no matching substring.

**Required:** correct the count to 11 item-addressed, and treat `sync_conflicts` as a fourth table carrying identity — the sizing currently names three.

---

### F2 — MAJOR. The hierarchy count is 1, not 7. Six config accessors were counted as addressing.

The hierarchy bucket was reported as 7 and used to carry a real argument — *"identifiers cannot express subtrees, so paths stay; this is addition, not replacement."* Six of those seven take **no path at all**:

```rust
pub fn set_include_prefixes_csv(&self, csv: &str) -> AppResult<()> { ... }
pub fn get_include_prefixes_csv(&self) -> AppResult<Option<String>> { ... }
pub fn set_ignore_globs_csv(&self, csv: &str) -> AppResult<()> { ... }
```

They read and write a **CSV blob in `app_config`**. They address nothing. Only `remove_remote_subtree` genuinely expresses hierarchy in the storage API.

**The conclusion survives; its evidence does not.** `remove_remote_subtree` alone still cannot be identifier-addressed, and the prefix filters still match on path shape at their *call sites*. So "addition, not replacement" holds — but it was being supported by a 7 that was really a 1, and a reader would have weighted it accordingly.

**Required:** correct the count and re-state the argument on the evidence that actually supports it.

---

### What survives, and is strong

- **The rename finding.** Confirmed by test, both tests shown to redden under two independent mutations **by name**. The directory case corrected its own assertion against observed behaviour rather than the other way round. This is the ticket's real output and it is properly evidenced.
- **The three-population split** — 23 must-convert / 15 contingent / 9 Windows-only — is sound and is the correction the "117" needed. F1 and F2 change the `db.rs` breakdown, not this.
- **The `cloudsc_ops` contingency is honestly labelled** rather than banked.
- **The local identity space finding.** Separate tables, a real window with no Dropbox id — genuine design cost that was not in the original framing.
- **Slice 3's two non-answers**, including discarding a plausible-sounding fetch result with no source behind it. That is the ticket's own standard being applied to itself.
- **The disclosure that Slice 4's threshold could not be set blind.** Fabricating one would have been the cheaper path.

---

### The pattern, now measured

This ticket set out to correct a misleading count — "18 SQL sites" measured queries instead of addressing. **Its own replacement count was misleading in the same family**, in both directions: understating item-addressed by two, overstating hierarchy by six.

That is the fourth instance across DBSYNC-79 and DBSYNC-96 of a number that was defensible, reproducible, and measuring something other than what it was used to argue. The defence is not more care with counts. It is: **when a number carries an argument, verify it against the thing itself — signatures, not names.**

---

### Verdict

Changes requested. The tests need nothing. Correct the two counts on #145, in the PR body, and in Slice 4's sizing on #148, then re-request review.

mbsanchez:
## Correction — the partition was wrong, and so was the review that caught it

Three successive counts of the same thing, each from a different regex, each wrong:

| Attempt | Item-addressed | Hierarchy | Method |
| --- | --- | --- | --- |
| Slice 1 findings | 9 | 7 | name patterns |
| CTO review | ~11 (said 13 take a path) | 1 | a `:\s*&?\s*str` type regex |
| **This correction** | **12** | **1** | **reading all 23 signatures** |

The review's own number was also short, and for a familiar reason: its regex required the type to start with `&str`, so `enqueue_job(source_path: Option<&str>, target_path: Option<&str>)` did not match. **A wrapped type hid a path argument from the check that was written to find hidden path arguments.**

### The corrected classification

`db.rs` has **48** public methods. 23 take an argument typed `str`/`String`/`Path`; the rest take none. Classifying those 23 by **what the argument denotes**, not by its name or its type:

```
ITEM-ADDRESSED (12)   get/upsert/remove_local_file, mark_local_file_for_rescan,
                      get/upsert/remove_remote_file,
                      upsert/remove_known_folder,
                      add_conflict, enqueue_delete_job, enqueue_job
HIERARCHY (1)         remove_remote_subtree(prefix)
INFRASTRUCTURE (2)    new_at(db file), set_sync_folder(sync root)
STRING-NOT-PATH (8)   app_config keys/values, job errors, retry stamp,
                      upload session id, the three CSV blobs
NO STRING ARG (25)
                      12 + 1 + 2 + 8 + 25 = 48 ✓
```

### What changes in the sizing

**`sync_conflicts` and `sync_jobs` are identity-bearing tables.** The sizing named three tables; it is **five**.

- `add_conflict` stores `local_path`, `remote_path` **and** `conflicted_copy_path`. Rename a file with an open conflict and the record points at nothing.
- `enqueue_job` and `enqueue_delete_job` address their targets by path. A rename mid-queue leaves a job aimed at a path that no longer exists — which is not hypothetical, given Slice #146 proved a rename *is* a queued delete plus a queued upload.

That second point is the one worth carrying to Slice 4: **the rename defect and the identity gap are the same defect seen from two ends.** The queue cannot survive a rename because the queue addresses by path.

### What does not change

**"Addition, not replacement" still holds** — but on the evidence that actually supports it, which is `remove_remote_subtree` plus the prefix filters at their *call sites*, not seven storage methods. Six of those seven read and write a CSV blob and address nothing.

The three-population split (23 must-convert / 15 contingent / 9 Windows-only) is unaffected; it was measured on the engine surface, not on `db.rs`.

### The lesson, stated because it has now happened four times

Slice 1 offered **reproducibility** as its quality bar. The script was reproducible and reproducibly wrong. So was the review's. Reproducibility guarantees that the same wrong answer comes back.

**When a number carries an argument, the final pass must read the things themselves.** The count above came from reading 23 signatures and asking of each what the argument denotes — no pattern in the loop. That is slower and it is the only version that has survived scrutiny.

mbsanchez:
## CTO Review, round 2 — CHANGES REQUESTED

The count corrections hold. **A different claim does not**, and it is stated in three places.

---

### F1 — MAJOR. "The only job types that exist are upload, download and delete" is false. There are five.

The authoritative source is the schema's own CHECK constraint:

```sql
job_type TEXT NOT NULL CHECK (job_type IN
  ('upload','download','delete','local_delete','hydrate_cloudsc'))
```

**Five types.** The claim appears in Slice #146's findings, in PR #149's body, and in DBSYNC-96's description.

The cause is the sharpest instance of this ticket's recurring failure yet. The evidence was:

```bash
grep -rhoE '"(upload|download|delete|move|rename)"' *.rs
  17 "delete"   16 "upload"   5 "download"
```

**The pattern asked about three types and found three types.** It could not have reported `local_delete` or `hydrate_cloudsc` because it never asked. Its output was indistinguishable from a genuine enumeration of the space, and was read as one. Re-run including them:

```
19 "delete"  18 "upload"  11 "local_delete"  5 "hydrate_cloudsc"  5 "download"
```

**The load-bearing conclusion survives and is now better evidenced.** There is no `move` or `rename` job type — and the CHECK constraint proves that far more strongly than counting literals ever could, because it is the closed set the database enforces. The rename finding is unaffected.

**Required:** correct the claim in all three places, and cite the CHECK constraint rather than a grep.

---

### F2 — MINOR. "12 item-addressed" merges two different kinds of claim.

Nine of the twelve are observations: `get_local_file`, `upsert_remote_file`, `remove_known_folder` and the rest **are** the index, and identity is what they address.

Three are **judgements**: `add_conflict`, `enqueue_job`, `enqueue_delete_job`. Saying they *must become* identifier-addressed is a design position — defensible, and the reviewer agrees with it — but it is not the same kind of fact. A job queue that addresses by path is arguably fine if jobs are short-lived; it becomes wrong only once you decide jobs must survive renames.

Presenting 9 observations and 3 judgements as a single "12" is the same merge the correction criticised in the original "117". **Required:** split them, or label the three.

---

### Supporting the queue claim, and marking its limit

The correction asserted *"the queue cannot survive a rename because the queue addresses by path."* The drain path supports it — there is no re-resolution:

```rust
"upload" => job.source_path.as_deref()
    .and_then(|rel| upload_local_file_internal(state, rel, job.id)),
```

The job trusts the stored path. So after a rename, an in-flight job operates on a path with nothing at it.

**What happens then was not observed.** Whether it errors, retries to exhaustion, or silently no-ops is unknown, and the three outcomes differ in severity. State it as code-supported-but-unobserved, or test it.

---

### What holds

- The rename finding, its two tests, and their mutation proof.
- The corrected counts: 12 (with F2's caveat), hierarchy 1, five identity-bearing tables.
- The insight that the rename defect and the identity gap are one defect from two ends — F1 does not touch it.
- Slice 3's two honest non-answers, and the disclosure that Slice 4's threshold could not be set blind.

---

### The tally, because it is now the ticket's most useful output

Five wrong claims across DBSYNC-96, in three rounds, **none in code**:

| # | Claim | Instrument |
| --- | --- | --- |
| 1 | 18 SQL sites | grep for a column name |
| 2 | 9 item-addressed / 7 hierarchy | name patterns |
| 3 | 13 take a path (the review's own) | a type regex that missed `Option<&str>` |
| 4 | Only three job types | a pattern that asked about three |
| 5 | 12, presented as uniform | merged observations with judgements |

Every one was defensible, reproducible, and measured something other than what it was used to argue. **The instrument answered the question it was given, and the question was narrower than the claim.** The fix is not a better regex — it is to reach for the authoritative artefact: signatures, schemas, CHECK constraints. Every claim that survived scrutiny this round came from one.

mbsanchez:
## Correction — round 2 findings

### F1 — there are five job types, not three

**Was:** *"The only job types that exist are `upload`, `download` and `delete`. There is no `move` or `rename`."*

**Is:** the schema's CHECK constraint is the closed set the database enforces —

```sql
job_type TEXT NOT NULL CHECK (job_type IN
  ('upload','download','delete','local_delete','hydrate_cloudsc'))
```

Five. Re-counting the literals with all five in the pattern:

```
19 "delete"  18 "upload"  11 "local_delete"  5 "hydrate_cloudsc"  5 "download"
```

`local_delete` appears **11 times** and was invisible to a grep that never asked about it.

**What survives, on better evidence.** There is no `move` or `rename` job type, and the CHECK constraint proves that far more strongly than a literal count ever could: it is enforced by the database, so a `move` job could not be inserted even if some code tried. The rename finding, its tests, and the conclusion drawn from them are unaffected.

### F2 — the "12" splits into 9 observations and 3 judgements

**Observations (9).** These *are* the index; identity is what they address, and no design position is involved:

```
get_local_file   upsert_local_file   remove_local_file   mark_local_file_for_rescan
get_remote_file  upsert_remote_file  remove_remote_file
upsert_known_folder  remove_known_folder
```

**Judgements (3).** These address by path today, and saying they *must* become identifier-addressed is a design position — defensible, but not a fact of the same kind:

```
add_conflict        (local_path, remote_path, conflicted_copy_path)
enqueue_job         (source_path, target_path)
enqueue_delete_job  (target_path)
```

A job queue addressing by path is arguably fine if jobs are short-lived. It becomes wrong only once you decide jobs must survive a rename — which is a call for whoever picks up DBSYNC-95, not a measurement this ticket can make.

**So the sizing should read: 9 certain, 3 contingent on a design decision.** Presenting them as a flat 12 repeated the exact merge this correction criticised in the original "117".

### A third figure, corrected while checking the others

`db.rs` has **47 `impl Db` methods** — the API every caller inherits. An earlier correction said 48; that count included `app_data_dir`, a module-level free function that takes no arguments and is not part of the `Db` surface at all.

```
47 impl methods
├── 23 take a str/String/Path argument
│   ├── 12 path-as-item  (9 observed + 3 judged)
│   ├──  1 hierarchy     (remove_remote_subtree)
│   ├──  2 infrastructure (new_at, set_sync_folder)
│   └──  8 string-but-not-path (config keys, hashes, revs, errors, CSV blobs)
└── 24 take no string argument
```

### The instrument, each time

| Wrong claim | What the instrument was asked |
| --- | --- |
| 18 SQL sites | find a column name |
| 9 item / 7 hierarchy | match method names against a hand-written list |
| 13 take a path | match types starting with `&str` |
| only 3 job types | count three specific literals |
| 48 impl methods | count `pub fn` anywhere in the file |

**Every one answered its question correctly.** The defect was never the tool; it was that the question was narrower than the claim it was used to support, and the output of a narrow question is indistinguishable from the output of a complete one.

The figures above came from the schema's CHECK constraint and from reading 23 signatures — artefacts that define the space rather than sample it.

mbsanchez:
## CTO Review, round 3 — CHANGES REQUESTED

The round-2 corrections hold. **The engine number does not, and it is the half of the sizing DBSYNC-95 will actually act on.**

Three rounds of scrutiny went into `db.rs`'s 47 methods. The engine surface — "23 sites must convert, 15 contingent, 9 Windows-only" — has never been examined beyond a single `grep -c 'rel: &str'`. That is the same instrument class that produced all five earlier errors, and nobody went back for it.

---

### F1 — BLOCKER. The engine surface was measured by one spelling of one parameter name.

`rel: &str` is not how the engine names paths. Widening to *any parameter whose name suggests a path and whose type mentions `str`/`String`/`Path`* changes the picture in kind, not just in size.

**No multiple should be quoted** — the two counts measure different units (occurrences of one literal vs. parameters). What matters is the **categories the narrow pattern could not see**:

**(a) The same thing, spelled differently.** These are index-identity sites by any reading:

```
sync_pipeline.rs   process_local_file_change(relative: &str)
sync_pipeline.rs   process_local_file_deletion(prev_rel: &str)
remote_index.rs    fetch_remote_file_metadata(relative: &str)
```

`relative` and `prev_rel` are the same concept as `rel`. The pattern asked for one spelling.

**(b) Identity carried in a collection — the structurally important miss.**

```
remote_index.rs    reconcile_remote_snapshot_with_breaker(remote_by_path: &HashMap<String, RemoteFileMeta>)
remote_index.rs    remote_sweep_delete_candidates(remote_by_path: &HashMap<String, RemoteFileMeta>)
sync_pipeline.rs   process_local_file_change(pending_targets: &mut HashSet<String>)
sync_pipeline.rs   process_changed_paths(paths: &[String])
```

**The entire remote index is passed around as a map keyed by path.** Converting to identifier addressing is not "change a parameter" — it is changing that map's key type, which propagates to every consumer and every iteration over it. `pending_targets` is a set of paths used for membership tests; `paths` is a batch.

A per-parameter count cannot express this. One `HashMap<String, _>` is structurally heavier than ten `&str` arguments, and the sizing has no row for it.

**Also caught, in the other direction:** `classify_path(path: &Path)` and `overlay_state_path(data_dir: &Path)` are filesystem paths, not index identities. Widening the net catches non-identity too — which is exactly why this needs the treatment `db.rs` got: **read them and classify by what the parameter denotes**, not by matching a shape.

**Required:** re-measure the engine surface the way `db.rs` was measured. Until then, "23 must convert" is unsupported and must not be quoted to DBSYNC-95.

---

### Verified and holding

- **`db.rs` has exactly one `impl` block** (`impl Db`, line 77), so "47 impl methods" is not diluted by a second impl. Checked because the claim would have been quietly wrong if one existed.
- Five job types, from the CHECK constraint.
- The 9-observed / 3-contingent split.
- The rename finding, its two tests, and their mutation proof — untouched by this.

---

### The pattern, sixth instance, and what is actually different about it

| Wrong claim | Instrument | Asked for |
| --- | --- | --- |
| 18 SQL sites | grep | a column name |
| 9 item / 7 hierarchy | name patterns | a hand-written list |
| 13 take a path | type regex | types starting with `&str` |
| only 3 job types | literal count | three specific strings |
| 48 impl methods | `pub fn` count | any indentation |
| **23 engine sites** | **`grep -c 'rel: &str'`** | **one spelling of one parameter** |

The first five were caught by re-measuring the thing being counted. **This one was caught by asking which numbers had never been re-measured at all** — and the answer was the larger, less examined half.

That is a different check, and worth keeping: after correcting a figure, list every *other* figure in the same document and note which have been verified and which are inherited. The engine number survived three rounds not because it was checked and passed, but because attention stayed on the number that had already been challenged.

---

### Verdict

Changes requested. Tests and the rename finding need nothing. Re-measure the engine surface by reading and classifying, correct #145, #148, the PR and the Jira description, then re-request review.

mbsanchez:
## Correction — the engine surface, re-measured by reading

292 string/path-typed parameters exist in non-test engine code. Classified by **what each one denotes**, not by matching a shape:

### Must convert — 20, and six of them are collections

**Single-item identity (14).** A relative path naming one indexed item:

```
sync_pipeline.rs   placeholder_exists(rel)            delete_suppressed_by_dehydration(rel)
                   process_local_file_change(relative)
                   process_local_file_deletion(prev_rel)
                   process_known_folder_deletion(rel) enqueue_targeted_deletions(rel)
remote_index.rs    fetch_remote_file_metadata(relative)
                   reconcile_remote_present(rel)      reconcile_remote_absent(rel)
dropbox_transfer   upload_local_file_internal(relative)
                   delete_remote_file_internal(relative)
                   delete_local_file_internal(relative)
sharing.rs         create_or_get_shared_link(remote_rel)  remote_rel_from_item(rel)
```

**Identity in a collection (6)** — the category a per-parameter count cannot express:

```
remote_index.rs    reconcile_remote_snapshot_with_breaker(remote_by_path: &HashMap<String, RemoteFileMeta>)
                   reconcile_remote_snapshot_with_breaker(pending_targets: &HashSet<String>)
                   remote_sweep_delete_candidates(remote_by_path: &HashMap<String, RemoteFileMeta>)
                   remote_sweep_delete_candidates(pending_targets: &HashSet<String>)
sync_pipeline.rs   process_local_file_change(pending_targets: &mut HashSet<String>)
                   process_changed_paths(paths: &[String])
```

### Must NOT convert — 9 path-shape predicates

`path_util.rs` takes `relative` in nine places, and **none of them is identity**. They ask questions about the *shape* of a path — ignore globs, editor temp files, include/exclude prefixes, traversal safety:

```
should_ignore_local_path  is_editor_temp_path  matches_ignore_globs  user_ignore_globs_match
is_builtin_ignored_local_path  is_ignored_local_path  is_path_allowed
validate_relative  safe_join
```

An identifier cannot answer "does this match `*.tmp`". These stay path-based permanently, on both platforms — the same finding as `db.rs`'s hierarchy bucket, and a second reason this is **addition, not replacement**.

### Contingent — 19 in `cloudsc_ops.rs`

Including 3 collections (`emit_placeholder_changed(created/removed)`, `degrade_to_per_file(deleted)`). Whether these convert depends on whether DBSYNC-95's macOS arm bypasses the sidecar model.

### Never seen by a macOS provider — ~15 in `cloud_filter.rs`

`#![cfg(windows)]`; the module compiles to nothing on macOS.

---

## What actually changed, stated honestly

**The headline number barely moved: 23 → 20.** The old figure was roughly right *by accident*. Had the review stopped at comparing magnitudes, it would have concluded the grep was fine.

**What changed is the composition, and that is the finding:**

1. **Six sites carry identity in a collection.** `remote_by_path: &HashMap<String, RemoteFileMeta>` is the whole remote index keyed by path. Converting it is a key-type change that propagates to every consumer and every iteration — structurally heavier than the fourteen single-parameter sites combined, and the sizing had no row for it.
2. **Nine `path_util.rs` sites must stay path-based forever.** The old count neither included nor excluded them; it simply never saw them, because they spell the parameter `relative`.
3. The two mistakes **cancelled out in magnitude** — missed identity sites offset by non-identity sites that were never counted. That is why a wrong instrument produced a plausible number.

**This is the most useful thing this ticket has produced about measurement:** a number can be right in magnitude and wrong in every constituent, and comparing totals will not reveal it. The check that works is re-deriving the composition, not the sum.

mbsanchez:
## CTO Review, round 4 — APPROVED

This round applied the check the last correction proposed, rather than only recommending it: **every figure in the document was listed, and each marked re-measured or inherited.** Two came back needing attention. Both are minor, and that is the news.

### The audit

| Figure | Status |
| --- | --- |
| 47 `impl Db` methods | re-measured three ways ✅ |
| 23 take a string arg | **needed explanation — resolves** (below) |
| 12 path-as-item (9 + 3) | re-measured by reading ✅ |
| 1 hierarchy / 2 infrastructure / 8 string-not-path | same read ✅ |
| Five identity-bearing tables | derived ✅ |
| 292 engine string params | fresh scan ✅ |
| 20 must convert, 6 collections | fresh, by reading ✅ |
| 9 `path_util` predicates | fresh ✅ |
| 19 contingent in `cloudsc_ops` | **verified: 16 identity + 3 collections = 19** ✅ |
| ~15 Windows-only in `cloud_filter` | **actually 13** ⚠️ |
| 218 passed / 0 failed | run ✅ |

### The two

**"23 methods take a string argument" versus 26.** A fresh scan counting *all* functions in `db.rs` finds 42 string params across **26** distinct functions. The description says 23. Both are right: the earlier scan counted `pub` methods only, and three of the 26 are private helpers (`write_local_file_row`, `add_column_if_missing`, `data_dir_for`). **Consistent once the unit is stated** — and the unit was not stated. Worth one clause, since the next reader will hit the same discrepancy.

**"~15 Windows-only in `cloud_filter.rs`" is 13.** Counted properly: 44 string params across 34 functions, of which 13 are relative-path identity and 5 are path collections. The figure was hedged with a tilde, which is honest, but it was an eyeball rather than a count and it was two high. It changes nothing — `cloud_filter.rs` is `#![cfg(windows)]` and a macOS provider never sees it — but a hedge is not a licence.

### Verdict

**Approved.** Fix the two in passing; neither blocks. Moving to QA.

### On convergence

Round 1 changed the sizing's shape. Round 2 found a false fact whose conclusion survived. Round 3 found the largest number had never been re-measured at all. Round 4 found a units mismatch that resolves and a hedged approximation two too high.

That is convergence, and it came from changing what the review looked for each time:

1. re-measure the number that was challenged
2. re-measure the numbers that were *not* challenged
3. re-derive the composition rather than the sum
4. audit every figure for whether it was ever measured at all

Each check found what the previous one structurally could not. The fourth is the cheapest and would have caught rounds 2 and 3 — **it belongs first, not last.**

The code has not changed since the first commit. Six defects, none of them in it.

mbsanchez:
## QA Result: FAILED ❌ — the sizing was produced; the artefact it exists to feed was not

The measurement work passes. **Slice 4's ADR deliverables were never produced**, and that is the half that makes any of this actionable for DBSYNC-95.

### Verified ✅

| AC | Evidence |
| --- | --- |
| Suite green | 218 passed / 0 failed, `cargo fmt --check` clean |
| Rename cost observed, not inferred | two tests, each reddened by two independent mutations, by name |
| Addressing surface partitioned by reading | 47 `impl Db` methods and 292 engine params classified by what each denotes |
| Prefix/subtree operations called out | `db.rs` hierarchy (1) and `path_util.rs` shape predicates (9) |
| No production code changed | diff is 82 lines, tests only |

### FAILED ❌ — ADR-0003 is untouched since DBSYNC-79 merged

```
$ git log --oneline -- docs/adr/0003-macos-file-provider.md
3aa6435 chore(#DBSYNC-79): ... (#144)
```

One commit. Nothing from DBSYNC-96 reached it. Three consequences:

**1. The value case was not restated.** Slice #148's AC says *"If the rename finding holds, the value case is restated in ADR-0003 — this stops being macOS-only work."* **The finding held, confirmed by test.** The ADR still justifies the work purely as File Provider enablement and does not mention the rename cost at all — `grep -i rename` returns three hits, all of them the *identifier* requirement, none the defect.

**2. It still says three tables, twice.** Lines 49 and 100: *"All three index tables key on `relative_path`"*. DBSYNC-96 established **five** identity-bearing tables — `sync_conflicts` and `sync_jobs` too. A reader of the ADR gets a superseded number.

**3. No go / amend recommendation exists.** Slice #148's AC requires one. The sizing comment explicitly declined to make it, which was right about *the decision* being the maintainer's — but the recommendation itself was still owed, and the threshold that would trigger amendment was never written.

### Why this is a fail and not a nit

DBSYNC-96 exists to feed ADR-0003. Its findings currently live in Jira comments and four GitHub issues. **The ADR is the durable artefact in the repository** — the thing DBSYNC-95 will open — and it now contains two claims this ticket superseded, and none of what this ticket learned.

The inherited-figure problem the round-4 audit named applies to the ADR itself: *"the 121 production references"* appears twice and has never been re-measured either.

### Required

- [ ] Restate the value case in ADR-0003: the rename defect, confirmed by test, on both platforms.
- [ ] Correct "three index tables" → five identity-bearing tables, in both places.
- [ ] Add the sizing and an explicit go / amend recommendation, with the amendment threshold stated.
- [ ] Re-measure or footnote the inherited "121 production references".

The maintainer still makes the decision. Producing the recommendation and the corrected ADR is this ticket's job; choosing is theirs.

mbsanchez:
## QA finding fixed — the sizing is now in the ADR, not only in comments

Commit `3bb4da2`. `docs/adr/0003-macos-file-provider.md`: +87 / −9.

### The four required items

- [x] **Value case restated.** The rename defect — confirmed by test, live on both platforms — is now in the Evidence section. It was absent entirely; `grep -i rename` used to return three hits, all about the *identifier* requirement, none about the defect. Now eight.
- [x] **"All three index tables" corrected** to five, with the reason: `sync_conflicts` stores three paths per row and `sync_jobs` addresses its target by path. Rename a file with an open conflict, or with a job queued against it, and the record points at nothing. The one remaining occurrence of the old phrase is the correction note quoting it.
- [x] **Sizing and an explicit recommendation added**, in a new section, with the amendment threshold.
- [x] **The inherited "121 production references" re-derived**, not trusted: 121 exactly, plus 44 in tests, 29 Windows-specific and 89 generic placeholder references. It holds, and now says so.

### The recommendation

**Proceed.** Three reasons, weightiest first: the value case is no longer macOS-only; the storage cost is smaller than the ADR feared (9 of 47 methods, path addressing surviving beside the new API); and the window is cheap until the first publish.

**The amendment threshold** is a single condition: Dropbox's `id:` turning out absent or unstable for **folders**. `known_folders` needs identity as much as files do, and folder metadata is a different response shape. Nothing else found is capable of flipping the direction — the collection work is heavy but bounded, and the local identity space is design work rather than a migration.

**Recorded alongside it:** that threshold was written *after* the numbers were known. DBSYNC-96's plan required the opposite, and the same agent produced both the findings and the threshold, so there was no moment at which it could be set blind. The ADR says so in place, so a reader can discount it rather than having to discover it.

### What this failure actually was

The work existed. It was recorded where the work was happening — Jira comments, four issues — and not where it would be read. **A Jira comment and an ADR look equally "documented" from inside the ticket**, and only one of them is what DBSYNC-95 opens.

That is the same shape as this ticket's six measurement defects, one level up: the check ("is it written down?") passed identically whether or not the thing was written down *somewhere that matters*.

Suite: 218 passed / 0 failed, `cargo fmt --check` clean. Branch is now two commits: the tests and this.

mbsanchez:
## CTO Review, round 5 — CHANGES REQUESTED

The ADR update fixes what QA asked for. **It also exposes that the premise the entire plan rests on has never been observed**, and the same document both asserts it and doubts it.

This round ran the figure audit first — the check round 4 concluded belongs first — against the *new* ADR content rather than the numbers already fought over.

---

### F1 — BLOCKER. "Dropbox supplies a stable `id:`" is asserted twice, doubted once, and verified nowhere.

Three statements in one document:

```
line  49  "Dropbox supplies a stable id: on every entry"                    ← asserted as fact
line 136  "the id already arrives and is silently dropped"                  ← asserted as fact
line 184  "Whether Dropbox's id: exists on folders ... Needs a live API call" ← unverified
```

**What was actually checked** is that `DropboxEntry` carries no `deny_unknown_fields`, so serde silently ignores fields it does not declare. That establishes exactly one thing: **if** an id arrives, it is dropped. It establishes nothing about whether Dropbox sends one.

**And there is no observational evidence in the repository at all.** No recorded response, no test fixture, no JSON literal anywhere under `src/` contains an `id` field. Searched.

**Why this is the blocker and not a wording nit.** The whole sizing is "capture Dropbox's id, store it, address by it". Line 136 prices that as *trivial* — one field on one struct — and the price is only trivial if the id is already in the response. If it is not, the plan is not one field; it is a different plan.

**Fair to the claim:** Dropbox's public API does document an `id` on file and folder metadata, and this is very likely true. But **nobody in DBSYNC-79 or DBSYNC-96 cited that documentation**, and this ticket discarded a fetch answer for precisely the offence of sounding right with no source behind it. The same standard has to apply to a claim that agrees with us.

**Required:** state it uniformly. Either cite the API documentation at every one of the three sites, or mark all three unverified. Not one of each.

---

### The shape of this one is new

The six earlier defects were all *measurement* errors — an instrument narrower than the claim. This is different:

**A claim that grew more confident each time it was copied forward, while its evidence never existed.**

```
DBSYNC-79   "Dropbox supplies a stable id: on every entry"        stated
DBSYNC-96   "we never capture it"                                  restated, implying it is there
DBSYNC-96   "the id already arrives and is silently dropped"       now load-bearing and priced
```

Nobody added evidence at any step. Each restatement inherited the previous one's authority and spent a little more of it. The figure audit catches unmeasured *numbers*; this needed the same audit applied to **premises**, which is a wider net than round 4 cast.

---

### What holds

- Five identity-bearing tables, the rename finding and its mutation proof, the 20/9/19/13 engine partition, the collection finding.
- **The "121 production references" was genuinely re-derived** — 121 exactly, plus 44 test, 29 Windows, 89 generic. The inherited figure held. Good instinct to check it.
- The threshold disclosure. Writing "set after the numbers were known" into the ADR, where a reader can discount it, was the right call and the cheap path was the other one.
- QA's three required items are all done.

---

### Verdict

Changes requested. One blocker, no code involved. Make the id premise consistent across all three sites, then re-request.

Worth noting for whoever picks up DBSYNC-95: **the amendment threshold is "Dropbox's `id:` is absent or unstable for folders" — and that is the one thing in this ADR with no evidence on either side.** The threshold and the unverified premise are the same fact, which means the decision is resting its escape hatch on the question it never answered.

mbsanchez:
## Review round 6 + QA — APPROVED, and this is the last round

**Stopping rule applied**, agreed with the maintainer after they pointed out the loop: *a review round that finds nothing in code and nothing that changes a decision is the last round.*

Commit `a9b28bc` touches `docs/` only, and what it changed — an assertion downgraded to an explicitly unverified premise — is what round 5 asked for. Nothing in code, nothing decision-changing. **Approved. Closing.**

### QA

| Check | Result |
| --- | --- |
| `cargo test` | 218 passed / 0 failed |
| `cargo fmt --check` | clean |
| CI, all five jobs on `a9b28bc` | green |
| QA's three original items | done in `3bb4da2` |
| Round-5 blocker | fixed in `a9b28bc` |

### What this ticket delivered

**83 lines of code** — two tests proving a rename costs a full delete plus a full re-upload, each shown to redden under two independent mutations, by name. **125 lines of ADR** — the sizing DBSYNC-95 needs, in the artefact DBSYNC-95 will open.

The substance, for whoever picks up DBSYNC-95:

- **12 item-addressed storage methods** of `db.rs`'s 47, and **five** identity-bearing tables — `sync_conflicts` stores three paths per row, `sync_jobs` addresses its target by path.
- **20 engine sites must convert, and 6 of them carry identity in a collection.** `remote_by_path: &HashMap<String, RemoteFileMeta>` is the whole remote index keyed by path; that key-type change outweighs the fourteen single-parameter sites combined.
- **9 `path_util.rs` predicates must stay path-based forever** — an identifier cannot answer "does this match `*.tmp`". With `db.rs`'s hierarchy operations, that is the second independent reason this is addition rather than replacement.
- **The rename defect is live on both platforms**, so this stopped being macOS enablement and became a cross-platform defect fix.
- **One question is unanswered and it is the amendment threshold**: whether Dropbox returns an `id:` at all. Answer it first. One API call.

### The honest retrospective

Six wrong claims across five rounds, **none in code**. Rounds 1–3 each changed the sizing materially. **Rounds 4 and 5 did not move anything** — a hedged `~15` that was 13, an undeclared unit, an assertion downgraded, two section headings merged, one cross-reference pointing the wrong way.

They found real things. Prose always has real things. **That is the failure mode, not the achievement:** reviewing one's own prose with no stopping rule terminates only when someone outside notices. The maintainer did.

The rule now exists, and DBSYNC-95 starts with it.

